PR #3, opened 2011-11-16
Easier development with local input

dpavlin:
Fixes --device option and adds --goto to open pdf at specified page
Also adds new test target which should be run as root for access to local /dev/input/event?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,3 +99,8 @@ display:
 	[ -p $(EMULATE_EINKFB_FILE) ] || mknod $(EMULATE_EINKFB_FILE) p
 	# ...and display from it
 	mplayer -rawvideo format=y8:w=$(EMULATE_EINKFB_W):h=$(EMULATE_EINKFB_H) -demuxer rawvideo $(EMULATE_EINKFB_FILE)
+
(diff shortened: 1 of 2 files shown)


hwhw:
This is great - however, in the meantime I've added a big chunk of code, sorry, I didn't keep an eye on github. I hope you'll like it - its an emulation for the framebuffer (and input at the same time), using SDL. So this will allow running completely on the PC for development and doesn't need root permission (or similar) for accessing input.
The problem however is that now I have a hard time merging your patches. As I see it, the input related stuff might have become obsolete? I like the goto-page, so it would be nice if you could send this in another pull request. The missing "d" option has been fixed in the mean time.

hwhw:
I hope you don't mind that I'll close this request for now. I would really love to cherry-pick now, but I'm too inexperienced with git to do so and github is just telling me "This pull request cannot be automatically merged." and I don't really know how to get around that.

hwhw:
nevermind, I found out how to cherry-pick here. Also, I had previously forgotten to commit my changes to reader.lua. I hope I integrated your changes well. I left out the input handling, however.

dpavlin:
Good work. SDL support makes it a joy to develop this reader :-)

PR #5, opened 2011-11-17
Change gamma using keypresses

dpavlin:
Just to make it clean: I don't think this naive implementation is good enough for merge (for example, we should really expire cache of all pages when changing gamma and UP/DOWN bindings are clearly wrong), but I would appreciate comments if I'm even on the right track.


hwhw:
I think that's just right. I'll commit it and will work on it in the near future.

PR #6, opened 2011-11-19
modify_gamma now use cache correctly

dpavlin:
This exposed small bug in checkcache which checked just page no
which didn't get invalidated when closing page, so we now check if page
is also not nil


hwhw:
ah, great. And you also found where to limit memory usage for muPDF. That's also a good idea.

dpavlin:
Memory usage is somewhat misleading: it's not total amount of memory used, but limit of single pixmap (there can be more than one pixmap on single pdf page), but conservative value does work and prevents OOM killer from kicking in (as a side-effect, it also destroys full-page color ads which is -- feature I guess :-)

PR #7, opened 2011-11-21
scp all *.lua files to Kindle 

dpavlin:
Latest changes to fit page are just... great :-) Are you planning to add arbitrary zoom/pan? I think we should move gamma bindings from 5way's UP/DOWN to some other binding (volume comes to mind ;-)


hwhw:
Thanks :-) It's not quite finished, panning should be added - and arbitrary zooming, yes. I think I have to restructure the call path for rendering. I tried not to, but it's too messy. Thanks for cleaning up after me :-) Appreciated!

PR #8, opened 2011-11-26
panning and lauchpad integration

dpavlin:
Panning was really easy to implement (once I read whole source again) and they I got carried away with implementation of smaller and smaller steps which are required to correctly zoom page on device :-)


hwhw:
This is great! You beat me to getting to do it, wonderful!

PR #10, opened 2011-12-23
Rotate page to better represent comics and other horisontal pdfs

dpavlin:
I was playing around with displaying comics, and page rotation came to mind. I implemented artificially small increment of 10 degrees to demonstrate point of rotation (which seems to be upper left).

It's eye candy in this phase, because all clipping are wrong after rotation. I was hoping that moving of rotation before or after clipping will change that, but it didn't ;-) Now, I'm stuck deciding whether to implement additional fixes after rotation or just special cases for 0, 90, 180 and 270 degrees.

Am I missing something in mupdf? I scanned through apps/pdfapp.c which comes with mupdf, and it seems they are mainlining clippings separate from mupdf's dc notion of viewport, but you might have better ideas, I'm sure :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,8 +127,19 @@ $(LUALIB):
 
 thirdparty: $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIBS)
 
+INSTALL_DIR=kindlepdfviewer
+
 install:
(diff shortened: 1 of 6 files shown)


hwhw:
Thank you, nice patches!

PR #11, opened 2012-02-05
navigation support in rotated screen

houqp:
most of the time I use kindle in landscape mode so I add this to the filechooser. :-)


hwhw:
I think 90-degrees-rotation is what most people want, fractional rotation is a nice proof-of-concept, but is rather esoteric in its practical use :-) So thank you for this!

PR #22, opened 2012-03-02
added Shift+B to add current page to jumpstack

dpavlin:
This allows us to bootstap jumpstack if we don't have table of contents


houqp:
that's nice, it can work as a temporal bookmark feature :-)

dpavlin:
Keyboard shortcut pushed me in right direction ;-)

PR #23, opened 2012-03-02
four bug fixes & one refactoring

houqp:
# fix
- change misplaced carriage ret characters (0x0d) in TOC entry
  to white space (0x20) to avoid displaying empty boxes.
- move shiftmode and altmode to global

now shiftmode and altmode is recoreded in Keys module.  So all the UIs
are sharing these two mode instead of remembering their own. These fix
the bug in UI switching.

For instance, you use combo 'Alt'+'b' to fire up some menu in pdfreader,
the altmode is set, but the alt key was released after the menu show
up. Then the key release event will only be captured by the menu. So
after you exit from the menu, the altmode is still set in pdfreader,
which is not what we want.
- refresh screen more elegantly after reader exit
- fix bug in recursive searching in filesearcher
# refactoring
- handle shift and alt key events in adjustKeyEvents(), to minimize code duplication.

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -146,9 +146,13 @@ function FileChooser:choose(ypos, height)
 		end
 		local ev = input.waitForEvent()
 		if ev.type == EV_KEY and ev.value == EVENT_VALUE_KEY_PRESS then
-			--print("key code:"..ev.code)
+			print("key code:"..ev.code)
 			ev.code = adjustFWKey(ev.code)
-			if ev.code == KEY_FW_UP then
+			if ev.code == KEY_SHIFT then
+				Keys.shiftmode = true
+			elseif ev.code == KEY_ALT then
+				Keys.altmode = true
+			elseif ev.code == KEY_FW_UP then
 				prevItem()
 			elseif ev.code == KEY_FW_DOWN then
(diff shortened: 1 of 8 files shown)


dpavlin:
Maybe could we could move alt and shift handing into adjustFWKey to keep code duplication down?

houqp:
Agree with you, but the function needs to have a new name. Or we should write a new wrapper function do the handing stuff and call adjustFWKey in side. Any suggestions for the name?

Also I am thinking about writing a inputloop module, and all the UIs using inputs can just register callbacks for each input event to the inputloop module. This can greatly minimize the code we need to write for UIs. How do you think about this?

dpavlin:
I would prefer to rename adjustFWKey if we need, so we can call just one function.

I don't really know how callbacks look in lua ;-) I'm lua newbie. If they look readable in source code, I'm all for it.

I would also love to have shortcut help somewhere near callback in source so we can generate shortcut help screen automatically and have definition in one place.

houqp:
OK, added adjustKeyEvents() 

Ha, I am a lua newbie too.  :-)  I will show you the code if I have the inputloop module done.

The help auto generation idea is great!

PR #40, opened 2012-03-06
change short cuts for menu items

houqp:
Uses Q-P and A-Del


houqp:
@dpavlin : what do you mean by visual queue? I just changed the background to indicate different row. Will be better if you can help improve it :)

dpavlin:
Changing background is great. I was thinking about line between chars, but background is nicer :-)

PR #42, opened 2012-03-06
Keyboard shortcuts to jump to 0%, 10%, 20%, .. , 100% of currently reading document.

traycold:
Shortcut to jump to pages corresponding to 0%, 10%, 20%, .., 90%, 100% of currently reading document using respectively keys 1, 2, 3, .., 9, 0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ ifdef EMULATE_READER
 	EMU_CFLAGS?=$(shell sdl-config --cflags)
 	EMU_CFLAGS+= -DEMULATE_READER \
 		     -DEMULATE_READER_W=$(EMULATE_READER_W) \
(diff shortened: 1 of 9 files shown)


traycold:
hi, possibly i made a mistake sending this pull-request (i'm a github noob). I meant to include only the commit `41e9b21` _"Shortcut to jump to page corresponding to 0%, 10%, 20%, .., 100%"_ (that actually only involve unireader.lua).
Sorry for the inconvenient; i will try do redo it correctly.

traycold:
i tried again to make a new pull request, but i cannot select the commits to include in the pull request only one the commit mentioned in the above comment. So I guess this is the correct way to go, at least I hope so..

PR #53, opened 2012-03-08
90 degree rotation support & 1 fix & 1 modification

houqp:
# Fix
- only get usedBBox in fit to content mode which makes rendering in fit to page mode a little bit faster.
# Mod
- read cache settings from .reader.kpdfview.lua
# Screen rotation implemented by ioctl

Now you get a instance rotate :) With only one bug.
I have to restore to previous rotation state of the
native system after exists. Because the native system
does poll the new rotation state that changed by ioctl.

Currently, I don't know how to force the native system
to detect the changes.

### DIFF
--- a/einkfb.c
+++ b/einkfb.c
@@ -180,6 +180,29 @@ static int einkUpdate(lua_State *L) {
 	return 0;
 }
 
+/* NOTICE!!! You must close and reopen framebuffer after called this method.
+ * Otherwise, screen resolution will not be updated! */
+static int einkSetOrientation(lua_State *L) {
+#ifndef EMULATE_READER
+	FBInfo *fb = (FBInfo*) luaL_checkudata(L, 1, "einkfb");
+	int mode = luaL_optint(L, 2, 0);
+
+	if (mode < 0 || mode > 3) {
+		return luaL_error(L, "Wrong rotation mode %d given!", mode);
+	}
+
+	/* ioctl has a different definition for rotation mode. */
+	if (mode == 1) 
+		mode = 2;
+	else if (mode == 2)
+		mode = 1;
+
+	ioctl(fb->fd, FBIO_EINK_SET_DISPLAY_ORIENTATION, mode);
+#endif
+	return 0;
+}
+
+
 static const struct luaL_reg einkfb_func[] = {
 	{"open", openFrameBuffer},
 	{NULL, NULL}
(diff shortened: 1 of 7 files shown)


hwhw:
Hm. I'm not to sure about the approach. Maybe, we even need both ways of rotation. I think the case is not clear when we want to rotate: Is it because the page has wrong metadata in the PDF (i.e. the PDF contains a wrongly rotated page) - or is it because we want to read in another orientation of the reader? Depending on this "why" we might want to switch meaning of direction keys or we might not.

When using built-in rotation sensors (like on the KDX), the case of rotating the device can be taken care of somewhat automatically (does it work right now? I don't think so, but I'm not sure). But I really wouldn't focus on auto-rotation on the KDX, that would probably need interfaces to get signaled about such changes.

So, we might need two seperate rotation modes. On the other hand, this would feel very awkward for the casual user. Bah. I'm not sure what we would really want :-)

Maybe we can beat those 2 seconds (which are basically an estimation anyway) by interfacing with the framework. I think there might be a DBus message when the display is rotated. I'd have to check that, though.

houqp:
> Hm. I'm not to sure about the approach. Maybe, we even need both ways of rotation. I think the case is not clear when we want to rotate: Is it because the page has wrong metadata in the PDF (i.e. the PDF contains a wrongly rotated page) - or is it because we want to read in another orientation of the reader? Depending on this "why" we might want to switch meaning of direction keys or we might not.

Ah, agree with you.

> When using built-in rotation sensors (like on the KDX), the case of rotating the device can be taken care of somewhat automatically (does it work right now? I don't think so, but I'm not sure). But I really wouldn't focus on auto-rotation on the KDX, that would probably need interfaces to get signaled about such changes.

No, it does not work for auto-rotation. As you said, now we are not able to capture that signal.

> So, we might need two seperate rotation modes. On the other hand, this would feel very awkward for the casual user. Bah. I'm not sure what we would really want :-)

I think we should just give enough freedom to users and let them choose what the need. So I am for the separate rotation modes.

> Maybe we can beat those 2 seconds (which are basically an estimation anyway) by interfacing with the framework. I think there might be a DBus message when the display is rotated. I'd have to check that, though.

Yes, I have heard that `lipc-send-event` is a high level wrapper tool for sending DBUS message. I am not familiar with DBUS. I tried to list the interfaces in kindle before, but failed. If you succeed, please teach me how to do it :)

houqp:
Or there might be a lower level way to rotate the screen without the help from framework. I don't know how Duokan made it.

houqp:
New implementation for screen rotation, see if this is better :)

PR #61, opened 2012-03-11
Add djvu TOC support

Germanc:
- Add djvu TOC support (outlines)

Feel free to ignore the pull request if I did it incorrectly (first time using git)


dpavlin:
It works on my single djvu file with TOC. I will have to find a few more, I guess :-)

Could you please indent your code using tabs so we don't introduce flowing layout?

PR #63, opened 2012-03-12
Prevent jump to percent to be added to the jumpstack

Germanc:
I still don't understand very well how git/github works, so have a little patience with me :P

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -209,17 +209,13 @@ function UniReader:setzoom(page)
 	local dc = self.newDC()
 	local pwidth, pheight = page:getSize(self.nulldc)
 	print("# page::getSize "..pwidth.."*"..pheight);
-	local x0, y0, x1, y1 = 0, 0, pwidth, pheight
-
-	-- only get usedBBox in fit to content mode
-	if self.globalzoommode <= self.ZOOM_FIT_TO_CONTENT and
-	self.globalzoommode >= self.ZOOM_FIT_TO_CONTENT_HALF_WIDTH_MARGIN then 
-		x0, y0, x1, y1 = page:getUsedBBox()
-		if x0 == 0.01 and y0 == 0.01 and x1 == -0.01 and y1 == -0.01 then
-			x0, y0, x1, y1 = 0, 0, pwidth, pheight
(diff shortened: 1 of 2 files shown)


unireader.lua
```diff
@@ -701,14 +701,15 @@ function UniReader:inputloop()
...
 			elseif ev.code == KEY_VPLUS then
 				self:modify_gamma( 1.25 )
 			elseif ev.code == KEY_VMINUS then
 				self:modify_gamma( 0.8 )
 			elseif ev.code == KEY_1 then
-				self:goto(1)
+				self:goto(1, 1)
 			elseif ev.code >= KEY_2 and ev.code <= KEY_9 then
-				self:goto(math.floor(self.doc:getPages()/90*(ev.code-KEY_1)*10))
+				self:goto(math.floor(self.doc:getPages()/90*(ev.code-KEY_1)*10), 1)
```
houqp: I think I just found something wrong here. When we use percent jump here, the page we jump to is the one that we don't want to record in jump stack. But we still want to save current page to jump stack right?

unireader.lua
```diff
@@ -431,13 +431,13 @@ function UniReader:del_jump(pageno)
 end
 
 -- change current page and cache next page after rendering
-function UniReader:goto(no)
+function UniReader:goto(no, nojump)
 	if no < 1 or no > self.doc:getPages() then
```
houqp, line 535: I recommend using different argument name for nojump here. When reading this definition, it might gives people a illusion like goto a specific page but not jump to it, which might brings in confusion.

houqp:
So far, you are doing well with github :) I don't understand git very well neither.

unireader.lua
Germanc: You're right ! I was reading the page numbers all wrong

Germanc:
Since I was wrong about which page was being add to jumpstack and also jump 10 pages at a time is no needed now that there is a proper go to page, I'll close the pull request.

houqp:
Your idea to not saving pages jumped by percent is right, please do remember to send a pull request when you rewrite that fix :)

PR #64, opened 2012-03-13
Try to fix some bugs in new cache system

houqp:
I am not sure my fixes are right or not, but at least it works for me. ;)

In draworcache method, I added `offset_x_in_page` and `offset_y_in_page` variables to represents offset within page the page to draw. So we don't need to use `self.offset_x` and `self.offset_y` in following calculations. `self.offset_x` and `self.offset_y` might be positive or negative, which is easy to introduce bugs. 
- The first bug is not handling positive self.offset values, which leads to this output: http://imgur.com/7Ku01
- I am the one to blame for the second bug: ZOOM_FIT_TO_CONTENT_WIDTH_PAN mode is not taken into consider. This leads to this output: http://imgur.com/1hrBu. I added ZOOM_FIT_TO_CONTENT_WIDTH_PAN mode to simulate the page turning experience in native reader days before. ;P

Besides what I mentioned above, the new cache system works great, I can observe that when panning in ZOOM mode, no page:draw() method is called :)

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -182,15 +182,24 @@ function UniReader:draworcache(no, preCache)
 	local page = self.doc:openPage(no)
 	local dc = self:setzoom(page)
 
+	-- offset_x_in_page & offset_y_in_page is the offset within zoomed page
+	-- they are always positive. 
+	-- you can see self.offset_x_& self.offset_y as the offset within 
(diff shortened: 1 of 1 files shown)


hwhw:
Merging this now, let's see if we get a stable state at least for release... In the worst case at least a few things are now working again.

PR #66, opened 2012-03-15
Cursor support in input box

houqp:
Now we have a cursor in inputbox which can be moved with fiveway keys. I hope this cursor class can be used to highlight texts while reading in the future.

Also the `Alt`+`Del` combo can be used to delete all contents in inputbox.

I am not sure whether this should be included in current release though :)


houqp:
I am integrating the cursor support with text highlight. So this pull request should be closed and I will fire another one after the integration.

Issue #79, opened 2012-03-30
memory handing and problems


dpavlin:
there are too many memory knobs as described by our problem to communicate correct settings on mobileread forum http://www.mobileread.com/forums/showpost.php?p=2023375&postcount=194

Right now, we in `unireader.lua` we have:

``` lua
    -- tile cache configuration:
    cache_max_memsize = 1024*1024*5, -- 5MB tile cache
    cache_item_max_pixels = 1024*1024*2, -- max. size of rendered tiles
```

but we also have in `pdfreader.lua`:

``` lua
function PDFReader:open(filename)
    -- muPDF manages its own cache, set second parameter
    -- to the maximum size you want it to grow
    local ok
    ok, self.doc = pcall(pdf.openDocument, filename, 64*1024*1024)
```

It seems to me that we need to do `grep MemFree /proc/meminfo` as sane default value (in lua, somehow :-) and then assign that in some ratio to `cache_max_memsize`, `cache_item_max_pixels` and mupdf if we are viewing pdf file (otherwise, it would be nice to have more memory for first two variables).

Any takers?

houqp:
Are you referring to change the default value on the fly(before every document open) or just before kindlepdfview starts?

dpavlin:
I think that getting free memory has to be done in lua, before first document is loaded, but re-distribution between buffers of that value should be done on each document open (for change between pdf and djvu #80 for example)

Even better, we can also check document size on open, and it's really massive ask user if they want to stop kindle framework to gain additional memory.

houqp:
That will some how make the memory cache uncontrollable by users. So we need a very intelligent auto cache setting system if take this approach.

Also I notice that it is hard to predict memory usage according to file size. One of the file in our share point is only 3M, but still can leads to reboot. And another 100M pdf file of mine consumes less memory than that one. :(

vdp:
Here is an interesting observation in respect to the memory issues I am experiencing...
I increased "cache_max_memsize" to 15MB and "cache_item_max_pixels" to 5MB from their default values of 5MB and 2MB respectively. Then I open a file (the behavior is the same no matter whether it's a DJVU or PDF) and zoom in at "sufficiently" high magnification level. Then I cycle between two pages back and forth several times while watching the memory usage in 'top'. With these new settings for the tile cache the memory usage stays the same, whereas with the original/smaller values it raises steadily at a suspiciously uniform rate(say 2% per page turn).
[...]

PR #82, opened 2012-03-31
Memory usage configurable from lua

### LINKED ISSUE #80: cache size control for DJVU
Hi,

I just found that djvulibre also provides API for cache size control(starts from ddjvuapi.h line 220). Hope somebody can take a look and add a knob in our code.

dpavlin:
If this changes seem sane, we can close #80 but for #79 will need /proc/meminfo parsing in lua, which I still don't know how to write.

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -455,6 +455,14 @@ static int getCacheSize(lua_State *L) {
 	return 1;
 }
 
+static int setCacheSize(lua_State *L) {
+	DjvuDocument *doc = (DjvuDocument*) luaL_checkudata(L, 1, "djvudocument");
+	int size = luaL_checkint(L, 2);
+	printf("ddjvu_cache_set_size = %d\n", size);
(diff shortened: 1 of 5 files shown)


houqp:
OMG, did I miss clicked close pull request? Can you send it again?

houqp:
what kind of information do you want to get from /proc/meminfo ?

dpavlin:
`grep MemFree: /proc/meminfo | cut -d: -f2 | sed 's/^ *//' | cut -d" " -f1` in lua :-)

houqp:
then just do a `os.execute` ;P

dpavlin:
I'm reluctant to os.execute because that involves using additional memory for shell and all little commands that get executed, so I would prefer to parse lines from lua so we can get more accurate number. I'm also thinking about using /proc/$pid/status to compare it with system's view of how much RAM we are using.

This might even allow to pre-cache more pages if we have spare memory.

PR #88, opened 2012-04-06
Screen rotation handling for entering/leaving screen saver

houqp:
Not sure whether this is the best way, but at least it works :P

I gave up the idea of using ioctl on `/proc/pmic` in `input.c` because I cannot come up with a good solution to watch the event with `select` system call. So I decide to fake an event file like those in `/dev/input`.  

My first try was using the `uinput` module. But the `uinput` module in kindle seems a little wired, I can create a new file under `/dev/input`, but write to it generates no fake event. Then I switched to the named pipe.

The big change in this commit is I stopped the framework on kpv start, so if kpv crashed, people need to restart it and exit normally to resume the framework.

This pull request only fixes screen saver in readers, not filechooser, selectmenu or filesearcher. To fix this, we need to rewrite those module with the new command module. Since I have already rewritten selectmenu with command in my epub branch and the epub branch is expected to be finished in two or three days, I think it is better to merge the epub branch to master first before we rewrite those modules. Otherwise we might end up with lots of conflicts :P

PS:
I guess the `slider_watcher.c` can be better improved by calling ioctl on `/proc/pmic`, but I don't have so much free time playing on it :)

PS2:
This pull request also includes a small fix for a bug reported by eLiNK.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,8 @@ THIRDPARTYLIBS := $(MUPDFLIBDIR)/libfreetype.a \
 
 LUALIB := $(LUADIR)/src/liblua.a
 
+all:kpdfview slider_watcher
+
 kpdfview: kpdfview.o einkfb.o pdf.o blitbuffer.o drawcontext.o input.o util.o ft.o lfs.o $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) $(DJVULIBS) djvu.o
 	$(CC) -lm -ldl -lpthread $(EMU_LDFLAGS) -lstdc++ \
 		kpdfview.o \
@@ -83,6 +85,9 @@ kpdfview: kpdfview.o einkfb.o pdf.o blitbuffer.o drawcontext.o input.o util.o ft
 
 einkfb.o input.o: %.o: %.c
 	$(CC) -c $(KPDFREADER_CFLAGS) $(EMU_CFLAGS) $< -o $@
+
+slider_watcher: slider_watcher.c
+	$(CC) $(CFLAGS) $< -o $@
 
 ft.o: %.o: %.c
 	$(CC) -c $(KPDFREADER_CFLAGS) -I$(FREETYPEDIR)/include $< -o $@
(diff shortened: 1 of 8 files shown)


dpavlin:
This is important improvement, so I think this code should be merged, but I'm wondering would it be possible to fork lipc-wait-event from reader instead of depending on external daemon?

We should probably add killall -cont cvm to `kpdf.sh` launchpad script to be cure we are not leaving -stop(ed) on crash (will will never happen :-)

houqp:
> This is important improvement, so I think this code should be merged, but I'm wondering would it be possible to fork lipc-wait-event from reader instead of depending on external daemon?

When we are waiting for the key events, we are blocked at the `select` system call, so if we want to notify the reader, we have to wake the reader up from the `select` system call. And the only approach I can come up with is to fake another "input event device" and have it "selected" with other input event devices. By doing so, we can make sure the system call will return once we write something to the fake device. 

The bad part here is the key events are binary data, so writing these data in C is very handy. I just picked up the tool that saves me most time :P Writng another daemon also makes the fake event transparent to Lua script and it feels cleaner.

Ok, I confess that the main reason is I have no experience in mutithread programming in Lua ;P

Feel free to help improve it :)

> We should probably add killall -cont cvm to kpdf.sh launchpad script to be cure we are not leaving -stop(ed) on crash (will will never happen :-)

Good suggestion! I have added it, thanks :)

When I double check kpdf.sh, I found that I actually forgot to stop the framework on kpv start and the kindle can still go into screen saver without any problem. So for going into screen saver, our Lua script is fast enough to retore the rotation mode before the native system start to redraw screen. Thus no need to stop the framework.

But when going out of screen saver, the framework will try to draw the item marker again which mess the screen. So I have to stop the framwork on resume. :(

hwhw:
This is great. Unfortunately, I'll be on the road until tomorrow - I hope to get finally back to coding again. Great work here. And overall.

As for Multithreading: I would really like to keep that out for now... Rather spawn a daemon that reports on a socket or a pipe when something relevant happens. We can catch that then in the select(). It's not critical to have a common memory space, so no need for threading. And such a daemon might prove to be useful for other projects, too.

dpavlin:
I'm OK with separate daemon, my main worry was leaving cvm in -stop state.

houqp:
A ha, I find a way to not stop the framework :)

Since the reason we stop the framework is it will redraw the item marker after resume from screen saver, we can do following to overcome it:

When we come back from screen saver, we can wait for 1 or 2 more seconds to let the framework draw the marker. Then we redraw current page page in unireader.

PR #92, opened 2012-04-08
fix time out crash in screen saver

houqp:
I modified `input.c` to return `nil`  on select timeout added a input.waitForEvent() wrapper in `key.lua` which handles nil return and keep waiting for the events.

I am wondering what is the best timeout value for select call? Current default timeout seems like about 4 minutes.

### DIFF
--- a/djvureader.lua
+++ b/djvureader.lua
@@ -292,7 +292,7 @@ function DJVUReader:startHighLightMode()
 
 	-- first use cursor to place start pos for highlight
 	while running do
-		local ev = input.waitForEvent()
+		local ev = waitForEvent()
 		ev.code = adjustKeyEvents(ev)
 		if ev.type == EV_KEY and ev.value == EVENT_VALUE_KEY_PRESS then
 			if ev.code == KEY_FW_LEFT then
@@ -538,7 +538,7 @@ function DJVUReader:startHighLightMode()
 	-- go into highlight mode
 	running = true
 	while running do
-		local ev = input.waitForEvent()
+		local ev = waitForEvent()
 		ev.code = adjustKeyEvents(ev)
 		if ev.type == EV_KEY and ev.value == EVENT_VALUE_KEY_PRESS then
 			if ev.code == KEY_FW_LEFT then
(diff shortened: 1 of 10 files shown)


hwhw:
In fact, an "endless" timeout would be possible. I don't think it is implemented now - but I could do it.

In the end, I'd like to have a flexible timeout, probably given in a global variable or at least according to viewer state. On my agenda, there is a "smooth scrolling" mode (using only black and white in a dithered blitting, which can reach up to 5-10 fps without going through black), using a very small timeout (and then continue to scroll/pan), then a bit larger timeout (after which the screen is refreshed in grayscale) and then in the end a potentially infinite timeout. Sooooo.... it depends.

houqp:
> In fact, an "endless" timeout would be possible. I don't think it is implemented now - but I could do it.

OH, then I think I should better leave that to you :) I am closing this request for now.

> In the end, I'd like to have a flexible timeout, probably given in a global variable or at least according to viewer state. On my agenda, there is a "smooth scrolling" mode (using only black and white in a dithered blitting, which can reach up to 5-10 fps without going through black), using a very small timeout (and then continue to scroll/pan), then a bit larger timeout (after which the screen is refreshed in grayscale) and then in the end a potentially infinite timeout. Sooooo.... it depends.

Sounds great. But I don't understand what is the reason for using a small timeout for "smooth scrolling" mode? Does the timeout affects the speed of user input?

hwhw:
> Sounds great. But I don't understand what is the reason for using a small
> timeout for "smooth scrolling" mode? Does the timeout affects the speed of
> user input?
> It would allow to trigger updates when the user holds down a key. Like on a
> PC keyboard, it is probably useful to have initial delay and then a
> repetition delay.

dpavlin:
It would also be useful to have some kind of chord support: since we don't have alt events on kindle keyboard I was thinking about implementing Aa which would toggle ALT flag somewhere which would be applied to next key. I haven't yet try it on device to see if this makes sense (e.g. do I really have down key event as opposed to key up which would work, but would require users to press key by key) but this could side-step our problem with missing shortcuts.

PR #95, opened 2012-04-09
show overlap when moving with pan_by_page

dpavlin:
This is nice feature of Sony e-book readers which is very useful when reading since it shows previous part of visible page dimmed (wash out) and provides hint where to continue reading after page refresh.

(here was my confusion about djvu -- it works without problem I was testing with file which had too much whitespace)

I'm not sure if `dimRect` in `blitbuffer.c` is the right way to go. This would make three functions which are similar (`paintRect`, `invertRect` and `dimRect`), but on the other hand, having separate functions does enable compiler to make better job (should I make pixel char instead of int?

What are you thoughts on this?


houqp:
Wow, I love that feature.

However, simply shifting the grey scale makes fonts look coarse. I think we need a better algorithm to dim the color.

hwhw:
in fact, we do need a font renderer that properly renders fonts upon grayscale backgrounds...

PR #104, opened 2012-04-12
API changes in highlighting code

houqp:
I found the naming of `UniReader:rectCoordTransform()` does not make sense at all and leads to confusion, so I split it into two methods: `UniReader:zoomedRectCoordTransform()` and `UniReader:getRectInScreen()`. I also added some more comments to explain these two methods, hope it helps.

Current highlight feature works very well for my PDFs and no breaks in DjVus, great work @hwhw :)

### DIFF
--- a/djvureader.lua
+++ b/djvureader.lua
@@ -54,14 +54,22 @@ function DJVUReader:_isEntireWordInScreenHeightRange(w)
 end
 
 -- y axel in djvulibre starts from bottom
+function DJVUReader:_isEntireLineInScreenHeightRange(l)
+	return	(l ~= nil) and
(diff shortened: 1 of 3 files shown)


dpavlin:
I pulled this changes and now I have error:

```
lua config error: ./unireader.lua:745: attempt to perform arithmetic on global 'tmp_l' (a nil value)
```

It's part of

``` lua
                        -- handle left end of first line as special case
                        tmp_l = math.min(tmp_l + 1, #t)
```

and wan't modified by last diff, but I don't understand enough code to fix it myself.

houqp:
I should close my request when fixing stuff ;P New request is sent.

dpavlin:
@houqp you are fast :-)

PR #107, opened 2012-04-13
show_overlap calculation bug in two column mode

houqp:
I suspect I found a bug in two column mode, though not sure.

When reading our two column mode test file, pan to the bottom of left column, then press one more five way key down, then press five way key right to go to next column, and you will get a -800 show_overlap, so the whole screen is dimmed.


dpavlin:
Good catch! I seem to notice that once, but I wasn't able to reproduce it.

Issue #103, opened 2012-04-11
PDF highlighting: engine now in unireader


hwhw:
I've just checked in PDF highlighting. Bad thing however is that it is still buggy. And I really hope I did not break djvu highlighting. A quick check showed it was still working. But please test extensively.

I still have to fully grasp the highlighting code. I found some rough edges that were DJVU specific and tried to flesh them out. But PDF highlighting still does not work correctly regarding zoom and some other things.

dpavlin:
Great work. It works nicely with pdfs which have real text in them, but when I tried our `test/2col.pdf` or `test/tall.pdf` I got:

```
Program received signal SIGSEGV, Segmentation fault.
getPageText (L=0x85ac058) at pdf.c:356
356                     linebbox = ptr->text[0].bbox; // start with sensible default
```

I'm not quite sure how to correctly check if that first element exists, so this is just a report :-)

houqp:
WOW, that's fast!

The segfault pointed out by @dpavlin is caused by reference of null pointer `ptr->text`, is this null pointer expected?

> But PDF highlighting still does not work correctly regarding zoom

Zoomming is not supported in DjVu neither ;P Current highlighting code should only work in FIT_TO_CONTENT_WIDTH mode, if it works in other mode, then we are just lucky. :)

If needed, we can discuss re-designing the structure for text table returned by getPageText method here. Following is a typical return from djvulibre's get text call:

``` lisp
 (line 912 1199 2525 1264
  (word 912 1200 1078 1264 "0,297")
  (word 1272 1200 1400 1264 "0,02")
  (word 1417 1211 1432 1263 "1")
  (word 1632 1199 1799 1263 "0,000")
  (word 1992 1199 2044 1263 "0,")
  (word 2060 1210 2161 1262 "157")
  (word 2357 1199 2525 1263 "0,266"))
 (line 1224 1095 2736 1163
  (word 1224 1095 1478 1162 "Anhang")
  (word 1501 1110 1564 1163 "61")
  (word 1580 1110 1593 1147 ":")
  (word 1625 1095 1974 1162 "Ergebmsse")
  (word 1996 1110 2104 1162 "der")
  (word 2129 1109 2382 1162 "Modells")
  (word 2401 1110 2514 1163 "M4")
  (word 2532 1096 2736 1162 "(cont.)"))
 (line 1000 1008 2980 1077
  (word 1000 1008 1593 1077 "(Signiikanzniveau:")
[...]

PR #108, opened 2012-04-13
make highlight working in all zoom modes #103

dpavlin:
This change introduce object's dest_x and dest_y coordinates which are needed in zoomedRectCoordTransform to make highlight scale and move according to current pan position.


dpavlin:
Sorry, it seem that I somehow merged this pull request myself :-)

hwhw:
doesn't matter, I tested and just wanted to do it now :-) great work!

PR #111, opened 2012-04-14
added reading of pan_margin settings

dpavlin:
requested at forum: http://www.mobileread.com/forums/showpost.php?p=2042643&postcount=238


dpavlin:
We should probably move margins to document settings when there is user interface to change them and leave reader settings as default

Issue #112, opened 2012-04-15
Widget framework


hwhw:
Hi,

to allow for easily creating Dialogs and Menus, I have started a Widget framework.
It's this commit: b6d75b84ecb315b9df15312e0f876bc0d571c56d

The commit message contains example code.

I think I like to migrate other dialogs over to use a Widget based interface when release is done.

I plan to add an implementation for message boxes and similar stuff.

The nice thing about it is that it can do alignment and size calculations.

Also, I added image handling. I reused mupdf code for that. It's commit 38afb59b3f5cf4e28c2afc95c13be5fd5222a28c
It is already integrated into the Widget framework so that there are now Text and Image widgets.

hwhw:
An example is the creation of the InfoMessage dialog. It is in dialog.lua, and a first use case is the error reporting when opening a document fails (in reader.lua).

houqp:
OMG, I love this one. I wish there is a '+1' button on github :P

dpavlin:
@hwhw +1 this is great

dpavlin:
I tried to add message to filesearcher and failed.

```
diff --git a/filesearcher.lua b/filesearcher.lua
index c042602..e8fcc45 100644
--- a/filesearcher.lua
+++ b/filesearcher.lua
@@ -2,6 +2,7 @@ require "rendertext"
 require "keys"
 require "graphics"
 require "font"
+require "dialog"

 FileSearcher = {
        -- title height
@@ -22,6 +23,7 @@ FileSearcher = {
 }

 function FileSearcher:readDir()
+       InfoMessage:show("Searching")
+       fb:refresh(0)
        self.dirs = {self.path}
        self.files = {}
        while #self.dirs ~= 0 do
```

And when I run it:

```
Program received signal SIGSEGV, Segmentation fault.
0x08063fb2 in fz_lock (lock=0, ctx=<optimized out>) at fitz/fitz.h:388
388             ctx->locks->lock(ctx->locks->user, lock);
```

What am I doing wrong? (aside from fact that it InfoMessage should go into setPath probably ;-)

dpavlin:
I'm getting same SIGSEGV if I try to add it to `crereader.lua` increase/decrease line spacing (which can be slow on device) so I'm **really** doing something wrong.

hwhw:
OK, this is probably something about the mupdfimg implementation I added. Does it work when you comment out the ImageWidget in the InfoMessage dialog?

[...]

PR #117, opened 2012-04-16
updated version of mupdf and some refactoring on help page

traycold:
hi guys, sorry i was quite busy last weeks.
Anyway, I've seen you implemented a lots of new features and improvements!
While I didn't have time to check all them out, this week-end finally I completed an activity I was working on some weeks ago: refactoring help page (hoping to improve it, and making it more clear) and update of mupdf to latest (at that moment!) revision (hoping kpdfview would benefit from it): this required some modification on the code; in particular yesterday I've seen that there were some changes regarding highlighting related code in mupdf that may be useful on kpdfview.
Honestly, I just tested that the build is ok, and the application starts on my pc (emulation); the pdf/djvu files are opened and rendered correctly; but I din't test all the features (like highlighting, for instance).
So I issued this pull request expecially to let you examine the differences and see if you find they could be useful (as I said, expecially regarding highlighting); but I think some tests of some features (expecially highlighing and CR-related ones) would be necessary before committing. I will do these tests these days, but maybe you can see at glance if something is wrong.


dpavlin:
When I try to enter hightlight mode I get following:

```
Program received signal SIGSEGV, Segmentation fault.
fz_lookup_text_style (ctx=0x86f9e18, sheet=0x0, text=<optimized out>, ctm=0xbfffeee8, colorspace=<optimized out>, color=<optimized out>, alpha=<optimized out>, stroke=<optimized out>)
    at fitz/dev_text.c:91
91              return fz_lookup_text_style_imp(ctx, sheet, size, font, wmode, 0);
```

traycold:
mmm... maybe the issue is caused by this line in `pdf.c`:

`tdev = fz_new_text_device(page->doc->context, NULL, text_page);`

the NULL is in fact where the text styles should be loaded (just like the text is loaded into variable `text_page`). I had a look to class `mudraw.c` of mupdf, where there is an usage example of text extraction, and there apparently NULL is used for styles; but may be something totally different; probably some further investigation will be necessary to fix it. I wil do it as soon as possible.

dpavlin:
Thanks. I also noticed that you removed Home (without modifiers) key return to file browser. Was this intentional? I think that home without modifiers should also return to file-browser, just like Amazon's framework does. See 53fcb2490bec193554707dff3e9046f011ab9b1b

I have a problem with zoom but It's fixed in b06927dcfed217798266d0931c7045d8bb0c9eeb 

Update: branch with few fixes on top of your changes: https://github.com/dpavlin/kindlepdfviewer/commits/traycold/master

traycold:
about Home command, you're right: it was not intentional. Thanks for pointing it out, and for other fixes.

dpavlin:
Update of mupdf also uncovered more issues with image support from mupdf #112. It seems that now image is totally wrong. It even changes from page to page :-)

hwhw:
I think an update of MuPDF is inevitable anyway. Since traycold invested time here already (thank you!), this is probably something we might need to get done before release. I'll look into the changes in the API. And maybe I can track down the image related bugs, too.

houqp:
Ha, good to see you back traycold :)

Just notice that image related bugs might be related to [this line](https://github.com/hwhw/kindlepdfviewer/commit/38afb59b3f5cf4e28c2afc95c13be5fd5222a28c#L3R127). Commenting out that line gets rid of the segfault. I am not familiar with the API and have no idea of what's wrong there.

dpavlin:
@houqp good gatch! From

```
  {"free", freeImage},
  {"__gc", freeImage},
```

I _could_ guess that lua calls again freeImage on gc and that segfaults. @hwhw would know better, I think.

hwhw:
yep, I think that's it. I'll check ASAP. Funny that I did not yet get segfaults...

houqp:
> yep, I think that's it. I'll check ASAP. Funny that I did not yet get segfaults...

LOL, you must did some magic on your toolchain ;P I am getting segfault nearly every time I fire up a menu or something. But it's is quite random.

dpavlin:
With 21a153cbc336ded35d9adf7e7a2990c53417f660 I can get `SIGSEGV` by pressing `B` and `T` fast every time in emulator :-)

rendertext.lua
```diff
@@ -44,10 +44,43 @@ function clearGlyphCache()
 	glyphcache = {}
 end
 
+function sizeUtf8Text(x, width, face, text, kerning)
```
houqp, line 47: what does the width mean here? does it mean the maximum end x offset for the text?

traycold, line 47: your guess is exact. In fact normally it should be the width of the display (`blit_buffer:getWidth()`).
In `renderUtf8Text`, where the blit_buffer is an input param,  this info is retrieved directly from that object; but since in current function there is no need of blit_buffer, I thought better to use as input param directly the maximum width that is considered valid; and remove blit_buffer as input param.

houqp, line 47: Ok, I see :)

Then I suggest adjust the naming of that variable. `width` might confuse people to think it means the width of the text. Name the (x,width) pair in the way you name (pen_y_top, pen_y_bottom) will make more sense I think. (Or (x0,x1))

traycold, line 47: absolutely right. 
I hope to dedicate some time to kpdfview this week-end, and I will fix this; or, of course, feel free to do it by yourself, no problem for me.
And I also should absolutely take the time to study all the new implementation / refactoring you guys made last 2 weeks: you work really hard to this project, and it's not easy to keep your pace! :-)

houqp, line 47: That's great. Looking forward to your patch :) I am trying to do line wrapping with help of this function.

traycold, line 47: yes, indeed that was my original idea when developing that function (applying it to the descriptions of the help page; but didn't had time to work to it, up to now); currently it's just used to create a background box around the keyborad keys of the shortcut, in order to better highlight them.
So if you find a way to implement line wripping in a generic way, we may reuse it in many context.

Issue #121, opened 2012-04-17
Regarding debug output wrapper


houqp:
I am inclined to use a debug output wrapper to print debug information instead of use print directly.

In this way, we can disable all these debug prints in release (users don't care about these output anyway) and even add a debug mode to redirect the debug log to a file.

dpavlin:
This was my original idea when I wrote dump, so I'm all for it :-)

houqp:
Ha, then will you take this task?

dpavlin:
I would love to have something like debug("message", var, " message ", var2, ...) which would check with type(var) and print tables using dump. Does this make sense?

houqp:
looks yummy. This issue has been assigned to you ;)

houqp:
Thanks @dpavlin , closing for now.

PR #123, opened 2012-04-17
Issue 121 debug

dpavlin:
This is huge patch for #121 with very little functionality.

I'm not completely sure if new style of debug messages is better than old one, so comments welcomed.

### DIFF
--- a/settings.lua
+++ b/settings.lua
@@ -27,12 +27,12 @@ function debug(...)
 	local line = ""
 	for i,v in ipairs(arg) do
 		if type(v) == "table" then
-			line = line .. dump(v)
+			line = line .. " " .. dump(v)
 		else
-			line = line .. tostring(v)
+			line = line .. " " .. tostring(v)
 		end
 	end
-	print("# "..line)
+	print("#"..line)
 end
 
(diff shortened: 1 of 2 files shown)


hwhw:
I think we could have use for a "quiet" switch that turns the debug function silent. Or better yet, a loglevel global and an annotation when calling, e.g.

LOG_DEBUG=1
LOG_INFO=2
LOG_WARN=3
LOG_ERR=4
LOG_THIS_CANNOT_ACTUALLY_HAPPEN=5

and then calling

debug(LOG_DEBUG, ...)

and in the function a check for the loglevel.

dpavlin:
To be honest, I would prefer separate functions for `debug()` (which can be quiet) and `error()` and info is usually `InfoMessage` anyway.

To me it makes code easier to read than debug with magic flags (which you can type wrong, and without any checking from lua side, I would prefer syntax error).

We could also return to practice of different debugging level by prefixing messages with variable number of `#` symbols which `| vi -R -` nicely color code :-)

hwhw:
Very good points. I wholeheartedly agree!

PR #143, opened 2012-04-26
add rounded corner border

houqp:
Just for fun...


dpavlin:
I merged it, but it's not used, right?

houqp:
Yes, as I said, just for fun ;P

Maybe it can be used to draw buttons.

dpavlin:
That's a good idea, I don't particularly like current blocks which are should have rounded corners :-)

Issue #142, opened 2012-04-25
incessant cursor movement while pressing and holding a five-way key


eLiNK2gl:
Hi,

this issues was originally mentioned at http://www.mobileread.com/forums/showpost.php?p=1879699&postcount=17.
It could be used in the file chooser, in the highlighting mode and in the input box.

Doable?

hwhw:
Yes, probably. I'll experiment with it a bit.

dpavlin:
Device and emulator do generate all three event types:

``` lua
-- event values
EVENT_VALUE_KEY_PRESS = 1
EVENT_VALUE_KEY_REPEAT = 2
EVENT_VALUE_KEY_RELEASE = 0
```

So simple change from `ev.value == EVENT_VALUE_KEY_PRESS` to `ev.value ~= EVENT_VALUE_KEY_RELEASE` in `UniReader:inputLoop` generate repeated events for all keys (including fiveway), but I'm afraid that repeat rate is a little bit too fast.

However, it's nice effect to see how fast we can refresh e-ink screen, and it does scroll pretty fast :-)

hwhw:
Funny, I did insert them but didn't bother playing with them. I think an easy filter on a certain number of EVENT_VALUE_KEY_REPEATs would be easy to implement. I'll give this a fast try before releasing.

dpavlin:
OOH, fast refresh is nice combined with partial screen refresh to browse over pages... It's closest possible feature to flipping real book, and it looks nice :-)

hwhw:
luckily, I got distracted :-P so we didn't do duplicate work (and you did it all...). great! I'll be home again in about half an hour and will test. I think this could go into release, couldn't it?

dpavlin:
sure :-)

dpavlin:
I didn't put it in `helppage.lua` and `selectmenu.lua` because I don't think that it makes sense.

hwhw:
I'm not so sure with selectmenu, but for now I'm fine with leaving this as-is, since selectmenu isn't used nearly as often as the other parts you cared for already.

PR #146, opened 2012-04-27
Issue #142 key repeat

dpavlin:
This also disables full screen refresh if key is repeating which is nice for fast flipping through pages on device.

On other hand, I noticed that SDL isn't generating repeated keys events for me.


dpavlin:
Feel free to ignore this pull request if it conflicts with your code @hwhw, it seems we are again working on same issue #142 :-)

PR #151, opened 2012-04-29
manually select page bounding box

### LINKED ISSUE #35: implement manual page crop
Some pdf files have bounding boxes totally wrong. djvu doesn't have it at all. Before I started using this reader, I was cropping pdfs manually to provide better experience, but now I can code that in lua :-)

Idea is to provide manual margins (with option to specify different ones for odd and even pages) and use that instead of page:getUsedBBox values. This would make two-column mode and all fit-to-content zoom levels work automatically.

To implement it, I would need a way to draw lines over rendered pdf page. Something like xor 0xff might be perfect, since I could just re-draw line if I move it. I did a quick look in blitbuffer.c, and it seems to me that I need paintRect which will do XOR with value, so I will give my rusty C knowledge a spin.

dpavlin:
I hope we can finally close #35 with this.


houqp:
Great job, this looks awesome!

PR #152, opened 2012-04-29
half done menu widget for new ui framework

houqp:
I will finish it tomorrow, but I think it is better to send the pull request now because I did some other modifications in the code base.
- I changed the parameters for TextWidget's bb size calculation to draw better icons in menu items. But I am not sure whether my modification is correct.

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -20,6 +20,20 @@
 #include <string.h>
 #include "blitbuffer.h"
 
+
+inline int setPixel(BlitBuffer *bb, int x, int y, int c) {
+	uint8_t *dstptr = (uint8_t*)(bb->data) + (y * bb->pitch) + (x / 2);
+
+	if(x % 2 == 0) {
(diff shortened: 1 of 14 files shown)


hwhw:
As for the local shortcuts: widgets are encapsulated into "outer widgets", and those "child" widgets can consume the events before they are propagated to the outer widgets. So if a "inner" widget defines special behaviour for key events, it can "consume" them, not allowing them to be handled by "outer" widgets. This is how the FocusManager works.

On the other hand, as for "opting out" to handle certain events, I would suggest to use the event mechanism for that: some widgets implement onEventInQuestion(), some don't.

hwhw:
(although, now I'm thinking about this, this will make displaying keyboard shortcuts on a help page more difficult, so another solution might be in order)

houqp:
I replied this in a line notes in commit: 08278d6beb3df14ff69d7078feb05ba063429198 :)

PR #153, opened 2012-04-29
second try to implement ruler idea by @eLiNK2gl in #35

dpavlin:


houqp:
That's really impressive!

eLiNK2gl:
should not the lines in unireader.lua

InfoMessage:show("new page bbox")
fb:refresh(1)

be better replaced with
showInfoMsgWithDelay("new page bbox", 2000, 1)
?

houqp:
Agree, I would recommend something like:

``` lua
showInfoMsgWithDelay("new page bbox", 2000, 1)
self:redrawCurrentPage()
```

eLiNK2gl:
How exactly was it supposed to work?
I mean if I "draw" the bounding box on page 3, is it to be applied to pages 3 to _last_ only? To even/odd pages?

There are some bugs, although I'm not sure where the problem lies as I have not been able to reproduce them.
What I came across so far:
1) sometimes the bounding box gets recognized correctly (automatically) with every second page only BUT after some manipulations with the file the bounding boxes get recognized correctly on all pages
2) on occasions zooming in (shift + page fwd) may change the position of the bounding box on a single page only.

I have been using the mixture of the last release and Nupogodi's patches.

Getting the settings back to default would be a great help in locating the problem.
Maybe through removing the file (not deleting the file) from the list of recently read books, the one which shows up on pressing 'L'.
If I understand correctly, all the settings are to be found under kpdfviewer/history and removing that lua file would do the trick, right?

dpavlin:
We should really update help page descriptions a bit it seems :-)

When using `Z` or `Shift+X` you will create bounding box which will be selected for current page and all following odd/even pages.

If you have just one custom bbox, it will apply from this page forward (both for odd and even pages), but when you have different odd and even bboxes they will apply correctly.

You can delete current page bbox using `Shift+Z` but this will leave odd/even bboxes (this is useful if you want to setup bboxes for odd/even pages, but you would like to do it on page 3 and use it for all odd pages for example).

If you are running with -d we will dump bboxes output, so you can inspect which custom bboxes are defined in current document. I hope this helps a little :-)

PR #155, opened 2012-04-30
add paintRoundedRect & use it in FrameContainer

houqp:


dpavlin:
Looks very nice. I will have to start working on new_ui_code branch, it seems :-)

PR #160, opened 2012-05-02
more changes in new_ui_code

houqp:
- use rendertext in textwidget so backgroud for text is supported
- more styles for menu item icons
- bug fix

### DIFF
--- a/widget.lua
+++ b/widget.lua
@@ -164,27 +164,36 @@ TextWidget = Widget:new{
 	color = 15,
 	_bb = nil,
 	_length = 0,
+	_height = 0,
 	_maxlength = 1200,
 }
 
-function TextWidget:_render()
-	local h = self.face.size * 1.3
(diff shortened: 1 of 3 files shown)


widget.lua
```diff
@@ -164,27 +164,36 @@ TextWidget = Widget:new{
...
-		self:_render()
-	end
-	return { w = self._length, h = self._bb:getHeight() }
+	--if not self._bb then
+		--self:_render()
+	--end
+	--return { w = self._length, h = self._bb:getHeight() }
+	
+	self._length = sizeUtf8Text(0, G_width, self.face, self.text, true).x
+	self._height = self.face.size * 1.3
```
dpavlin, line 184: Where does 1.3 comes from? Is that conversion between points and pixels or heuristics?

houqp, line 184: Just heuristic... Also with the 0.8

Issue #166, opened 2012-05-14
NuPogodi patch: How to integrate?


hwhw:
Hi,

here @NuPogodi published a patch: http://www.mobileread.com/forums/showpost.php?p=2080025&postcount=380
It has some nice things, however, it's not easy to separate the new functionality and I'm having a hard time figuring out how to integrate it... Ideas?

houqp:
Um... maybe we can create an other branch for his patched version so we can do some reviewing and gradually integrate the changes. It will be a lot better if we can persuade him to join us in github :)

dpavlin:
I tried to checkout version on which this patch is based and copy all changed files over to create branch with changes. However, since most of code also does some re-indenting this created large and wrong diff. Then I tried to make diff -w (which should at least ignore white-space changes) but that patch failed to git apply (although it was generated by git itself).

PR #167, opened 2012-05-16
first cut at implementing cleanup of NuPogodi patch

dpavlin:
This is my try at #166


houqp:
Just notice that NuPogodi has started to push his changes to github: https://github.com/NuPogodi. Should we invite him to work on this pull request?

dpavlin:
I would just ignore this, since it would be nicer to have changes in smaller commits, so @NuPogodi work will be better then one.

PR #177, opened 2012-05-27
First try on Toc Menu

houqp:
Not sure whether this is the right way to add Toc feature, if I did something wrong, please point it out :)


hwhw:
Yes, this is somewhat how I figured it could work. Great!

PR #178, opened 2012-05-28
NuPogodi code drop from 2012-05-27

dpavlin:
- zip files don't require double extensions any more
- filechooser has 'H' shortcut to display keys
- new function for files and folders (^ == shift)
  ^N - new folder
  Del - delete document/folder
  ^C - copy file into clipboard
  ^X - cut (move) file to clipboard
  ^V - paste files from clipboard
  ^R - rename file
- configurable keyboard for input box

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -9,15 +9,15 @@ CREReader = UniReader:new{
 
 	gamma_index = 15,
 	font_face = nil,
-	-- NuPogodi, 17.05.12: to store fontsize changes
 	font_zoom = 0,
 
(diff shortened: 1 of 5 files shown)


dpavlin:
clipboard uses rename to move files around, so you can't really use it in emulator when your documents are on different filesystem than source (this won't happen on device)

hwhw:
OK, I will merge and test this in 2 hours. If everything works as expected, I'll release... Feel free to merge it right now, I'm just AFK for the next 2 hours.

PR #188, opened 2012-06-03
djvu and crengine support for readerui

houqp:
Still I am not sure whether moving all the duplicate code to document.lua is the right way to go.


hwhw:
I just skimmed through the code and it looks very good! I like the readerrolling :-) I think I will have time on the coming weekend to dive into new_ui_code again. Merging this now to not let it get too far behind...

hwhw:
OK, github doesn't let me merge right now. I'll sort it out manually, but it will take about two days until I find time to do that.

houqp:
Glad that I hack in the right way :)

We must hurry, there are lots of conflicts between master and new_ui_code branch now...

houqp:
Um... I just merged it... Is it OK?

PR #191, opened 2012-06-06
three changes

houqp:
- bump crengine version
  remember to do git submocule update :)
- adjustment in bookmark feature
- new getPosFromXPointer method in cre module


hwhw:
yay, I think CREngine made performance improvements :-)

PR #195, opened 2012-06-07
add usb charging events

houqp:
I found that the USB plugin/out events have to be watched with a different command:

``` bash
    lipc-wait-event com.lab126.hal usbPlugOut,usbPlugIn
```

Any idea of how to integrate it into the fake event generator? Fork another child process?


dpavlin:
If I'm not mistaken lipc-wait-event is thin layer on top of dbus. Can we use dbus bindings for lua to side-step this issue? https://github.com/luaforge/lua-dbus

houqp:
Yes, but I don't know how to interact with kindle's dbus system, I even failed to list all their interfaces :(

It will be a lot better if we can communicate with the dbus system directly.

dpavlin:
I found one example of dbus usage with Kindle, unfortunately it seems that blog is in Chinese which is problem for me :-) https://github.com/tjwei/KindleChewing Does this help at all?

houqp:
OMG,  thanks @dpavlin, that's a awesome link :) I will take a closer look at it when I have time!

PR #198, opened 2012-06-09
remove dependency on <linux/fb.h>

tchaikov:
this changeset completely decouples the SDL front-end from its eink counterpart. hopefully it can ease the pain of potential developers without much cross-compiling experiences.


hwhw:
Thank you! That's nice :-)

PR #199, opened 2012-06-10
K4 framebuffer support

houqp:


hwhw:
pretty cool.

hwhw:
I think I will eventually change the behaviour, though. I'd like to use a 4bpp blitbuffer in between rather than using the 8bpp framebuffer as scratch space for 4bpp data and then doing the conversion.

houqp:
Go ahead :) I can help with the testing work.

hwhw:
I think it can wait for now. There are probably much more important things. In fact, it might even be very efficient how it currently is - at least in terms of memory used :-)

PR #202, opened 2012-06-10
fixes for mac

tchaikov:
- per http://luajit.org/install.html, otherwise LuaJIT always fails
  to get desired memory chunk from mmap(2)
- appeases clang a little bit


hwhw:
Now wait, you say it compiles on Mac OSX now? Whoa. Great :-)

tchaikov:
yes, i was fighting with the arm toolchain yesterday on my mac, but cowardly resorted to build the emulator ... =)

PR #203, opened 2012-06-11
add sub menu support for menu widget

houqp:


hwhw:
Ah, that's nice.

However, maybe we need to adapt the menu handling a bit, I'm still not too sure. I think the other parts of the reader should have the possibility to "register" menu entries with the main menu handler. So the instance of ReaderMenu we use in ReaderUI will probably need to have a known name. The definition of the menus - except maybe for global menus like "settings" or "quit" - will have to go into the functionality-specific submodules. But I think this can grow while we're on the way.

The Menu should probably also be extended to support arbitrary widgets as items. On the other hand, it might be too much for the menu implementation (variable size items vs. fixed size items would affect the paging a lot)...

houqp:
Agree with both of your points :) We still have a long way to go, especially for the arbitrary widget item menu...

Maybe a little bit OT here. I think the ReaderMenu, as well as the FileChooser, should be made into a widget that contains menu widget. That way we can add more contents to them, like the progress bar or battery/datetime information. Now they are just a sing menu...

PR #223, opened 2012-07-06
screen.lua

ghost:
request to update screen.lua only: added the functions to save the fb-content in common graphic files - bmp (default) & pgm (ATM, not used)
The rest commits are outdated.

### DIFF
--- a/commands.lua
+++ b/commands.lua
@@ -56,7 +56,7 @@ function Command:new(keydef, func, help, keygroup, order)
 	obj.help = help
 	obj.keygroup = keygroup
 	obj.order = order
-	--debug("creating command: ["..tostring(keydef).."] keygroup:["..(keygroup or "").."] help:"..help)
+	--Debug("creating command: ["..tostring(keydef).."] keygroup:["..(keygroup or "").."] help:"..help)
 	return obj
 end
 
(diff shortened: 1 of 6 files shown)


houqp:
Aha, finally we have a usable screenshots feature :)

Seems that @NuPogodi needs to first do a merge with current master in his own repo? Some code from these patches have already been applied before and thus leads to a conflicts...

ghost:
I've already mentioned that github is too complicated for me, each time I press a button "Submit" I always expect the worse... And the expected shit sometimes does happen. 
As concerns to the screenshot feature, I may say the following. I've generally tried to convert the framebuffer by two methods: 1) with lua only and 2) by making the standalone converter (I've wrote the c-code and compiled it with tiny c compiler) and calling it from lua-code by os.execute()
Recently, I've also added the lua-function that converts framebuffer to tiff - it is the fastest method among the lua functions, it takes only ~0.04s at K3 (that is already comparable to standalone converter). 
If someone is interested, the updated files may be found here http://rghost.net/39214721 (tested only at 4bpp (K2, K3, KDX)

dpavlin:
@NuPogodi do we now have all your changes in master? If so, you will want to full current master in your repository.

Since they diverged quite a lot, you can replace it with:

git checkout -b old-master
git branch -D master
git fetch origin
git checkout origin/master

Next time you push to github, it't won't pass without -f flag:

git push origin master -f

If you want to also have local branch old-master on github, do

git push origin old-master

and it will be visible. Hope this helps :-)

PR #244, opened 2012-09-02
Minor optimization of DjVu drawPage() and set dither bits to 4 to match Kindle's screen.

tigran123:
1. There is no need to create and destroy djvu pixel format on each redraw of the page as this can be done once on open and close of the document. 
2. Set dither bits to 4 to help djvulibre choose the most optimal dithering algorithm for the Kindle.
3. Make coding style of "if(" -> "if (" consistent (both instances were used, but "if (" was more frequent).
4. No need to set gamma value in djvu pixelformat as we handle it "manually" in drawPage().

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,6 +207,8 @@ thirdparty: $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) $(DJVULIBS) $(CRENGINELIBS)
 
 INSTALL_DIR=kindlepdfviewer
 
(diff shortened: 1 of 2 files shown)


tigran123:
I added a trivial change to Makefile for the customupdate target to avoid shipping garbage *.lua files in releases.

tigran123:
The setting of dithering bits commented out as I checked inside djvulibre and see that ignores anything less than 8 bits anyway.

PR #246, opened 2012-09-03
Bugfix: screen refresh needed on exit from zoom mode menu.

tigran123:
We need to redraw the screen on any return from the zoom mode menu selection, otherwise the user gets the impression that the program hangs (I just got one such report from eLINK2gl) if the user selects the mode matching (or "effectively" matching) the current one.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -2123,11 +2123,11 @@ function UniReader:addAllCommands()
 				current_entry = - unireader.globalzoom_mode
 				}
 			local re = zoom_menu:choose(0, G_height)
-			if not re or re==1 or re==8 or re==9 then -- if not proper zoom-mode
-				unireader:redrawCurrentPage()
-			else
+			if re and re ~= 1 and re ~= 8 and re ~= 9 then -- if proper zoom-mode
+				Debug("setting zoom mode re=" .. re)
 				unireader:setglobalzoom_mode(1-re)
 			end
+			unireader:redrawCurrentPage()
 		end)
 	-- to leave or to erase 8 hotkeys switching zoom-mode directly?
 
--- a/unireader.lua
+++ b/unireader.lua
@@ -2124,7 +2124,6 @@ function UniReader:addAllCommands()
 				}
 			local re = zoom_menu:choose(0, G_height)
 			if re and re ~= 1 and re ~= 8 and re ~= 9 then -- if proper zoom-mode
-				Debug("setting zoom mode re=" .. re)
 				unireader:setglobalzoom_mode(1-re)
 			end
 			unireader:redrawCurrentPage()


tigran123:
@NuPogodi  Yes, sure. Just open any djvu file and then press "M" to call up the zoom mode menu. Now select "W" for "Fit zoom to page" and press FW selector. All is fine so far. Now press "M" again and select "W" again (which will already be current) --- now observe the program "to hang" (prior to my fix). In actual fact it does not hang, of course. It simply fails to refresh the screen. Very easy to reproduce.

ghost:
I see... I just missed the most stupid case when user selects the same mode as the active one instead of the exit by pressing 'Back'. Well, a proper solution for is to add this condition in the following way
if not re or re==(1-unireader.globalzoom_mode) or re==1 or re==8 or re==9 then ...

Thanks for report, but I donot like your fix. What you have done, namely

if re and re ~= 1 and re ~= 8 and re ~= 9 then -- if proper zoom-mode
        Debug("setting zoom mode re=" .. re)
         unireader:setglobalzoom_mode(1-re)
       end
unireader:redrawCurrentPage()

in 7 from 11 case (~63%) requires the reader to draw the current page TWICE - once, with unireader:setglobalzoom_mode(1-re) and then again, with unireader:redrawCurrentPage().

So, I will include a fix to your fix in the next pull request.

tigran123:
Yes, yes, I realized that _immediately_ after commenting to your reply, but my daughter was nagging me to go to the swimming pool and I didn't bother to switch on the computer to correct myself. And I realized the proper fix also.

Now, I'll read your comment (I am sure you noticed exactly the same thing, so I am posting the above even _before_ reading your comment :)

tigran123:
Yes, the proper fix would be to add to the blocking test the condition to check if return from choose() is the same as the (negated) current zoom mode.

Ok, you go ahead and correct my fix, since it was your original code to begin with anyway :)

tigran123:
Ah, I've just looked at your code --- I forgot the "+1" as well --- i.e. by "negated" I meant "negated +1" = "1 - unireader.globalzoom_mode"!

tigran123:
@NuPogodi Doesn't InfoMessage:show() about setting the zoom mode cause an extra screen redraw as well? (well, depending on refresh_mode it may be a partial one).

I would suggest commenting out that InfoMessage:show() (as you also suggest it yourself in the comments).

ghost:
Ok, this week I'll try to focus on user-configurable font faces -- not as it was made before (i.e. for the "cfont" only), but for every item in Font.fontmap. Fortunately, the reader already able to store these fontfaces as global settings, so I only need to add the GUI-based possibility for user to change these... I hope to finish this feature till this Friday-Saturday so that since 10/09 I could light-heartedly enjoy my two-week vacations without work, internet & programming.
Best regards.

tigran123:
Ok, and I am currently focusing on the hierarchical TOC thing (issue #255) --- I had no idea (and that after 25+ years of C programming!) that one can do such wonderful things with tables in Lua... Initially I thought the solution in terms of n-ary trees, but now I begin to "feel" these "Lua table" beasts deeper and deeper and see that I don't even need trees as such. Anyway, I am sure that you, @NuPogodi could probably solve the whole thing in 10 minutes and I am grateful for allowing me to do it by myself, slowly, while learning Lua at my own [very slow] pace and enjoying it thoroughly :)

houqp:
It's great to have two more active hackers for KPV :) 

It is likely that I will be buried under some real day work for the coming 3 moths. But when I have time, I will work on the new ui branch so K4 user can enjoy KPV too.

eLiNK2gl:
@houqp Is there an easy way to copy your issues (https://github.com/houqp/kindlepdfviewer/issues) into the main repo?

houqp:
@eLiNK2gl , I am afraid there is not...

eLiNK2gl:
@houqp, Then, if it's OK I'll copy it one by one later (probably next week).

Issue #59, opened 2012-03-10
Bug in panning with alt key


houqp:
Seems that alt key is very different from shift key. When you press alt key alone, it will not generate alt key press event. And when you press alt key with following keys, it will not generate alt key events too:
- Q-P (only numbers, not combinations of alt+char)
- fiveway keys (only fiveway key events, no alt key event)

This leads to a bug in panning. When we use alt + fiveway to pan, it should move in 20 pixels. It actually does in EMU mode, but in real kindle, it still moves 100 pixels.

dpavlin:
Good catch!

My first thought was to suggest Q-P keys as configurable steps for panning, but then I tried that on Kindle and it doesn't seem intuitive. After pressing a few more keys here is one possible alternative to pan step change:
1. we record last fiveway direction
2. if next event is from fiveway with shift in _same_ direction we _increase_ step to double size and move
3. if next event is from fiveway with shift in _opposite_ direction we _decrease_ step to half size and move

This also allows easy fix of "oh, I scrolled too much" errors so it might be usable. What do you think?

p.s. I think that same bindings might be good for moving page margins in manual crop #35

houqp:
I found this bug when I was trying to manual crop my book in djvu. And I found it some how impossible to move the main content to the center of the screen just with 100 pixels and 50 pixels move. When I want to try 20 pixels with alt key, the bug jumped out. I guess 20 pixels wont help in some situations too.

> ```
> we record last fiveway direction
> if next event is from fiveway with shift in same direction we increase step to double size and move
> if next event is from fiveway with shift in opposite direction we decrease step to half size and move
> ```
> 
> This also allows easy fix of "oh, I scrolled too much" errors so it might be usable. What do you think?

This do solve the problem of "oh, I scrolled too much", but I am afraid decreasing step to half is not enough as I mentioned above.

> My first thought was to suggest Q-P keys as configurable steps for panning, but then I tried that on Kindle and it doesn't seem intuitive. After pressing a few more keys here is one possible alternative to pan step change:

I think that looks good. To make it intuitive, we can pop up a input box and just let users specify the step they want. How do you think about that?

kljohann:
However on 2012.03 if I use Alt+> or Alt+< first and then press Alt+five way it works correctly. I have not looked at the source though so this is only what I'm experiencing.

houqp:
[...]

PR #247, opened 2012-09-03
unireader: storing rcountmax; user-configurable zoom step & pan steps, etc

ghost:
1. In order to close issue #59, I've introduced user-configurable values for panning steps - for X- & Y-axis, separately. Now panning is called by pressing fiveway without mod_keys. When one presses Shift+FW, he/she changes panning steps in the following way - Shift-Left/Right = decrease / increase X-panning step (divide or multiply on factor 2), Shift-Up/Down = increase / decrease Y-panning step. The lowest value for panning steps is set 1. Both parameters, shift_x & shift_y, are saved to local settings (i.e. history file).
2. Manual zoom step was also made user-configurable. So, instead of two old zooming functions (with fixed 10% and 20%-steps), I made one function to zoom (Shift+<>), while the other (Alt+<>) changes the zoom step downto minimum 1%. 
3. Fresh introduced function to change parameter rcountmax was a bit improved by storing rcountmax to both setting files - global settings & local ones (i.e. history file) - so that the rcountmax-priority (from low to high) is the following: default rcountmax=5 < its value in global settings < rcountmax stored for each concrete document.
4. Dirty, the extention-based hack to avoid reading not crengine-related parameters stored in history files. TODO: one has to finally introduce loadSpecialSettings() & readSpecialSettings() for PDFReader & DJVUReader and to store therein the reader-specific parameters (like globalgamma, bbox, globalzoom, globalzoom_mode; render_mode = for djvu-files only, etc.)
5. Some lua-code cosmetics (to make code more readable); for example: one-line version "self.pan_margin = settings:readSetting("pan_margin") or self.pan_margin" instead of the same
   "local pan_margin = settings:readSetting("pan_margin")
   if pan_margin then
       self.pan_margin = pan_margin
   end"
   in 4 lines.
6. Fix for the today's tigran123 fix: issue #246 (details are included therein). I've also added InfoMessage:show() to inform most impatient users that the reader tries to redraw new zoo mode.
7. The fontface for the reading info (called by 'Menu') is restored to be in accordance with original intentions and with the same fontface as in crereader.lua - namely, Font:getFace("rifont", 20)
8. Moved the exit hotkeys (Alt+Back & Home) to the end of hotkey list.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -56,6 +56,8 @@ UniReader = {
 	-- set panning distance
 	shift_x = 100,
 	shift_y = 50,
+	-- step to change zoom manually, default = 16%
+	step_manual_zoom = 16,
 	pan_by_page = false, -- using shift_[xy] or width/height
 	pan_x = 0, -- top-left offset of page when pan activated
 	pan_y = 0,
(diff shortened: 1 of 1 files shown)


ghost:
Dear colleagues,

please, test whether everything I've submitted today works properly.

Yet one has also to discuss, let me say, somewhat "strategic" question - how to store the reader-specific parameters. I've just noticed that unireader saves a lot of own parameters to history files that actually have no direct relation to other readers. Say, it has no sense to store globalgamma, bbox, globalzoom, globalzoom_mode, etc in history files for cre-documents; the parameter render_mode is related to djvu-docs only and so on...

ATM, I see two approches:
1. Since functions to save certain parameter are already the reader-related, one may store each parameter just after its changing (now I use this simple method to store the zoom- & pan-steps)
2. Imho, more suitable looks the way that was initially introduced in crereader.lua - namely, adding readSpecificSettings & saveSpecificSettings to PDFreader & DJVUreader.
   In both cases, reader will save only necessary parameters to history files. In this way one may increase (at least, just a little bit) a speed of opening the documents [and, hence, make my dirty, the extention-based hack in function UniReader:loadSettings(filename) not necessary]

What do you think?

tigran123:
I completely agree that render_mode should go into djvu special settings. (and likewise the other format-specific settings, I just speak for djvu as I can do it myself safely).

Also, as I mentioned on #246 I suggest to comment out the InfoMessage:show() about setting the zoom mode, as unnecessary waste of cpu cycles and battery.

ghost:
1. So, please, look at the code of UniReader:inputLoop() where the parameters are saved before exit. To my opinion, only 4 first parameters
               self:saveLastPageOrPos()
       self.settings:saveSetting("jump_history", self.jump_history)
       self.settings:saveSetting("bookmarks", self.bookmarks)
       self.settings:saveSetting("highlight", self.highlight)
   are not reader-specific. The rest (including the commented parameters -- like shift_x, manual_zoom_step, etc. -- that are stored now just after their change by user) should be included to DJVUReader:saveSpecificSettings() & DJVUReader:readSpecificSettings() 
2. Concerning InfoMessage about setting the zoom mode... I did it just to avoid misunderstanding, since some new zoommode redrawing may take upto 2 seconds. The message costs definitely much less CPU- & battery resources as redrawing the page in new zoommode itself. But, please, feel you free to remove it, if needed.

tigran123:
I see, so that even though, say highlight, is not djvu-specific it should go to djvu AND pdf special settings, because it is not applicable to cre. Ok, I understand.

But if in the future one implements highlight for cre (hmmm, didn't @dpavlin already commit something related to it recently?) then we will have the new task of moving things in the opposite direction, i.e. out of reader-specific special settings back into unireader.lua.

I am not at all saying that we shouldn't do what you suggest. No, I am just saying we should carefully consider which settings are likely to become truly reader-independent in the near future before moving them to the reader-specific domains.

tigran123:
@NuPogodi Is there a way to test your changes before your pull request is applied? I mean other than by reading the commit and manually applying to my tree, of course. I.e. is there a git command to apply your changes to my local copy of the master tree?

ghost:
1. I'm definitely not an expert in the git-related questions. But you may anytime just copy my commits and test...
2. The delay before applying any of our pull requests depends on a good will of @hwhw, @dpavlin & @houqp

dpavlin:
There is a way to add additional remote repository for example for @NuPogodi:

```
$ git remote add nupogodi git://github.com/NuPogodi/kindlepdfviewer.git
$ git fetch nupogodi
$ git checkout -b nupogodi/master nupogodi/master
Branch nupogodi/master set up to track remote branch master from nupogodi.
Switched to a new branch 'nupogodi/master'
```

tigran123:
@dpavlin Thank you, that is exactly what I was looking for.

houqp:
Great that we finally come up with a solution to #59 :) But I suggest adjust the key binding a little bit. Say when zoomming, `Alt` is used to change the zoom steps, while for panning, `Shift` key is used to change the pan steps. Shouldn't we use `Shift` to change steps in both panning and zoomming? So users are less likely to mess the bindings up.

PR #250, opened 2012-09-05
fix: to prevent crashes at opening wrong zips or zips with not crengine-related documents

ghost:
1. crereader.lua: function CREReader:open(filename) now tests (i) whether the zip to be opened is really properly packed zip and (ii) whether the first zip entry is the crengine-related document. If some of these conditions is not satisfied, it returns false (as well as the error description) back to function openFile(filename) in reader.lua  
2. reader.lua: function openFile(filename) shows warning with the error description returned by CREReader:open(filename).

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -34,25 +34,36 @@ function CREReader:init()
 end
 -- inspect the zipfile content
 function CREReader:ZipContentExt(fname)
-	local outfile = "./data/zip_content"
-	local s = ""
-	os.execute("unzip ".."-l \""..fname.."\" > "..outfile)
+	local outfile, s = "./data/zip_content"
+	os.execute("unzip -l \""..fname.."\" > "..outfile)
 	local i = 1
 	if io.open(outfile,"r") then
 		for lines in io.lines(outfile) do 
 			if i == 4 then s = lines break else i = i + 1 end
 		end
 	end
-	-- return the extention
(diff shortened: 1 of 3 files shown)


tigran123:
@NuPogodi  While you are at it, do you mind also fixing "G" command to round the page to whole number, like this:

`if not pcall(function () page = math.floor(page) + 0 end) then...`

otherwise, currently it allows non-whole page numbers (like 7.8) which appear both as the current value via "G" and also in the footer of "Menu" "Page 7.8 or 500". I didn't bother opening a separate issue for it and I don't want to put it in my big commit for multilevel TOC changes. I noticed it because I introduced a new user-configurable variable toc_maxdepth and copied your code for setting the page number and while testing it noticed that it accepts non-whole numbers.

Btw, maybe you (or someone) cares to explain what the Lua idiom "x = x + 0" means? It was mentioned somewhere in the Lua book, but I forgot what it was...

ghost:
1. "+" gives Lua to know that both the additives & the sum are digits.
2. Another way to do the user-input protection is just selecting first sequence of chars 0-9 from the string

if not pcall(function () page = string.match(page, "%d+") end) or page < 1 or page > unireader.doc:getPages() then
      page = unireader.pageno
end

tigran123:
@NuPogodi  Thank you.

PR #251, opened 2012-09-06
New menu to change font faces; new parameter to handle files; etc

ghost:
1. Added menu to select proper font faces Font:chooseFonts() for not only "cfont", but for every item in Font.fontmap. Outdated version is replaced in all lua-modules. There are also minor font-related corrections... 
2. Added global parameter that allows FileChooser to work either in 'safe' or in 'advanced' mode. Added the mode selection menu (Alt+M). Since 'advanced mode' allows to feed CREReader by any shit (sorry, by improper documents), it may result to crashes (say, all tested *.azw2 'successfully' crashed crengine).
3. To avoid duplication, the function 'make screenshot' is moved to commands.lua.
4. The rest are imho not too important changes... the details are described for each submitted lua-module.
   WARNING: please, test these files thoroughly before merging this pull request and, of cause, feel you free to modify everything what you want.

### DIFF
--- a/commands.lua
+++ b/commands.lua
@@ -222,5 +222,12 @@ function Commands:new(obj)
 			G_charging_mode = false
 		end
 	)
+	-- Shift+P would be overwritten in inputbox by entering char 'P'
+	-- I suggest one should probably change the hotkey to, say, Alt+Space
+	obj:add(KEY_P, MOD_SHIFT, "P", "make screenshot",
+		function()
+			Screen:screenshot()
+		end
+	)
 	return obj
 end
--- a/dialog.lua
+++ b/dialog.lua
@@ -2,7 +2,6 @@ require "widget"
 require "font"
 
 InfoMessage = {
-	face = Font:getFace("infofont", 25)
 }
 
 function InfoMessage:show(text,refresh_mode)
@@ -22,7 +21,7 @@ function InfoMessage:show(text,refresh_mode)
 				}),
 				TextWidget:new({
 					text = text,
-					face = Font:getFace("cfont", 30)
+					face = Font:getFace("infofont", 30)
 				})
 			})
 		})
--- a/extentions.lua
+++ b/extentions.lua
@@ -3,20 +3,24 @@
 ext = {
 	djvuRead = ";djvu;",
 	pdfRead  = ";pdf;xps;cbz;",
-	--creRead  = ";epub;txt;rtf;htm;html;mobi;prc;azw;fb2;chm;pdb;doc;tcr;zip;"
+	creRead  = ";epub;txt;rtf;htm;html;mobi;prc;azw;fb2;chm;pdb;doc;tcr;zip;"
 	-- seems to accept pdb-files for PalmDoc only
 }
+
 
 function ext:getReader(ftype)
 	local s = ";"
 	if ftype == "" then
 		return nil
-	elseif string.find(self.djvuRead,s..ftype..s) then
-		return DJVUReader
 	elseif string.find(self.pdfRead,s..ftype..s) then
 		return PDFReader
-	else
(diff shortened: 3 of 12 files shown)


houqp:
@NuPogodi, seems that you are quite comfortable with Git :)

houqp:
Would be better if you guys take some time to record all the new features in the Wiki. There are so many new improvements now and since I am not the original author, I might miss some of them if I do it by my own ;P

tigran123:
@NuPogodi  Has the behaviour of showInfoMsgWithDelay() changed after these changes? Now I cannot call it in some situations, but still can in others. For example, if I use it in showToc() (where before your changes it worked), now I see this in the terminal output:

```
# command to execute: [28]T: show table of content (TOC)
# # InfoMessage  This document does not have a TOC.
# E: T=1  V=1  C=28  DUR= 2129045
```

but no window with the message pops up. But if I press F11/F12 to change gamma (this is all in emulator) then the window with "New gamma=" is correctly displayed. So, are there situations now when we cannot call showInfoMsgWithDelay() function?

tigran123:
Btw, I have just retested with only plain master sources (with your changes applied, but _without_ my multilevel TOC changes) and it does not work there either. I.e. you can reproduce it yourself by simply opening any file which has no TOC in it --- the message about it is not displayed.

tigran123:
Ah, ok, now I see that if I change infont -> infofont in dialog.lua then TOC window is displayed correctly. Btw, in your latest pull request this change is not present --- please check.

ghost:
The message is shown only when it fits to the screen width...

tigran123:
@NuPogodi  In this case it had absolutely nothing to do with the width of the screen. It was due to "infont -> infofont" change in dialog.lua (see my previous comment).

ghost:
I'll check... And you, please, try to check whether reducing the width of infotext helps or not, ok? Such a behavoir was long before I made the mentioned above 'infofont'-corrections

tigran123:
1. I didn't change the message, so it is exactly the same length as before.
2. It fits on the screen perfectly (and always did).
3. Making it shorter makes no difference.
4. Changing the infofont -> infont in dialog.lua causes it to stop working.

ghost:
If you leaved the same message text, but changed the font to render this text, you changed the SIZE of InfoMessage-window. Infofont is just a name, nothing else... changing 'infofont' <> 'infont' has no effect if both fonts are, say, "Droid/DroidSans.ttf" If they are different, using wider font may cause what you have reported - silent dissapearing the infomessage itself. So, please, RECHECK item 3 by reducing the text of your message to "TOC".

I may check my part when will finally arrive home.

tigran123:
Ah, I see, I thought that by "size" you meant the number of bytes in the text, not the number of pixels in the width.

Ok then, that makes a lot of difference. Yes, reducing the text from the old value of "This document does not have a TOC." to a shorter version of "No table of contents" makes it work regardless of the infont/infofont stuff.

tigran123:
Hmmm, why did I say in "3." that making it shorter makes no difference? Because whilst testing I put various messages (but looks like all of them were too long) in various places and this made no difference :)

ghost:
I told nothing about size of text, I wrote "The message is shown only when it fits to the screen width..."

Ok, You know now that you shoud be relatively brief when writing text for InfoMessage:show or showInfoMsgWithDelay() what is generally the same function. Or, alternatively, you may spend some time to modify their code and to replace a bit outdated function renderUtf8Text() by more flexible renderUtf8Multiline(). It was mentioned in my TODO (see filechooser.lua, menu to switch filemanager modes), but I may gratefully refuse from this right...

tigran123:
Thank you, @NuPogodi 

Enjoy your holidays! :)

PR #252, opened 2012-09-07
Bugfixes for typos in last pull request

ghost:
1. restored: own font 'infofont' for messages & warnings
2. fixed: typo that caused helppage to crash

### DIFF
--- a/dialog.lua
+++ b/dialog.lua
@@ -21,6 +21,7 @@ function InfoMessage:show(text,refresh_mode)
 				}),
 				TextWidget:new({
 					text = text,
+					-- info messages should have own font face
 					face = Font:getFace("infofont", 30)
 				})
 			})
--- a/helppage.lua
+++ b/helppage.lua
@@ -59,7 +59,7 @@ function HelpPage:show(ypos, height, commands, title)
 	while true do
 		if is_pagedirty then
 			fb.bb:paintRect(0, ypos, fb.bb:getWidth(), height, 0)
-			DrawTitle(title or "Active Hotkeys",self.margin_H,0,self.title_H,self.bg_color,tfont)
+			DrawTitle(title or "Active Hotkeys",self.margin_H,0,self.title_H,self.bg_color,tface)
 			local c
 			local max_x = 0
 			for c = 1, perpage do


dpavlin:
I had to merge this request to master manually because of conflict, but now commit 6c9fb8b023ef48670c6dce29892bd5a64e89d205 looks correct :-)

Thanks @NuPogodi for all your hard work it's really appreciated.

dpavlin:
Unfortinatly now it dies with

```
# # InfoMessage  Registering fonts... 1
# #! Font infofont (./fonts/infofont) not supported: freetype error
lua config error: ./widget.lua:108: attempt to index field 'face' (a nil value)
```

houqp:
OK, I found out the cause of the crash. It is in `settings.reader.lua`. So if you start with a fresh setting file, everything should go right. But in my old setting file, I have:

``` lua
    ["fontmap"] = {
        ["hfont"] = "droid/DroidSans.ttf",
        ["infont"] = "droid/DroidSansMono.ttf",
        ["scfont"] = "droid/DroidSansMono.ttf",
        ["tfont"] = "NimbusSanL-BoldItal.cff",
        ["rifont"] = "droid/DroidSans.ttf",
        ["ffont"] = "droid/DroidSans.ttf",
        ["cfont"] = "droid/DroidSansFallback.ttf",
        ["pgfont"] = "droid/DroidSans.ttf",
        ["hpkfont"] = "droid/DroidSansMono.ttf"
    },
```

This over-write the fontmap table in font module. So it hides infofont and adds pgfont. Given that many other users might also have similar settings files like me, we need to handle the `fontmap` section specially while reading `settings.reader.lua`. I think the same goes for other setting sections with subtables.

ghost:
Ohhh, my bad... I did tested the filemanager (^+P+D), but not run the reader directly (^+P+P). The former runs smoothly (the messages are really rendered with 'infofont'), while the latter crashes. ATM, I have no other suggestion than the reader try to show message (now with infofont), but Font.fontmap (or FreeType) is not yet defined. One has to check the sequence of commands in reader.lua

houqp:
Um.. Maybe we can fix it in this way:

Currently, what we does for reading fontmap config is:

``` lua
fontmap = G_reader_settings:readSetting("fontmap")
```

A quick dirty workaround can be:

``` lua
fontmap = G_reader_settings:readSetting("fontmap")
if fontmap ~= nil then
    for ftname, face in ipairs(fontmap) do
        if Font.fontmap[ftname] then
            Font.fontmap[ftname] = face
        end
    end
end
```

This make sure we do a entry by entry update. Maybe someone can come up with a better solution.

dpavlin:
@houqp you are fast. I'm coding more-or-less same thing right now :-)

dpavlin:
I will close this pull request and if you agree with fix, @houqp can merge it to master :-)

Issue #235, opened 2012-08-29
Multi-level TOC in PDF and DjVu files should not be expanded.


tigran123:
If a book contains a large multi-level outline tree (aka "bookmarks" or "TOC") it will be very hard to navigate if it is fully expanded (as is the case with kindlepdfviewer at present). For example, if you look at this file (http://www.bibles.org.uk/The-British-Study-Edition-of-the-Urantia-Papers.pdf.zip) it contains four Parts, each Part contains many numbered Papers, each Paper contains  6-15 numbered Sections. (actually each Section contains numbered Paragraphs but I didn't create bookmark entries for these to avoid excessively large size of the PDF file).

The solution to this problem would be to indicate in the standard way (i.e. with the '+' sign, like in Acrobat, CoolReader, etc) that a node can be expanded and then display the appropriate submenu if the user selects it. And if the node is expanded, then selecting it should collapse it.

Also, I should emphasize that by default the tree should be _collapsed_ (like in Acrobat), not expanded (like in CoolReader), because if you expand it and wish to navigate to a node at the bottom you would have to collapse all the previous parents.

A separate small problem is that pressing 'T' to get the TOC should also position you to the current place in the tree, not to the beginning.

NuPogodi showed me how to use menus in kindlepdfviewer and I quite liked the mechanism used. So, if there is some kind of "assignment" mechanism in this project then feel free to assign me to this task --- I would like to try implementing it.

ghost:
[QUOTE]A separate small problem is that pressing 'T' to get the TOC should also position you to the current place in the tree, not to the beginning.[/QUOTE]
I added this feature to my version of unireader.lua just before the release in May, but it was somehow disappeared in the released version.
[CODE]
function UniReader:showToc()
    if not self.toc then
        -- build toc if needed.
        self:fillToc()
    end
    -- build menu items
    local menu_items, item_no = {}, -1
    for k,v in ipairs(self.toc) do
        table.insert(menu_items, ("    "):rep(v.depth-1)..self:cleanUpTocTitle(v.title))
        -- to find current TOC-entry
        if v.page <= self.pageno then
            item_no = item_no + 1
        end
    end
    if #menu_items == 0 then
        showInfoMsgWithDelay("This document does not have a TOC.", 2000, 1)
    else
        toc_menu = SelectMenu:new{
            menu_title = "Table of Contents",
            item_array = menu_items,
            -- to autoselect current TOC-entry
            current_entry = item_no,
        }
[...]

PR #255, opened 2012-09-08
Multi-level TOC support.

tigran123:
The need for multi-level Table of Contents (aka "outline" and "PDF bookmarks") was raised in #235.
Now I have implemented it and tested on a variety of PDF and DjVu files (and even FB2 files).

### DIFF
--- a/selectmenu.lua
+++ b/selectmenu.lua
@@ -43,6 +43,7 @@ SelectMenu = {
 	selected_item = nil,
 
 	commands = nil,
+	expandable = false, -- if true handle Right/Left FW selector keys
 
 	-- NuPogodi, 30.08.12: define font to render menu items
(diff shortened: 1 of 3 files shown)


tigran123:
Actually, before you apply this, please wait for me to commit the change to selectmenu to allow a variant of choose() method which does not refresh the header. This would avoid irritating header redraws on expand/collapse requests.

tigran123:
No, ignore my previous comment. Please apply the pull request "as is", because if we don't do a full refresh in selectmenu, then the screen will get very messy, i.e. there will be ghosts on the screen.

houqp:
awesome patch! It's really handy :)

tigran123:
Thank you! :)

PR #257, opened 2012-09-09
 fix crash on documents without TOC

dpavlin:
Showing TOC on documents without one results in following error:

lua config error: ./unireader.lua:1742: attempt to get length of a nil value


tigran123:
Oh, sorry, I should have tested more thoroughly. Of course I tested on documents without TOC initially, but not after the rewrite, which introduced this bug.

dpavlin:
No problem, I just need another pair of eyes on my code to be sure that I didn't do something strange :-)

Issue #206, opened 2012-06-12
slow at open some epub/mobi


dracodoc:
This issue was mentioned before in 
https://github.com/hwhw/kindlepdfviewer/issues/179

That issue included multiple issues so it was closed because some of them were solved.

kpdf is slow at opening certain epub/mobi. It took 1 min to open this 2M epub or mobi every time. The original software can open the mobi immediately.

https://www.box.com/s/42bc8e9df7cbe11a6126
https://www.box.com/s/59a86614ef552d3f86de

Then I tried coolreader in kindle from here
http://www.mobileread.com/forums/showthread.php?t=148850

Opening a 800K epub took it 7-8 seconds, but it took just 2 seconds open it in second time. While kpdf will need a full 25 seconds to open the same book every time.

So it seemed not a issue of CRengine itself.

houqp:
Wow, good to know that it is not a upstream bug :) There must be something wrong with our Lua glue code...

dracodoc:
I'm thinking maybe coolreader created some cached file after the first time opening a book. There are some pdb file in cached directory under cr3 directory, maybe that's why it can open book much quicker after the first time.

dracodoc:
I looked at the creader.lua. It seemed kpv always use local style sheet instead of the built in css from the epub book. I'm not sure if this is desired. Will there be some performance problem if the local css have something mismatch with the epub?

I replaced the epub.css with cr3.css, the time to open one epub seemed reduced a little bit, probably because the cr3.css is simpler.

houqp:
um... to check this out, maybe we need to doing some profiling on crengine to see whether it is spending most of the time on CSS parsing.

dpavlin:
I have been playing with coolreader on Android a bit, and I can see a (small, but Android device is much faster) difference when using internal document CSS (which we do now) and ignoring it.

I also noticed that ignoring document CSS is only way to make font size usable -- on DXG Linux Journal epub always have fonts too small to be readable, and our font size selector seems to change just default font size, so whole document looks very strange.

There is also option to render per-page and scrolling. We are currently using scrolling, but I don't see speed difference between them -- it seems that crengine layouts whole document on open time, so I took a look into Android implementation and found `setCacheDirectoryInternal` (and not back to code...)

dracodoc:
dpavlin, thanks for the explanation! 
[...]

PR #258, opened 2012-09-09
use ldomDocCache::init to cache crengine DOM

dpavlin:
This should improve performance of crengine #206


houqp:
wow! does it speed up the opening process on the second run?

dpavlin:
I don't have exact numbers (i counted seconds in my head :-) but it seems to shave off about 3 seconds from document opening time on my documents after first open.

PR #260, opened 2012-09-09
Slight tidy up to the fix for \#257

tigran123:
For documents with no TOC we need to redraw the screen after showing the "No Table of Contents" message, otherwise the message "Retrieving TOC..." will stay on indefinitely.


tigran123:
I added a couple of fixes to filechooser.lua unrelated to the main topic of this pull request (see description in the commit).

PR #266, opened 2012-09-09
Shorten the message about opening the document.

tigran123:
Shorten the message about opening the document to make it displayed at all.
(sorry for the very trivial change).

I learned how to use git branches properly and now I have interesting branches like "experimental", "filemanager" and "reflow" which have much bigger, but still very buggy and unfinished pieces of work.


houqp:
Branch is cheap in git, so happy branching :)

PR #268, opened 2012-09-10
set crengine cache size in lua code

houqp:


dpavlin:
@houqp useful patch, thanks.

PR #271, opened 2012-09-10
TOC cleanup in crengine.lua

tigran123:
1. Shorten the "This document does not have a TOC." message to "No Table of Contents" in order to make it visible.
2. Get rid of all the calls to self:fillToc() from the places in the code which cannot affect TOC, such as rotation, font change and interlinear spacing change. I assume the reason for these calls is that they are needed in the original CoolReader application (because there the TOC displays the virtual "page numbers" which can change after any of these operations) and were brought over to our crengine.lua during the port.

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -346,7 +346,7 @@ function CREReader:gotoPrevNextTocEntry(direction)
 		self:fillToc()
 	end
 	if #self.toc == 0 then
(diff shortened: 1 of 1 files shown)


houqp:
If I remember correctly, the first edition of crereader uses percent to mark position inside documents, so I have to do manually reload when document length changes. But I can't remember why these code are still here after I switched to xpointer ;P

Anyway, you are right it seems that deleting all these does not break anything.

Issue #269, opened 2012-09-10
kpdfviewer crashed using jump history


eLiNK2gl:
while navigating though the jump history (shift + B) in a pdf file, the kpdfviewer crashed with the following error message:

"lua config error: ./unireader.lua:1693: bad argument #1 to 'ipairs' (table expected, got nil)"

houqp:
Could you tell us which version are you using? I can't find a way to reproduce in latest code.

eLiNK2gl:
@houqp '397d3aa037'
By navigating I meant calling up the jump history (Shift + B), selecting an item and pressing 'Enter'.

tigran123:
Oh, line 1693 points to my code in UniReader:findTOCpos(). I will investigate, leave it to me.

tigran123:
@eLiNK2gl I just tested it on a PDF file and cannot reproduce it. Can you please tell me which steps you performed to cause this to happen?

tigran123:
@eLiNK2gl 'Enter' is not a valid key for menus, unless you are using KDXG with long enough menus to reach it as a shortcut. On a K3 pressing 'Enter' in a menu will cause nothing to happen, which is (and always was) the expected behaviour.

tigran123:
Also, that code is never called from jump history, but only from "show table of contents" function, i.e. "T". Are you sure you were not navigating Table of Contents?

eLiNK2gl:
hmm... now I am not so sure.

Actually, I was going to test the bookmarking feature but mixed it up and called the jump history instead.
By 'Enter' I meant the middle key of the five-way.

I could have unintentionally pressed some other button.
But I was not looking at the dispay at the moment. Just saw afterwards that the kpdfviewer crashed.
(It would be probably the only reasonable explanation)
Sorry for the mix-up.

tigran123:
@eLiNK2gl  No problem at all. If you can reproduce it again please tell me the exact steps and I will try to fix it.

tigran123:
Ah, I think I have reproduced it now. The sequence of steps is this:
1. Open some PDF file on a page other than 1.
2. Go to page 1 manually (via "G" command enter "1")
3. Press "T" to get TOC, the crash follows:

lua config error: ./unireader.lua:1694: bad argument #1 to 'ipairs' (table expected, got nil)

Line number 1694 in my code is the same as 1693 in the one you are running because I just added one Debug() statement.

Ok, I will investigate now, leave it with me.

[...]

PR #272, opened 2012-09-10
The fix for the bug issue #269.

tigran123:
The main change is the fix for the issue #269 but also I fixed a little bug in findTOCpos() to make it correctly determine TOC position for the current page being _before_ the first page with a TOC entry.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1609,7 +1609,52 @@ function UniReader:cleanUpTocTitle(title)
 end
 
 function UniReader:fillToc()
+	InfoMessage:show("Retrieving TOC...", 1)
 	self.toc = self.doc:getToc()
+	self.toc_children = {}
+	self.toc_xview = {}
+	self.toc_cview = {}
+	self.toc_curidx_to_x = {}
+
+	-- To combine the forest represented by the array of depths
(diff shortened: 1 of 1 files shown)


tigran123:
Please note that although it is very tempting to get rid of the "Retrieving TOC..." message (and the corresponding page redraw(s)) I had to leave it because there are real books (e.g. my Hebrew Bible editions with morph tags) which have over 24,000 TOC nodes and on a Kindle 3 it takes about 35 seconds to generate the TOC, so without this message one would presume that the application hang.

tigran123:
Actually, it would be curious to find out how much of those 35 seconds is spent in lowlevel retrieving the physical TOC and how much is spent in my Lua code walking those depths arrays and determining parent/child relationships...

tigran123:
On my desktop PC (3.2GHz dual-core) the numbers are (on a Hebrew Bible with 24185 TOC entries, i.e. bookmarks for every single verse --- this is needed for morph-tagged editions because each page contains no more than 2-3 verses):

Physical TOC is ready in 1.47(s) 
Logical TOC is ready in 0.04(s)

So, most of the time is spent in the low-level format-specific code. The time spent in Lua TOC manipulation is negligible.

PR #277, opened 2012-09-14
Generate expandable TOC only if makes sense.

tigran123:
Currently the message "Press Right/Left FW-selector keys to expand/collapse items" is displayed even for books that have no expandable items in the TOC at all. Now we only generate expandable TOC menu if there really is at least one expandable item.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -92,6 +92,7 @@ UniReader = {
 	bookmarks = {},
 	highlight = {},
 	toc = nil,
+	toc_expandable = false, -- if true then TOC contains expandable/collapsible items
 	toc_children = nil, -- each element is the list of children for each TOC node (nil if none)
 	toc_curitem = 0, -- points to the current location in TOC
 	toc_xview = nil, -- fully expanded (and marked with '+') view of TOC
@@ -1610,6 +1611,7 @@ end
 
 function UniReader:fillToc()
 	InfoMessage:show("Retrieving TOC...", 1)
+	self.toc_expandable = false
 	self.toc = self.doc:getToc()
 	self.toc_children = {}
 	self.toc_xview = {}
@@ -1635,6 +1637,9 @@ function UniReader:fillToc()
 			table.insert(self.toc_children[prev], k)
 			parent[k] = prev
 			self.toc_xview[prev] = "+ "..self.toc_xview[prev]
+			if prev > 0 then
+				self.toc_expandable = true
+			end
 		elseif (v.depth == prev_depth) then --> k and prev are siblings
 			parent[k] = parent[prev]
 			table.insert(self.toc_children[parent[k]], k)
@@ -1776,7 +1781,7 @@ function UniReader:showToc()
 			menu_title = "Table of Contents (" .. tostring(#self.toc_cview) .. "/" .. tostring(#self.toc) .. " items)",
 			item_array = self.toc_cview,
 			current_entry = self.toc_curitem-1,
-			expandable = true
+			expandable = self.toc_expandable
 		}
 		ret_code, item_no = toc_menu:choose(0, fb.bb:getHeight())
 		if ret_code then -- normal item selection
@@ -2112,6 +2117,7 @@ function UniReader:inputLoop()
 	-- do clean up stuff
 	self:clearCache()
 	self.toc = nil
+	self.toc_expandable = false
 	self.toc_children = nil
 	self.toc_curitem = 0
 	self.toc_xview = nil


tigran123:
The corresponding change to kpdfdjview branch I have merged myself. If you want me in the future to go through pull request procedure for kpdfdjview branch as well, please let me know.

Actually, in the future, the correct protocol should be:
1. make the change in the local master tree.
2. make a pull request to HW's master tree
3. once it is merged, also merge it to the local kpdfdjview branch
4. merge in the remote kpdfdjview branch (or go through pull request to it)

This way I don't have to do the same change in two places manually, as I did just now.

eLiNK2gl:
Any chance to get expand / collapse all for a given node (maybe 'shift' + 'right')?
Could come handy for some very elaborate TOCs.

tigran123:
@eLiNK2gl  Good idea, working on it right now, thanks.

tigran123:
Oops! I pushed it to upstream instead of origin! Sorry about this, I was supposed to push to origin and then it would be merged via pull request. I will be more careful next time.

Anyway, @eLiNK2gl I implemented what you wanted --- Shift-Right expands all subitems of a given item.

tigran123:
@eLiNK2gl I put today's build in my Wiki area, so you can give it a try if you wish: https://github.com/tigran123/kindlepdfviewer/wiki/Download

houqp:
wow, cool :)

tigran123:
I had to add the commit to selectmenu.lua (breaking the pull request protocol) as part of this "exception because of my mistake". But in the future commits I will follow the pull request protocol.

tigran123:
@houqp Thank you, the function is recursive and very nicely symmetrical to the corresponding "collapse all" function.

eLiNK2gl:
@tigran123 we'll be glad to try it out :)
Thanks

eLiNK2gl:
@tigran123  Looks good but 'shift + left' (for 'collapse all') does not work. 'Left' does it instead. Is it by design?

tigran123:
@eLiNK2gl Yes, it is by design. Collapsing only outermost level is much harder to implement than collapsing all levels because one would have to remember the state of which hidden items are expanded to which level. So it is easier to assume that everything that is hidden is also collapsed.

ghost:
My suggestons & propositions:
0. The expandable TOC is better to open so that (i) the current TOC-childitem & all their parents are expanded, (ii) current TOC-item (rather than its highest parent) is selected.
1. Expanding one expandable TOC-item might be better to make to be accompanied by autocollapsing of all the rest expandable items.
2. Try to never change the sequence of return parameters without vital necessity! You've done it with 
   old version --> item_no, all = toc_menu:choose(0, fb.bb:getHeight())
   new version --> ret_code, item_no, all = toc_menu:choose(0, fb.bb:getHeight())
Otherwise, make it absolutely sure that you have also corrected ALL functions (rather than in unireader.lua) that call old version. The result: TOC in crereader.lua does no longer work properly
3. Try to avoid long hotkey prompts in the footer (changing the 'ffont' may cut the footer text): helppage.lua was written to cover exactly such purposes. All you need is add to selectmenu-commands smth like
        self.commands:add(KEY_H,MOD_ALT,"H",
            "show help page",
            function(sm)
                HelpPage:show(0, G_height, sm.commands)
                sm.pagedirty = true
            end)

tigran123:
Hi @NuPogodi 

Nice to see you back from the holiday, hope you enjoyed it! :)

On your suggestion 0.  I completely and very strongly disagree, because to narrow down the current location to the lowest level (i.e. to the leaf node) all you need to do is to press Shift + Right and then Back and then T again, i.e. just three key presses in total. But if you fully expand the whole TOC in the initial view then you may have to press Prev key INFINITE number of times to get back to the outermost parent (i.e. depending on the total number of items and the depth of the current location). But the whole point of adding expandable TOC support is precisely to avoid this.
1. Why? I see no point in doing it and I can imagine annoyance of the user who suddenly discovers that some of his expanded items were auto-collapsed. And this is not the behavour of standard PDF viewers like Adobe Acrobat Reader and evince, so we shouldn't put such "surprizes" into ours either.
2. I have actually checked the usage in crereader. Ok, I will recheck it again and fix it if I have really broken it. Thank you for pointing out a potential problem.
3. Ok, I'll shorten it, thank you. And thanks for the tip about the help page.

ghost:
> But if you fully expand the whole TOC in the initial view then you may have to press Prev key INFINITE number of times to get back to the outermost parent

Just redefine pressing Q..P, A..Del, Z..Enter for expandable TOCs not to return selected_item, but just to select it within selectmenu... Then, there is no need to navigate between TOC-items via FW and, hence, no need to use INFINITE number of presses to select proper item either to return (via Enter) or to expand / collapse. Is it not what selectmenu was written for?

tigran123:
Hmm, I didn't quite understand how redefining those keys would help. Let us consider a specific example and then maybe I will understand, if you make it clearer. Suppose the TOC contains 24185 items with the maximum depth level 4 (this is an actual real book --- morphologically parsed Hebrew Bible based on Leningrad Codex). And your current position is on the item 2Chronicles 36:23 (the last verse of the Hebrew Bible). In the fully expanded TOC this happens to be (on Kindle 3) item 5 on page 1210. So, to get back to Genesis 1:1 I would have to press Prev key 1209 times. I see no obvious shorter way. And what is the shorter way that you see? Ok, let's suppose I redefined Q..P etc keys to select an item (not actually jump to its destination) as you suggest. So what? What will I press then to get to Gen 1:1 from 2Ch 36:23? I still don't get it, sorry :)

tigran123:
@NuPogodi 
Concerning crereader: yes, I see that TOC handling in there is quite broken at the moment, though I am not yet sure that it has anything to do with SelectMenu:choose() function. I will investigate and will try to fix it. Thank you again for highlighting this issue.

ghost:
REread what I wrote, please, before answering. Only the direct parents of the current TOC-item should be expanded, the rest ones have to be collapsed. If we suggest that each of 4 depth levels contains 20 items, then simultaneously seen in selectmenu should be
1. all highest level items, i.e. parents = 20 items, one is expanded so we can see also
2. the second level = 20 items, one is expanded so we can see also on the
3. third level = 20 items, one is expanded so we can see also on the
4. lowest level = 20 items (one of them, i.e. current TOC-item should be selected)
Thus, we should see 80 items on 4 pages instead of 20^4=160'000 on 8000 pages including all the parents & neighbours of the current TOC-item

Path from 20:20:20:20 to 1:1:1:1 (i.e. definitely the longest path as you friendly suggested) = 
1) 3 x push "<" to arrive 1st TOC-page with the needed parent item 1 or
3 x push "collapse previous depth level" or
one can add 'goto page' inputbox to select proper TOC-page (again 3 pushes = call inputbox, type 1st page + Enter)
2) 3 x (push "proper character" to select proper item and expand it) = 6 times
3) "Enter" to return the value

So, if one have to go to the TOC-item within the same, lowest level (i.e. 20:20:20:20 > 20:20:20:XX) = 2 presses (char to select + Enter)
if one have to go one level up and then one down (i.e. 20:20:20:20 > 20:20:XX:XX) = collapse lowest level (it might be avoided if one make autoclosing the alien parent items on expanding the needed one), select proper item from the 3rd level, expand it, select proper item + Enter = 5 presses and so on...

Now try to calculate how many presses should you perform now to jump from 20:20:20:20 to say 1:20:20:20 not for ideal, but for realistic case - namely, you have already used TOC and expanded 2-3 highest parents (somewhere between the current TOC-item and the target) simultaneously (that was the reason I suggested autoclosing alien TOC-parents)

ghost:
And I've just forgotten to mention... The way I suggested is more effective if one knows aprroximate position of the target inside the document - goto approximate page/percent, then TOC to show neighbours and a couple of pushes to jump to the exact position

tigran123:
Ok, let me think about it (while fixing the crereader/TOC breakage) and then we'll come back to your idea of starting with all direct parent of the current item expanded (originally I thought you were suggesting to expand ALL items, not just the direct parents of the current one).

tigran123:
Btw, just as I expected, the problems with TOC in crengine have nothing to do with my modifications to SelectMenu:choose() method (as I said, I checked carefully before doing it). Rather, they have to do with the fact that crereader has its own gotoTocEntry() method which uses "xpointer" but I assumed that one can simply store the page number from a TOC entry and then goto by page number. Now I see that this is applicable only to PDF and DjVu, but not to the files handled by crengine.

I'll think about the best way of fixing it (it is not trivial, because I rely on those page numbers also for finding the current position...) but at least we know now what is causing the TOC havoc with crengine-related files...

tigran123:
There seems to be two problems:
1. TOC navigation problem because of my assumption that goto by page number is a globally valid operation. This is now fixed in my tree by commit SHA: 3056bf440c26550bb8ac9fba087b7044aa87bc9a, namely by the change from calling `self:goto(pagenum)` to `self:gotoTocEntry(toc_entry)` in `UniReader:showToc()`
2. For some reason CREReader:gotoTocEntry() doesn't refresh the screen. I am not yet sure why. With the above fix one can get to the correct TOC destination and then press Menu to see that this is the case. Obviously, we need to eliminate the need to press Menu, i.e. find out why self:redrawCurrentPage() has no effect in this situation. I'll continue thinking about this tomorrow.

ghost:
> Btw, just as I expected, the problems with TOC in crengine have nothing to do with my modifications to SelectMenu:choose() method

I don't think so

> I'll think about the best way of fixing it (it is not trivial, because I rely on those page numbers also for finding the current position...)

You again overestimated the task. It's trivial. All you need is to return the right item_no that is readily calculated when one knows how many previous TOC-items are expanded/collapsed.The rest in crereader.lua is already adjusted to jump to this item...

tigran123:
Hmmm, maybe we are looking at different programs :)

In kindlepdfviewer the right item_no is already returned by SelectMenu:choose() to UniReader:showToc() and CREReader does NOT have its own showToc() method, so the "calculations" you refer to are non-existent (and they never existed, i.e. before my changes either).

Anyway, the second part of the fixwill be coming soon. I vaguely remember having changed something recently which caused an almost subconscious thought "ah, this could in theory make crereader miss one page redraw..." but I forgot what it was. I'll go through the "git log" output and remember and will fix it properly. It was in response to @eLiNK2gl report that pressing Menu overlays the info on top of the text and so I have re-arranged page redraws but looks like my rearrangement was only good for PDF and DjVu but not for crereader files. I will re-check those changes and adjust them to work properly with both crereader and PDF/DjVu.

When I have tested the whole fix properly I will do a pull request.

tigran123:
Btw, an independent proof that my analysis is correct is by using bookmarks: if you create two (or more) bookmarks and invoke the bookmarks list by pressing "B" and then select any bookmark pointing to the page other than the current one then: a) page is not redrawn b) pressing Menu shows the wrong section and xpointer but pressing Menu again shows the correct section and xpointer and the page is redrawn. (ok, there is an alternative logical explanation that page is actually redrawn but it is the wrong (i.e. still current) page until Menu is pressed for the _second_ time and then the page is correct and so it is shown correctly too --- but this seems very fantastic to me). Still investigating...

ghost:
> Hmmm, maybe we are looking at different programs :)

Hmmm, sometimes much easy to fix it by myself

tigran123:
It would have been even easier if you read my comments more carefully. I have already done that part (I called it "part 1") yesterday. It was trivial indeed. The non-trivial part still remains...

ghost:
The same wishes to you!

tigran123:
Ok, we shall both read each other's comments more carefully next time, let us put this issue to rest now :)

PR #278, opened 2012-09-14
Stop calculator from crashing on empty input.

tigran123:
If you invoke the calculator via Shift-K and just press Enter or the middle button of FW-selector then the calculator will crash. Now it displays "No input" message instead of crashing.


tigran123:
I added another minor bugfix to the calculator: on exit from its help page it didn't refresh the screen properly because it tried to do some clever calculations of which portion of the screen to refresh (which doesn't work, but that is a separate bug). Refreshing the full screen always works.

tigran123:
Another bugfix that can go into the same pull request: currently calculator assumes that the cursor position depends only on the number of characters typed and so pressing "Del" repeatedly would only delete the last digits (as many as there were in the input). But if, for example, you type a single character symbol "pi" then the output contains quite a few digits (up to the double precision used in Lua) and therefore pressing "Del" only once will not delete the whole expression. With this fix "Del" works correctly, i.e. you can press it as many times as there are characters in the output and it will let you delete as much of it as you like.

PR #279, opened 2012-09-14
Calculator: make it more robust.

tigran123:
Previously it was very easy to crash the application by entering invalid input into calculator, for example
things like these: ")(", "inf/0", "f()", etc. By using Lua pcall() function to execute not only the construction of
the expression, but also the calculation of the final result, the calculator is made more robust and safe-guarded from such crashes.


tigran123:
With this fix the only way to crash the calculator that I could find was to enter "os.exit()" in the input box. I don't know how to guard against it :)

PR #281, opened 2012-09-15
Minor changes to djvu and filesearcher.

tigran123:
1. Change the name of DjVu context created to "kpdfview" as it is common practice to call the context the name of the program. (though I understand the reason for the current name, i.e. pointing to Lua object DJVUReader)
2. When changing DjVu rendering mode we don't need to clear the djvulibre document cache, only the internal tile cache.
3. In file searcher: a) show the number of hits, b) show the actual pattern entered, c) shorten the delete confirmation message to make it visible, d) make the search case-insensitive (so searching for "DJVU" will find djvu files, previously it didn't).

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -79,9 +79,9 @@ static int openDocument(lua_State *L) {
 	luaL_getmetatable(L, "djvudocument");
 	lua_setmetatable(L, -2);
 
-	doc->context = ddjvu_context_create("DJVUReader");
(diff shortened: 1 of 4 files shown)


tigran123:
Also added an interesting fix for TOC: when it contains a link to external file (which we don't support) we should redraw the current page rather than just silently returning from UniReader:goto() (with the TOC still displayed on the screen)

PR #283, opened 2012-09-15
Fix for calculating TOC position.

tigran123:
In the (rare) cases when TOC contains a link to external files the calculation of the current position in TOC was incorrect, fixed now.


tigran123:
I included a better fix for the external TOC problem: display a message "External links unsupported" and remain in the TOC menu rather than quietly return to reading the current page and leave the user wondering what is going on (and possibly keep retrying to access that mysterious TOC entry).

PR #285, opened 2012-09-16
Update readme

houqp:
I also directly pushed my modification to Makefile to upstream by accident... 

It was intented to be included in this pull request. So please also check commit ea325146d3738b0b1f8b76fe436ea68129ea5593 in hw's master branch. ;P

BTW, seems that git is not intelligent enough to track renamed README, I will use `git mv` next time...

### DIFF
--- a/README.TXT
+++ b/README.TXT
@@ -12,24 +12,26 @@ The application is licensed under the GPLv3 (see COPYING file).
 Building
 ========
 
+
 Follow these steps:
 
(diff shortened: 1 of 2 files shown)


dpavlin:
FWIW , git mv isn't much better

git still guesses about renames from commit itself, so safeest way is to do git mv, and commit only that before making changes to file. It's like rename commit...

houqp:
I see :) Thanks for the tip.

PR #286, opened 2012-09-16
Stop writing battery info messages to crash.log every few seconds.

tigran123:
We only need standard output of `gasgauge-info -s` command, so standard error can be safely redirected to /dev/null and thus prevents those messages in crash.log every time the user presses "Menu" or does anything else that invokes the BatteryLevel() function.

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -57,7 +57,7 @@ end
 function BatteryLevel()
 	local fn, battery = "/tmp/kindle-battery-info", "?"
 	-- NuPogodi, 18.05.12: This command seems to work even without Amazon Kindle framework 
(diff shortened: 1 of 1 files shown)


tigran123:
Btw, I found a binary of vim compiled for Kindle and so I could make a change and test it directly on the device (obviously this could not be tested in emulator). Also, I copied the compiled binary of luajit to my Kindle so I could experiment with small Lua programs without turning on computer (I guess I could use the kpdfdjview binary directly, but a vanilla luajit seemed a more sane option :)

houqp:
WOW, using Vim within kindle is cool :)

PR #293, opened 2012-09-18
Cleanup: use util.isEmulated() == 1 instead of looking at KEY_FW_DOWN value.

tigran123:
As the subject says, use util.isEmulated() function in screen.lua to decide that we are running in an emulator. (probably when this code was written the util.isEmulated() function didn't exist)


tigran123:
A couple of changes added to my pull request:
1. Use pipes when calculating battery level instead of writing to and reading from temporary file.
2. Fix the calculation of the total unpacked size of a zip file in case it contains multiple entries. And also use pipes for this as well, instead of writing to a temporary file and then reading from it.

Issue #295, opened 2012-09-19
polishing highlighting feature


eLiNK2gl:
it is possible to highlight a text (press 'n'), which I personally use quite a bit.

1) but after a part of text has been highlighted, the highlight mode is automatically deactivated, so if one intends to highlight another part of text on the same page, you'd need to move the cursor all way down again.
2) highlights cannot be comfortably removed. (You need to move the cursor to your highlight and press delete), Ideally, one should be able to remove them one by one ( or all) like bookmarks.

Please finish the beast :)

P.S.: On a side note, I believe whenever something gets (be it a file, bookmark or a highlight) deleted, it would be more comfortable if the previous (or next item) gets activated. Currently, if you delete a file, the cursor gets placed at the begining of the folder (or the list, as is the case with 'Last read').

tigran123:
I agree with your P.S. comment. However, it is not hard to workaround because Page Next/Prev keys work in the file manager as well and they allow you to quickly jump to the required location. I.e. you don't need to move there in small one-line steps with FW-selector.

tigran123:
On the removal of highlights --- it should be very similar to removal of bookmarks, so leave it with me, I'll have a look.

tigran123:
It (removal of highlights) is actually slightly more complicated than deleting bookmarks, but still should be doable, consider me assigned to this (item 2) on your list).

As for item 1) I can also have a look, but I don't promise.

BtwI have to go somewhere in a few minutes and will be back in the late afternoon, so I may not have a chance to look at any of this until tomorrow.

eLiNK2gl:
@tigran123  That would be great. Thanks.

I suppose it would be pretty straightforward to use highlighting coordinates, so we might at some time make use of the following project -> http://code.google.com/p/kindle-annotations/ (at least, for pdf files)

tigran123:
That program would be wonderful if it was not written in Java and didn't require some massive Java library for PDF manipulation.

The sort of thing you are asking is even beyond the capabilities of Adobe Acrobat Reader but on par with those of Adobe Acrobat Professional. You don't expect one to be able to embed "Acrobat Professional" in a Kindle, do you? :)

tigran123:
[...]

Issue #294, opened 2012-09-19
imposition of text in info box


eLiNK2gl:
The text from the info box (press 'menu') is overlaid over the book's text (epub), which makes it hardly readable.
Calling up the info box for the second time helps.

tigran123:
@eLiNK2gl Are you sure that you tested it on a real Kindle and not in the emulator? In the emulator this is just a quirk of SDL (press Menu a couple of times to make it go away) but if you reproduced this behaviour on a real Kindle (or DXG) then it is a bug.

eLiNK2gl:
@tigran123 yes, sir

P.S.: mine is a DXG.

tigran123:
I am back. I see what is happening here: it is not a SDL quirk, but a side effect of retrieving TOC on the first press of Menu because the info at the bottom wants to show which section the current page belongs to and this involves generating TOC which then causes a refresh of the current page. Ok, leave this one with me also, I'll see how to fix it...

tigran123:
Ok, I have fixed this bug, so it will be in my next pull request which will also include the highlight deletion from the list.

I also fixed another bug: if the file has external links in the TOC then Menu info shows the wrong section. This will also be in the same pull request.

Now, I am focusing on the highlight deletion from the list feature which you raised in another issue.

eLiNK2gl:
fantastic, thanks

tigran123:
All fixes are in the pull request #296 now, so I am closing this issue.

PR #296, opened 2012-09-19
Simple cleanups...

tigran123:
... which I didn't want to get mixed up with the more serious changes to follow in the next pull request. In this one we have just:
1. Don't exit on pressing "Back" in filemanager
2. Make info messages about highlight short enough to be displayable.
3. Change "Updating HighLight data..." to Debug() as it is non-blocking and irrelevant to the user (especially because it is displayed on files which have no highlight data anyway and this may cause confusion).
4. Don't crash when user presses ENTER or FW select buttons on empty history, just display an appropriate (short) message.

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -532,7 +532,7 @@ function FileChooser:addAllCommands()
 			self.pagedirty = true
 		end
 	)
-	self.commands:add({KEY_BACK, KEY_HOME}, nil, "Back",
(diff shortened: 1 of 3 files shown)


tigran123:
The bigger commit with more serious fixes (for issues #295 and #294 ) is included in this pull request. The explanations are in the commit message.

Issue #62, opened 2012-03-11
Coding Styles


houqp:
The cleanup work should leave to next release I think.

I went through the Lua wiki and following guidances relavent to our project:
- Booleans 

It can be helpful to prefix Boolean values or functions used as predicates with is, such as is_directory rather than directory
- Function naming
  camelCap
- Variable naming
  - All variables should use underscores: offset_x
  - global variables should be prefixed with `G_`: G_width
- Constants 
  ALL_CAPS
- Module/package naming
  Module names are often nouns with names that are short and lowercase, with nothing between words. (we are doing fine with this)
- i, k, v, and t are often used as follows:

``` lua
for k,v in pairs(t) ... end
for i,v in ipairs(t) ... end
mt.__newindex = function(t, k, v) ... end
```
- Class names

Class names (or at least metatables representing classes) may be mixed case (BankAccount), or they might not be. If so, acronyms (e.g. XML) might only uppercase the first letter (XmlDocument). 

Using of TOC breaks this ;P

**Some other tips I found might be useful:**
- Use locals rather than globals whenever possible. 
- End terminator

Because "end" is a terminator for many different constructs, it can help the reader (especially in a large block) if a comment is used to clarify which construct is being terminated:

``` lua
for i,v in ipairs(t) do
    if type(v) == "string" then
        ...lots of code here...
    end -- if string
end -- for each t
```
- Check empty table
  Determine if a table t is empty (including non-integer keys, which #t ignores):

``` lua
    if next(t) == nil then ...
```

Reference: http://lua-users.org/wiki/LuaStyleGuide

houqp:
>  So: underscores are fine to me. But: what style will result in the least amount of needed fixes in the code? :-)

For current code, should be camelCap. :)

But given the better readability of underscores, it might reduce potential  bug in future.

dpavlin:
- tabs for indenting

houqp:
Since other developers are not so enthusiastic on changing style for function name, I suggest we stick to camelCap and use underscores for variable name (which is what most of current code does). Any comments?

hwhw:
works for me :-)

houqp:
I suggest if there is still no objection after the release, we should start cleaning up our code in next release.

dpavlin:
I can't decide if link to this issue should go into README.TXT, should we move it to wiki and include link to that or 
create new file DEVELOPMENT.TXT in repository and distribution (and my slight preference is last one). What do you think?

[...]

PR #307, opened 2012-09-24
Added: Folder Info; fixed: crash in getUnpackedZipSize(); etc.

ghost:
1. filechooser: added possibility to get info about selected folder
2. fileinfo: i) fixed crash in function getUnpackedZipSize() applied to zips which filenames contain spaces; ii) added function getFolderContent() to get the folder size & contents (how many subfolders, files & the kpdfviewer-realted books are inside the selected folder); minor corrections
3. crereader: piping in function ZipContentExt() to avoid temporary file

### DIFF
--- a/fileinfo.lua
+++ b/fileinfo.lua
@@ -36,7 +36,8 @@ function FileInfo:FormatSize(size)
 end
 
 function getUnpackedZipSize(zipfile)
-	local cmd='unzip -l '..zipfile..' | tail -1 | sed -e "s/^ *\\([0-9][0-9]*\\) *.*/\\1/"'
+	-- adding quotes allows us to avoid crash on zips which filename contains space(s)
+	local cmd='unzip -l \"'..zipfile..'\" | tail -1 | sed -e "s/^ *\\([0-9][0-9]*\\) *.*/\\1/"'
 	local p = assert(io.popen(cmd, "r"))
 	local res = assert(p:read("*a"))
 	p:close()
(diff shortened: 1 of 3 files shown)


ghost:
I hope that this time I've missed nothing important. And, nevertheless, I ask you to test the changed files carefully before merging the pull request.

tigran123:
@NuPogodi At line 40 of crereader.lua you probably meant "while true" and not "while tmp" because relying on the (undocumented and thus not guaranteed) fact that tmp will become nil after tmp:close() is a bad idea and you don't need it because you explicitly break out of the loop.

I haven't reviewed the rest of the code, I'll let you know if I have any other comments, but the above should be corrected, imho. Thanks.

tigran123:
@NuPogodi Also, I object to starting two processes "df" and "tail" unnecessarily when the very same single system call statvfs(2) (already present in util.c) is perfectly capable of returning the same information in a more stable and much faster way. Why more stable? Because "df" could be an alias or change output format (I always do alias df="df -Th") and then your parsing would break. Please modify util.df to return also the two extra fields you require instead of all that Lua code. Thank you.

But I really like your idea of returning info on a directory --- I was going to do it, but you beat me to it. Thank you very much! :)

fileinfo.lua
```diff
@@ -44,63 +45,123 @@ function getUnpackedZipSize(zipfile)
 	return tonumber(res)
 end
 
+function getDiskSizeInfo()
+	local s = {}
+	local tmp = assert(io.popen('df /mnt/us/ | tail -1', "r"))
```
houqp, line 50: I suggest using `'df . | tail -1'` instead of `'df /mnt/us/' | tail -1'`, so it works also for emulator mode.

OK, please ignore this note, @tigran123 's suggestion is better ;P

tigran123:
@NuPogodi  Also, in FileInfo:show() you overwrote my code which used FileExists() function. Why? FileExists() closes the file handle internally, while your version leaves it to the garbage collector, which is less tidy. In fact, an interesting question is --- will the garbage collector actually close an unnamed file handle? I don't remember reading about it in the "Programming in Lua" book, but if you know the answer please let me know. I.e. if one calls io.open(...) inside a function and forgets about it --- will the file be closed (eventually, after some unpredictable amount of time) after returning from the function? I hope so :) Otherwise we'll run out of file handles if we press Right key a few thousand times :)

tigran123:
@NuPogodi Ignore my last comment about FileExists() --- it is not in the master tree, but only in kpdfdjview branch.

ghost:
All ovewritten functions in fileinfo.lua were changed for the only simple reason: the last lua-code changes in fileinfo.lua were just incompatible with the latest binary. If someone made some critical corrections to lua-code (like df) based on the compiled kpdfview-binary, please, make sure that the updated binary is available in the master-brunch download section and, eventually, inform your collaborators about it.

Concerning the idea to compile some system-available functions (like the mentioned above df), I see both advantages and drawbacks. Why? This way is certainly a bit faster (that is imho not so important for rare calls), but... when following this way, we would need to compile the other system utils like (killall, unzip, gasgauge-info, sed, tail, etc.) Do we really want to replace the whole unix system? I dont think do...

ghost:
Almost forgotten the Tigran's question about while tmp do .. end
If one looks on the code carefully, one can see that tmp:close() is directly followed by break. It means that tmp inside the "while-do"-cycle is never closed and, hence, my code while tmp do .. end is just short version of 

if tmp then
while true do
..
end
end

tigran123:
@NuPogodi Concerning your philosophical comments about "replacing the whole unix system", sorry, but a) you are not a person qualified to teach Tigran Aivazian about "replacing the unix system" (unless you are Linus Torvalds or Ken Thompson and are hiding under the pseudonym, and you can't be Ritchie as he died last year ;) and b) they are irrelevant and factually incorrect, as I never suggested that you should do that. I simply pointed you to the flaw in your code and showed you the correct way to do it. If you didn't understand the explanation, that is fine, I'll fix it myself later, if I have a spare minute.

Concerning the while tmp issue --- it was not a question, it was a pointer to a _bug_ in your Lua code and a kindly provided solution. If you didn't understand it --- that is also fine, I'll fix it later (again, only if I have time, as I am very busy with other things right now).

tigran123:
@NuPogodi Concerning the comments on the unavailability of compiled binary --- they are also irrelevant, because as a KPV developer you are expected to be able to compile the binary and inability to do so is not an excuse for replacing good efficient _existing_ C code with those unnecessary _new_ lines of Lua code.

houqp:
> Concerning the idea to compile some system-available functions (like the mentioned above df), I see both advantages and drawbacks. Why? This way is certainly a bit faster (that is imho not so important for rare calls), but... when following this way, we would need to compile the other system utils like (killall, unzip, gasgauge-info, sed, tail, etc.) Do we really want to replace the whole unix system? I dont think do...

@NuPogodi , I think @tigran123 is just suggesting replacing `df` command call with `util.df`, not all the other system utils :) 

I agree with you that speed should not be a major concern here. But for me, the most important advantage for doing so is to avoid the potential `alias` bug. Because the default output from `df` command is so human unreadable that many Unix users will at least ailas it to `df -h`.

tigran123:
@NuPogodi Although I am normally very patient with newbies and never point a finger when someone does or says something stupid, having shown such flagrant disrespect you leave me no choice:

Your comments about "if tmp" when the value of tmp is returned by assert() show that you don't really understand the Lua assert() function. It cannot return nil ever, rather it issues an error if the first argument is false or nil. So, saying something like this:

```
x = assert(v)
if x then...
```

or like this:

```
x = assert(v)
while x do
```

is stupid because x can never be nil. Please look up assert() in the Lua 5.1 Reference manual or just write a simple test program that illustrates the behaviour of assert()

tigran123:
@houqp well, clean implementations of escaping to shell usually ignore aliases. I don't know if Lua does this or not --- but even if it does, one should not rely on it because being undocumented it may change, of course.

But even apart from the alias issue --- my point was that there is already an existing very simple function in util.c which makes the single system call statvfs(2) which returns all the info and all we need to add is just the return of two extra arguments. Why writing a completely new Lua function from scratch in such case?

ghost:
> but a) you are not a person qualified to teach Tigran Aivazian about "replacing the unix system"

Dont you think that the person who names himself by the name is a bit insane?

> If you didn't understand the explanation, that is fine, I'll fix it myself later, if I have a spare minute.

You also understand nothing... I wrote that compiling the code already presented in unix system may save you, say, 10ms per each call that is absolutely not compatible with, for instance, the kindle page refresh. Is 10ms speed gain so critical for you? Then you should pay your first attention to the code of gasgauge-info, since this utility is used MUCH more frequently (each header contains the battery level).

> they are also irrelevant, because as a KPV developer you are expected to be able to compile the binary...

Are you trying to teach me how to spend my time? I've already said that I prefer pure Lua-programming to installing toolchain and learning how to make cross-compillation. Until now it did not prevent me to contribute to kpdf-code properly, did it?

tigran123:
@houqp I just checked, Lua 5.1.4 does ignore aliases when escaping to shell:

```
$ type df
df is aliased to `df -Th'
$ df .
Filesystem     Type  Size  Used Avail Use% Mounted on
/dev/sda5      ext4  858G  360G  455G  45% /home
$ lua
Lua 5.1.4  Copyright (C) 1994-2008 Lua.org, PUC-Rio
> os.execute("df .")
Filesystem     1K-blocks      Used Available Use% Mounted on
/dev/sda5      899108248 376752428 476683784  45% /home
```

However: a) it is not wise to rely on it and b) see my other point concerning replacing _existing_ simple C code with more complex potentially unreliable _new_ Lua code.

houqp:
@tigran123 thanks for telling me this :) And I agree with your points on using util.df.

@NuPogodi and @tigran123 , let's calm down... Please don't post any replies that includes word like "stupid" or "insane". You are not going to able to prove somebody else wrong in a wrong way, right?

Just one last notes on cross-compile:

6 months ago, @hwhw had discussed similar topic with @dpavlin, @traycold and me (refer to his reply in #62), which I totally agree with. I think this is one of the advantages of using Lua to develop KPV: you don't need any knowledge on C or cross-compile or ARM, just edit that Lua file. It makes hacking on KPV so much easier. So I think it is OK for us to have pure Lua KPV developers. 

@NuPogodi , I think @dpavlin has a quite frequent update for dev release in his own [download page](https://github.com/dpavlin/kindlepdfviewer/downloads), you should be able to get the latest binary from there. And if you cant find the latest binary there, please send any of us an email. Sometime people might be too busy doing other things and forget updating the dev release ;p

tigran123:
@houqp I totally agree.

@NuPogodi I suggest that instead of trying to "fix" each other's code we make a reasonable assumption that the other person is knowledgable and experienced programmer and actually _ask_ his opinion (on why he did this or that in a particular way) before going ahead and overwriting his work. Ok?

tigran123:
Being more cooperating and friendly and helpful to one another may seem like "slowing things down" initially, but, imho, at the end it is the only way to productive and enjoyable work.

houqp:
> I suggest that instead of trying to "fix" each other's code we make a reasonable assumption that the other person is knowledgable and experienced programmer and actually ask his opinion (on why he did this or that in a particular way) before going ahead and overwriting his work. Ok?

Cant agree more :)

PR #309, opened 2012-09-24
show pdf links using L key

dpavlin:
I'm stuck, so I'm looking for feedback.

I don't really know how user interface for this feature should look like.

On one hand, using fiveway to select link might make sense, but I would probably like more direct keyboard shortcuts for each link.

Please feel free to implement it, because I am having quite busy week here.

### DIFF
--- a/pdf.c
+++ b/pdf.c
@@ -577,6 +577,59 @@ static int cleanCache(lua_State *L) {
 	return 0;
 }
 
+
+static int getPageLinks(lua_State *L) {
+	fz_link *page_links;
+	fz_link *link;
+
+	int link_count;
+
+	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
+
+	page_links = fz_load_links(page->doc->xref, page->page); // page->doc->xref?
+
+	lua_newtable(L); // all links
+
+	link_count = 0;
+
+	for (link = page_links; link; link = link->next) {
+		lua_newtable(L); // new link
+
+		lua_pushstring(L, "x0");
+		lua_pushinteger(L, link->rect.x0);
+		lua_settable(L, -3);
+		lua_pushstring(L, "y0");
+		lua_pushinteger(L, link->rect.y0);
+		lua_settable(L, -3);
+		lua_pushstring(L, "x1");
+		lua_pushinteger(L, link->rect.x1);
+		lua_settable(L, -3);
+		lua_pushstring(L, "y1");
+		lua_pushinteger(L, link->rect.y1);
+		lua_settable(L, -3);
+
+		if (link->dest.kind == FZ_LINK_URI) {
+			lua_pushstring(L, "uri");
+			lua_pushstring(L, link->dest.ld.uri.uri);
+			lua_settable(L, -3);
+		} else if (link->dest.kind == FZ_LINK_GOTO) {
+			lua_pushstring(L, "page");
+			lua_pushinteger(L, link->dest.ld.gotor.page); // FIXME page+1?
+			lua_settable(L, -3);
+		} else {
+			printf("ERROR: unkown link kind: %x", link->dest.kind);
+		}
+
+		lua_rawseti(L, -2, ++link_count);
+    }
+
+	printf("## getPageLinks found %d links in document", link_count);
+
+	fz_drop_link(page->doc->context, page_links);
+
+	return 1;
+}
+
 static const struct luaL_Reg pdf_func[] = {
 	{"openDocument", openDocument},
 	{NULL, NULL}
@@ -599,6 +652,7 @@ static const struct luaL_Reg pdfpage_meth[] = {
 	{"getSize", getPageSize},
 	{"getUsedBBox", getUsedBBox},
 	{"getPageText", getPageText},
+	{"getPageLinks", getPageLinks},
 	{"close", closePage},
 	{"__gc", closePage},
 	{"draw", drawPage},
--- a/pdf.c
+++ b/pdf.c
@@ -492,9 +492,9 @@ static int getUsedBBox(lua_State *L) {
 		return luaL_error(L, "cannot calculate bbox for page");
 	}
 
-       	lua_pushnumber(L, ((double)result.x0)/100);
+	lua_pushnumber(L, ((double)result.x0)/100);
 	lua_pushnumber(L, ((double)result.y0)/100);
-       	lua_pushnumber(L, ((double)result.x1)/100);
+	lua_pushnumber(L, ((double)result.x1)/100);
 	lua_pushnumber(L, ((double)result.y1)/100);
 
 	return 4;
--- a/pdf.c
+++ b/pdf.c
@@ -623,7 +623,7 @@ static int getPageLinks(lua_State *L) {
 		lua_rawseti(L, -2, ++link_count);
     }
 
-	printf("## getPageLinks found %d links in document", link_count);
+	printf("## getPageLinks found %d links in document\n", link_count);
 
 	fz_drop_link(page->doc->context, page_links);
 
--- a/pdfreader.lua
+++ b/pdfreader.lua
@@ -45,3 +45,16 @@ function PDFReader:getText(pageno)
 	page:close()
 	return text
 end
+
+function PDFReader:getPageLinks(pageno)
+	local ok, page = pcall(self.doc.openPage, self.doc, pageno)
+	if not ok then
+		-- TODO: error handling
+		return nil
+	end
+	local links = page:getPageLinks()
+	Debug("## page:getPageLinks ", links)
+	page:close()
+	return links
+end
+
--- a/unireader.lua
+++ b/unireader.lua
@@ -2883,12 +2883,20 @@ function UniReader:addAllCommands()
 				Debug("shortcuts",SelectMenu.item_shortcuts)
 				local face = Font:getFace("rifont", font_size)
 
+				local page_links = 0
+
 				for i, link in ipairs(links) do
 					if link.page then
 						local x,y,w,h = self:zoomedRectCoordTransform( link.x0,link.y0, link.x1,link.y1 )
 						fb.bb:dimRect(x,y,w,h) -- black 50%
 						fb.bb:dimRect(x,y,w,h) -- black 25%
+						page_links = page_links + 1
 					end
+				end
+
+				if page_links == 0 then
+					showInfoMsgWithDelay("No links on this page", 2000, 1)
+					return
 				end
 
 				Screen:saveCurrentBB() -- save dimmed links
--- a/unireader.lua
+++ b/unireader.lua
@@ -2906,13 +2906,16 @@ function UniReader:addAllCommands()
 				local render_shortcuts = function()
 					Screen:restoreFromSavedBB()
 
+					local shortcut_nr = 1
+
 					for i = 1, #SelectMenu.item_shortcuts, 1 do
 						local link = links[ i + shortcut_offset ]
 						if link == nil then break end
 						Debug("link", i, shortcut_offset, link)
 						if link.page then
 							local x,y,w,h = self:zoomedRectCoordTransform( link.x0,link.y0, link.x1,link.y1 )
-							renderUtf8Text(fb.bb, x, y + font_size - 1, face, SelectMenu.item_shortcuts[i])
+							renderUtf8Text(fb.bb, x, y + font_size - 1, face, SelectMenu.item_shortcuts[shortcut_nr])
+							shortcut_nr = shortcut_nr + 1
 						end
 					end
 
--- a/unireader.lua
+++ b/unireader.lua
@@ -2921,7 +2921,7 @@ function UniReader:addAllCommands()
 						end
 					end
 
-					if link then
+					if link and links[link] ~= nil and links[link].page ~= nil then
 						goto_page = links[ link ].page + 1
 					end
 
--- a/unireader.lua
+++ b/unireader.lua
@@ -1391,10 +1391,12 @@ function UniReader:show(no)
 
 	-- draw links on page
 	local links = self:getPageLinks( no )
-	for i, link in ipairs(links) do
-		local x,y,w,h = self:zoomedRectCoordTransform( link.x0,link.y0, link.x1,link.y1 )
-		if link.page then -- skip non-page links
-			fb.bb:invertRect(x,y+h-2, w,1)
+	if links ~= nil then
+		for i, link in ipairs(links) do
+			local x,y,w,h = self:zoomedRectCoordTransform( link.x0,link.y0, link.x1,link.y1 )
+			if link.page then -- skip non-page links
+				fb.bb:invertRect(x,y+h-2, w,1)
+			end
 		end
 	end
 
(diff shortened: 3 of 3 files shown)


houqp:
Since the old code base will be focused on kindle keyboard anyway, I vote for using keyboard shortcuts to select each link.

tigran123:
@houqp But how would you inform the user about shortcuts? Isn't it easier to just draw some small underline under the links and let the user select the one he wants by using FW-selector? And the link with the focus on should be highlighted stronger (bolder?) than the others. I think that this scheme is easier than having to assign shortcuts and worry about pages with 66 links (e.g. my Bibles' TOC pages have exactly 66 local links! :)

dpavlin:
I was thinking about drawing keyboard shortcut on top of beginning of the link.

tigran123:
Oh, and of course the jump should be added to the jump list, so that "Back" key would return us to where the link was.
(But this is obvious anyway)

dpavlin:
@tigran123 adding links to jump list is handled by `UniReader:goto` if it's called without second param which is ignore.

tigran123:
@dpavlin Oh no, that would put a lot of mess on the page. Usually those elements with the link are very small things (like page numbers in a printed TOC) and drawing something on top of them would make them hardly legible.

In fact, I always disabled even colour change or any other indication (via pdfLaTeX or XeLaTeX) of the link's existence to not distract the reader. So the reader only discovers that there is a link by hovering the mouse over it. Ok we don't have a mouse, so I suggest drawing an underline under the link.

dpavlin:
Another approach might be to show user that there are links on page (by showing small L on top right for example) and then destroy whole page with shortcuts if users presses a key :-)

I would really like to avoid having to press fiveway 28 times to get to last link on page, but than again, we might not have enough keys to assign each link it's own key.

houqp:
When I see this, the first thing come to my mind is the vimperator plugin for Firefox. They use keyboard shortcuts to selects all the links on the webpage, it works very well for web browsing.

But @tigran123 is right, small eink device is a little bit different, links might be too small to put a shortcut hint beside...

tigran123:
@dpavlin That is also not optimal because the user will most likely press any key and destroy it and then will regret and wonder "how can I get all those links back?" and will start paging back and forth anxiously...

houqp:
But keep moving fiveway also makes me uncomfortable, especially when you know that the fiveway key is easy to break down...

@dpavlin , what do you mean by destroy whole page with shortcuts?

Another approach is to combine fiveway key with numbers, so users can press 5 and fiveway down, to move down 5 times.

tigran123:
@dpavlin Concerning pressing FW 28 keys --- good point. Let's think about it... what can be done for speed-jumping to any desired link? Hmmm..... can't think of anything short of embedding a little ESP module to read the user's mind and guess which link he is going to jump to next.

tigran123:
@houqp I think he means redrawing the page as if it had no links at all, i.e. as "plain"

houqp:
> Hmmm..... can't think of anything short of embedding a little ESP module to read the user's mind and guess which link he is going to jump to next.

LOL, or you can make use of the microphone built inside kindle to jump by voice control ;P

tigran123:
@houqp Oh, that is a good idea: combining numbers with FW-down/up to jump to the Nth link instead of one by one. Then one can reach Revelation from Genesis in no time! :)

tigran123:
But wait -- number is really "Alt + Top row", so that is two keys. Are you expecting the user to press THREE keys simultaneously?

eLiNK2gl:
@tigran123 for numbers, see the implementation of the goto function

@all the question is how to count the links. What if we have 2 or 3 columns on a page or some other complex design, which could be the result of OCR?

tigran123:
@eLiNK2gl for multiple columns there is an order used by the highlighting code. I.e. the first column is traversed first and then the second. I assume the same traversal order will be used for local links, no?

dpavlin:
I implemented keyboard shortcuts on top of dimmed links and it looks better than I expected.

I had to use strange `1.6` factor to get font size which corresponds to size of link box. From reading source, it seems that `Font:getFace` expects font size in pixels, so I'm probably doing something wrong.

houqp:
@tigran123 , no need to press at the same time, just in sequence, like what we do in Vim. 

But what @eLiNK2gl points out is right, for multi columns documents, it might be a little bit difficult for users to calculate the count of moves they want to reach the link. Because there are too many links in one page... In this case, keyboard shortcut might be a better and faster solution. Since @dpavlin has already implemented it, let's see how it works :p

@dpavlin , great work! I will try it out tonight :)

dpavlin:
I'm really interested in pdf documents which have more than 30 links on page. We could add another 30 using shift and probably 30 more using alt.

Do you have any real-world examples? I can generate some, but I'm wondering does it make sense.

tigran123:
@dpavlin Excellent work! I have only one problem with it: one has to know in advance that there are links on the page. We need some indication of it. As I said there are real world examples which make the links invisible.

Just saw your last message: yes, there are, of course. I will provide you with them in a minute, wait...

dpavlin:
@tigran123 we might use similar navigation for links as we do for highlight (with or without shortcuts ;-), but from looking at code it doesn't seem to me that we can reuse much of it, but @houqp will probably know better...

tigran123:
@dpavlin Ok, try this one:

http://urantia-book.org/downloads/English-UK/Urantia-ENG-UK-A4-twocolumn.zip

Pressing L on the TOC page makes the shortcuts quite unreadable.... But on the 6" eReader version it looks fine:

https://github.com/downloads/tigran123/urantia-study-edition/urantia-study-edition-MinionPro.pdf

tigran123:
@dpavlin Cool, that is much better! Now, one little bug, trying to follow an external URL causes a crash:

```
lua config error: ./unireader.lua:2933: attempt to perform arithmetic on field 'page' (a nil value)
```

Try following the URL on the title page of the above A4 PDF.

tigran123:
But still the shortcuts on the A4 version are quite unreadable. So we definitely need some way of navigating the links other than by shortcuts.

dpavlin:
I decided to hide external urls since we don't know how to send them to browser, and shortcut letters should now be readable. We are still missing a few shortcuts to cover all links in A4 version, but my last tea is wearing out for today so I'll go to bed now...

tigran123:
and for testing double-column TOC with plenty of links here is one:

https://github.com/downloads/tigran123/russian-synodal-bible/Russian-Synodal-Bible-USletter-twocolumn.pdf

dpavlin:
Hm, this is nice example. What about using fiveway to move shortcuts around? This might be easier than using shift and/or alt which require strange key acrobatics to press?

tigran123:
Yes, wonderful idea, go for it! I assume you are going to move the whole
bunch of them in one keypress so as to cover as many links as possible with
fewest keystrokes.

houqp:
Just tested this feature in EMU mode, looks really good :)

> What about using fiveway to move shortcuts around?

Good idea! This is so far the best solution I think.

tigran123:
@dpavlin Are you sure that the last three shortcuts work correctly: dot ".", Sym and Ent don't work in the emulator.

Btw, let me tell you something: tracking your remote branches (via git) was such an enjoyable exercise that I am very close to abandoning the kpdfdjview idea in favour of strengthening crengine support (I can cause it to demand 6GB of RSS, i.e. make its faulted-in address space grow practically infinitely!) This is so that we all work on the same master tree as it is great fun! @NuPogodi this has good news for you also, as it would mean that you will get a freshly build binary almost every day! :)

tigran123:
this would make KEY_DOT and KEY_SYM work:

```
$ git diff unireader.lua
diff --git a/unireader.lua b/unireader.lua
index c149563..972f6ee 100644
--- a/unireader.lua
+++ b/unireader.lua
@@ -2923,9 +2923,9 @@ function UniReader:addAllCommands()
                                                elseif ev.code >= KEY_Z and ev.code <= KEY_M then
                                                        link = ev.code - KEY_Z + 21
                                                elseif ev.code == KEY_DOT then
-                                                       link = 29
+                                                       link = 28
                                                elseif ev.code == KEY_SYM then
-                                                       link = 30
+                                                       link = 29
                                                elseif ev.code == KEY_BACK then
                                                        goto_page = unireader.pageno
                                                end
```

but not KEY_ENTER.

dpavlin:
On Tue, Sep 25, 2012 at 02:06:35AM -0700, Tigran Aivazian wrote:

> Btw, let me tell you something: tracking your remote branches (via git) was such an enjoyable exercise that I am very close to abandoning the kpdfdjview idea in favour of strengthening crengine support (I can cause it to demand 6GB of RSS, i.e. make its faulted-in address space grow practically infinitely!) This is so that we all work on the same master tree as it is great fun!

I'm all for it :-)

tigran123:
After my pull request #315 is merged kpdfdjview will be gone for good and I'll focus on the master tree :)

dpavlin:
I added implementation of fiveway move (up/left and right/down do the same thing) and if @tigran123 doesn't find another stupid mistake on my part you can merge it to master.

tigran123:
Great, I will test it now.

tigran123:
It crashes on DjVu files like so:

```
lua config error: ./unireader.lua:1394: bad argument #1 to 'ipairs' (table expected, got nil)
```

tigran123:
Also, the jump history is slightly broken, namely do this:
1. goto a page with links
2. press L
3. select a link pointing to a page with links and press the shortcut, say Q
4. press L
5. again press Q
6. Now press Back. Instead of returning one level back it returns all the way to the original page.

To test this just use this file:

https://github.com/downloads/tigran123/urantia-study-edition/urantia-study-edition-MinionPro.pdf

and follow the link to any paper --- the first page will always contain a mini-TOC with links to the sections to which you can jump.

tigran123:
One other comment (I know it means adding code for splitting the range up to "L" and after "L", sorry)  --- it would be nice to sacrifice "L" as a valid shortcut for the sake of making the second press "L" do what currently you have to press "Back" for. I.e. pressing "L" the second time should, imho, make the link shortcuts disappear (somewhat like pressing "Menu" the second time).

tigran123:
It crashes on fb2 files as well:

```
# command to execute: [46]L: page links
# getPageLinks not supported in this format
lua config error: ./unireader.lua:2884: bad argument #1 to 'next' (table expected, got nil)
```

Maybe we should check the return value of :getPageLinks() method?

tigran123:
Interesting, after your latest commit the jump history issue got fixed as well. But I thought only crashes on fb2/djvu were supposed to be fixed by it...

tigran123:
One last bug: pressing L on a page with (invisible) URL link does activate shortcuts --- it should not.

tigran123:
Also, make sure you deal with the case when there are both URL and local links on the page --- only the URL links should be ignored but the local ones should get proper shortcuts starting from Q.

dpavlin:
Shortcuts are ignored if they are on next page - take a look in `UniReader:goto` -- there is `math.abs(self.pageno - no) > 1` which ensures that normal page turns won't be inserted in jump history, but this also affects any goto which points to next page...

tigran123:
Ah, so for the mini-TOC I was following a link to the section that happened to start on the next page, ok, I understand.

tigran123:
To test URLs, normal links and "invalid links" all on the same page, here is the test PDF file I have just produced for you:

https://github.com/downloads/tigran123/urantia-study-edition/test-url-invalid-links.pdf

tigran123:
With this fix URLs eat shortcuts, thus decreasing the number available for normal links....

tigran123:
Also, another tiny bug: pressing L on a page with lots of invalid links does NOT display a message "No links on this page", it just silently does nothing making the user stare at the page carefully trying to see where is this little link hidden :)

dpavlin:
I'm not quite happy with last commit. There is still a problem if only link on page is link to URL. In that case, L won't say that there are no links on page, but won't show any shortcuts and only way to exit is to press back which is very non-obvious.

I have to think about it. On one hand, `getPageLinks` might skip all non-page links by itself, but I would love to provide url links if we ever figure out a way to send links to browser...

tigran123:
Yes, I should correct my previous message --- the case is not "lots of invalid links" but "one or more URL links and NO normal links".

tigran123:
Yes, some users will think the application hang and open "pressing L hangs the viewer" issues :)

tigran123:
Actually, about sending links to browser, I was toying with the idea of execing an external program (say, in Advanced Fileexpert mode) to bind an "E" key to "Edit the file". This would be very useful and we could invoke an external editor like vi (or mcedit or mcview, though I had to put them in my rootfs manually as VFAT doesn't support symlinks). And if we could exec an external program then we should be able to exec the browser...

Actually, even better would be to have elinks compiled for Kindle that we could exec. Ah, where is @NuPogodi who mightily compiled mc (Midnight Commander) for which I am truly grateful! Can you (@NuPogodi) compile elinks (or lynx) too? When I tried compiling vim I was stuck on ncurses for ARM and so instead of fixing it I just found a ready compiled binary on github....

tigran123:
Ok, this change works, i.e. we get "No links on this page" message. But the eating of shortcuts by invisible URL links issue is still there.

tigran123:
Just to let you know that I am testing your pdf-links branch _merged_ into [my local] master on a Kindle 3 and Kindle DXG (not emulator). I will let you know tomorrow if anything is wrong.

And anyone wishing to test it as well, here is the full package:

https://github.com/downloads/tigran123/kindlepdfviewer/kindlepdfviewer-c4b04d9.zip

tigran123:
Here is a bug:
1. Go to a page with more than 30 links but less than 60 links (e.g. TOC of A4 Urantia Papers, but NOT TOC of A4 Bible).
2. Press L to get the first 30 links.
3. Press FW down to get the next lot that would cover all the remaining links.
4. Now press a key corresponding to a shortcut _beyond_ the range assigned to the actual links. E.g. if the range is Q-J then press K or any key after K.

Observe the crash:

```
lua config error: ./unireader.lua:2971: attempt to index a nil value
```

You have to limit shortcut_offset with the index of the last visible shortcut before indexing links[link + shortcut_offset], i.e. ignore the requests for shortcuts beyond it.

dpavlin:
Damn. I always have to repeat to myself: reading your own commits with `git log -p` is your first review. I should have done something similar when I was adding `shortcut_offset` back in 01b1f5ba6608c59b9907d1c271d00ce35a143be2

tigran123:
Your fix works. The only remaining issue that I see is the "eating" of shortcuts by URL links.

dpavlin:
If you are happy with it, feel free to merge this pull request...

tigran123:
Ok, will test thoroughly now, wait...

tigran123:
Hmmm, now I can't use any of the shortcuts after pressing fiveway down. I.e. only the first lot of shortcuts works.

tigran123:
Yes, works now. I'll test it a bit more and then merge it if I can't break it any more :)

tigran123:
In the general Wiki/Downloads page I added a link to my builds (after @dpavlin's builds) so people can check in both places and choose the most recent build to try. And I have just added today's build which has the pdf local links implementation.

eLiNK2gl:
@dpavlin looks good. You're the man. Thanks.
The only suggestion is to adjust the description in the active hotkeys' list to read 'show internal page links'.

tigran123:
@dpavlin And I have a minor suggestion too: please comment out the printf() in the pdf.c (which is executed on each page turn), in line with the way this is done in djvu.c and other printf() in pdf.c. If need be, this can be uncommented for development purposes, like the other ones.

On a slightly related topic: do you (or anyone) know how to disable those massive amounts of debug code in crengine? There are so many DEBUG defines that I couldn't ascertain which one of them is the master. We have NDEBUG=1 (which I thought means "no debug") but we still have PLENTY of debug messages coming out of crengine which I would like to be able to disable for the production release. For the Lua side all we need to do is just to _not_ pass "-d" to reader.lua in launchpad/kpdf.sh, that is all.

PR #310, opened 2012-09-24
Fix a crash in getDiskSizeInfo() (in emulator) and use util.df function.

tigran123:
1. When executed in the emulator the viewer will crash if you press
   Right on any zip file in the filechooser. This is because the assertion
   on "df /mnt/us | tail -1" will fail as "/mnt/us" is normally
   non-existent.
2. The proper (faster, reliable, portable) way of obtaining the number
   of total and free bytes on a filesystem is by using the statvfs(2)
   system call via util.df Lua interface, see util.c.
3. Removed the "used" field in the function's return as unused and
   unneeded.

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -35,18 +35,15 @@ function CREReader:init()
 		self.default_font = default_font
 	end
 end
+
 -- inspect the zipfile content
 function CREReader:ZipContentExt(fname)
 	local i, s = 1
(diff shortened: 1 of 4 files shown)


tigran123:
I added the fix for "while tmp" stuff in creareader.lua in the same pull request as all this is simple cleanup, really.

tigran123:
Please note the latest commit contains the fix for the serious issue #311 (TOC crash on broken fb2 files) so, not all commits in this request are cleanups.

PR #312, opened 2012-09-24
crereader: avoid the 2nd page redraw after jump to selected TOC-entry

ghost:


crereader.lua
```diff
@@ -197,8 +197,7 @@ function CREReader:gotoPercent(percent)
 end
 
 function CREReader:gotoTocEntry(entry)
-	self:goto(entry.xpointer, nil, "xpointer")
-	self:redrawCurrentPage()
+	self:goto(entry.xpointer, true, "xpointer")
```
houqp, line 200: You mean we should not add en entry to jump history on toc entry jump?

dpavlin, line 200: You changed is_ignore_jump to true (middle param).

I don't know if users would expect back key to return at TOC location, but I suspect they would...

tigran123:
@NuPogodi Your "fix" is incorrect. The commit message is false --- you did not "avoid repeated page redraw" (as if the person who wrote that line code (who just happens to be myself :) didn't know better) --- in fact you have MODIFIED the behaviour of the application and created divergency between crereader and pdf/dfjvu behaviour. Namely, you have passed "true" as the flag to the goto which tells it NOT to add it to the jump list. But we do want it in the jump list, as is the behaviour in PDF and DjVu case, see the code in UniReader:gotoTocEntry().

So, please always check the PDF and DjVu behaviour when you do something in crereader. I am rejecting this pull request as invalid.

tigran123:
@dpavlin and @houqp Oops, I didn't see your comments, sorry. I was in a terminal re-testing @NuPogodi 's changes just to make 100% sure that my reading of the code is correct.

ghost:
I told about avoiding double redrawing the page, but actually in the current version it happens 3 times. Namely, since someone (I'd like to know who did it?) added the function self:redrawCurrentPage() to UniReader:addJump(), the crrent page is redrawed once [the 1st redraw], then the function self.doc:drawCurrentPage(self.nulldc, fb.bb) in CREReader:goto() [the 2nd redraw] and again self:redrawCurrentPage() added (by Tigran) to CREReader:gotoTocEntry(entry) [the 3rd redraw] .

Changing the middle parameter to "true" causes the function CREReader:goto() not to call UniReader:addJump() [now containing self:redrawCurrentPage() = 1st redraw], while removing the corresponding line self:redrawCurrentPage() in CREReader:gotoTocEntry(entry) allows to avoid the 3rd redraw. 

One can readily see this visually and confirm it via estimating the processing time in the current version & after my dirty fix in the following way

function CREReader:gotoTocEntry(entry)
        local t = os.clock()
    self:goto(entry.xpointer, nil, "xpointer")
       self:redrawCurrentPage()
        t = os.clock()-t
        showInfoMsgWithDelay(string.format("%.2f(s)", t), 1000, 1)
end

vs 

function CREReader:gotoTocEntry(entry)
        local t = os.clock()
    self:goto(entry.xpointer, true, "xpointer")
        t = os.clock()-t
        showInfoMsgWithDelay(string.format("%.2f(s)", t), 1000, 1)
end

The processing time in my version should be approximately  3 times smaller

ghost:
BTW, in the current version (with self:redrawCurrentPage() in UniReader:addJump()) any jumps (say, Alt-Q..P) within crereader do not refresh the page... It requires one more call of self:redrawCurrentPage(). Isnt it?

> So, please always check the PDF and DjVu behaviour when you do something in crereader.

The same is valid if written in opposite order - before doing something with unireader, please, always check how it influences crereader behavoiur

houqp:
Ok, I see your point. I agree that `self:redrawCurrentPage()` should not be added in `addJump()` method, which should focus on data processing instead of GUI stuff. This line is probably added by me long time ago ;p

@tigran123 , could you briefly explain why we need to call `self:redrawCurrentPage()` at the end of `CREReader:gotoTocEntry(entry)`? If I remember correctly, it was a fix for a bug in the collapse TOC feature?

ghost:
@houqp , I'm afraid this line was added just recently, 5 days ago - namely, when I enjoyed my holidays
https://github.com/hwhw/kindlepdfviewer/commit/f11a0067750b46222985258c8ca974a995d22f0a (see line 1481)

tigran123:
@houqp Please wait, I'll retest and see what went wrong there and why.

@NuPogodi Please stop the finger-pointing and witch-hunting, for doing so becometh not a true gentleman. "Who did this?" is not a valid question in the friendly free software community, but ask rather: "Why was this done?".

tigran123:
Ok, I see what is happening here and will be happy to explain. When I added two calls to self:redrawCurrentPage() (one each to addJump() and addBookmark()) there was already one such call in _getTocTitleByPage() (this is probably the one which @houqp referred to when saying "I probably did this a long time ago") and I naturally assumed that there was a good reason for it. And so there was at the time. However, as this function (_getTocTitleByPage()) was also called on the first press of Menu (to retrieve TOC via showMenu() code path) it was causing the Menu info text to be overlayed on the page, so I removed the page redraw from _getTocTitleByPage() BUT added it to ALL callers thereof which are NOT on the showMenu() codepath, namely to addJump() and addBookmark() and later also to CREReader:gotoTocEntry(). In the first approximation this is the right thing to do when you are modifying a large program. I.e. first you ensure that all other code paths are unaffected by your change and only LATER you examine to see whether those other code paths could be further optimized AFTER your change.

Now, as it turns out, the page redraw in addJump() and also in CREReader:gotoTocEntry() code path is NOT needed --- only the one added (or, more correctly, moved to) addBookmark() is, so those two can be safely removed.

In summary, my fix for the Menu overlay issue #294 allowed to "further fine-grain" the page redraws enough to optimize away two of them, which @NuPogodi has now done. Therefore, I recommend merging this pull request --- it is a valid optimization, made possible by my earlier fix. This would have been a good example of team work if @NuPogodi learned to refrain from childish finger-pointing and false accusations like "this breaks all blah-blah...".

Let me clarify that by "this pull request" I really meant the #313 (removed two self:...) and NOT #312 which was rightly rejected as invalid.

PR #313, opened 2012-09-24
removed two self:redrawCurrentPage() functions

ghost:
1. in unireader.lua > addJump() - this line seems to break all the jump-related functions in crereader.lua (i.e. correct refresh of the target page in document)
2. in crereader.lua > gotoTOCentry() - this line seems no longer needed
   I hope that it should restore, at least partially, jump-related functions in crereader. 

Please, test it before merging... rejecting this pull request is also quite easy. 
And, PLEASE, let me know who added self:redrawCurrentPage()  to UniReader:addJump()

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -198,7 +198,7 @@ end
 
 function CREReader:gotoTocEntry(entry)
 	self:goto(entry.xpointer, nil, "xpointer")
-	self:redrawCurrentPage()
+	--self:redrawCurrentPage()
 end
 
 function CREReader:nextView()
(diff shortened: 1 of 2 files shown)


tigran123:
@NuPogodi Concerning your enquiry "Who did this?" please see the advice in the issue #312.

tigran123:
@houqp and @dpavlin please see the explanation in issue #312 before merging this pull request.

tigran123:
@NuPogodi A word of avice on commits: these should be grouped not by files, but by the logical change. For example, removing these two page redraws is a single change logically, so it should be in a single commit, not two (one per each file). This makes it much easier to analyze when reviewing the code, instead of having to manually track which commit is logically linked with other commits. Please do this in the future. Thank you.

PR #315, opened 2012-09-25
Pulling bits from kpdfdjview branch (plus other fixes)

tigran123:
This pull request will contain the bits I am pulling back from the kpdfdjview branch into the master. For explanation see the message in each individual commit. It also contains some other minor fixes found today.

Please don't merge it just yet as I am going to add a few more commits to this request, each for a logically separate change for easy peer reviewing.


tigran123:
Ok, I think I am done for now. I also wanted to remove the callback() stuff (see issue #317) but this can go into the next pull request as I am NOT going to try removing someone else's code without first ascertaining the reason and need for it.

PR #318, opened 2012-09-25
Show real time, not cpu time when making screenshots.

tigran123:
@NuPogodi this is your code, so I invite your comments and if you have strong justified objections to this change then this pull request should not be merged. So, here is what I am proposing --- to show the accurate real time spent when making a screenshot instead of the approximate estimate of the CPU usage. The reason is: when making a screenshot the program does not spend all of the time making calculations but also communicating with the eInk controller and eventually writing the dump to the flash device. Therefore, showing a mere "estimate of CPU usage" is misleading, but showing the real time is more useful. What do you think?

Oh, and also I commented out the currently unused function for making screenshots in pgm format.

The other change in this pull request is trivial cleanup (get rid of compile warning and stop crashing screenshots in the emulator).

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -45,7 +45,14 @@ function CREReader:ZipContentExt(fname)
 		if i > 3 then tmp:close(); break; end
 		i = i + 1
 	end
-	return s and string.lower(string.match(s, ".+%.([^.]+)"))
+	if s then
+		local ext = string.match(s, ".+%.([^.]+)")
+		if ext then
+			ext = string.lower(ext)
+			return ext
+		end
+	end
+	return nil
 end
 
 -- open a CREngine supported file and its settings store
--- a/reader.lua
+++ b/reader.lua
@@ -1,4 +1,4 @@
-#!./kpdfview
+#!./kpdfview.emulator
 --[[
     KindlePDFViewer: a reader implementation
     Copyright (C) 2011 Hans-Werner Hilse <hilse@web.de>
@@ -153,16 +153,11 @@ if ARGV[optind] and lfs.attributes(ARGV[optind], "mode") == "directory" then
 	local running = true
 	FileChooser:setPath(ARGV[optind])
 	while running do
-		local file, callback = FileChooser:choose(0, G_height)
-		if callback then
-			callback()
+		local file = FileChooser:choose(0, G_height)
+		if file then
+			running = openFile(file)
 		else
-			if file ~= nil then
-				running = openFile(file)
-				print(file)
-			else
-				running = false
-			end
+			running = false
 		end
 	end
 elseif ARGV[optind] and lfs.attributes(ARGV[optind], "mode") == "file" then
--- a/reader.lua
+++ b/reader.lua
@@ -1,4 +1,4 @@
-#!./kpdfview.emulator
+#!./kpdfview
 --[[
     KindlePDFViewer: a reader implementation
     Copyright (C) 2011 Hans-Werner Hilse <hilse@web.de>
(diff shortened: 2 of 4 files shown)


ghost:
The CPU-time message remained after the performance comparison between the previous Lua vs LuaJIT. I thought that either @dpavlin or @houqp will remove it at merging this code. In principle, I think that it has no sense to show anything except for just "Making screenshot" or -- it might be another option -- even to make it possible for user to replace this and some other not too important screen message(s) by the TTS-voice informing about the performed operation.

tigran123:
I think TTS-voice message would be a bit of an overkill for this :)

Not showing anything at all is not a good idea (but you are not suggesting it) because the user would wonder whether he pressed Shift-P or not and will press it again, just in case.

Showing just a plain message "Making screenshot" has practically the same overhead as showing the message with the time spent, but the latter adds a bit of info. So I take it you have no serious objections to this part of the pull request.

And yes, in the future, this message (and many other scattered throughout the code) could be disabled depending on the value of some configurable verbosity level option, I agree.

ghost:
> I think TTS-voice message would be a bit of an overkill for this :)

Why? Voice info is relatively easy to organize; it does not require user to wait when popup window will be closed (not all users know about a possibilty to close it by pressing any key); it is not too restricted by the length of the transferred message as the visual info (at least, in the current InfoMessage:show()-version). 

I generally thought about
1) separation of messages on various importance levels and
2) user-configurable value that tells kpdfviewer what to do with messages for each level - either a) visual notification or b) TTS-voice or c) both of them or d) nothing. Let the user decide what he wants to use... If, certainly, it has the sound-equipped device.

houqp:
Using TTS-voice is an interesting idea. But it is not a trivial task, thus I suggest we discuss it in a new standalone issue ;-p

Anyway, if @NuPogodi has no objection to this request, please kindly merge it :)

tigran123:
@NuPogodi in the latest commit I have modified your code in ZipContentExt(). Please feel free to comment or object or approve.

The offending zip file is:

http://www.mupdf.com/download/mupdf-thirdparty.zip

Btw, a more complex alternative to my fix would be to keep scanning the output of unzip -l until the first regular file entry and then extract its extention to decide whether to pass it to crengine or not.

tigran123:
Btw, @NuPogodi the way @dpavlin and @houqp merge our pull requests is simply by clicking on this big green "Merge pull request" button, so we cannot expect them to edit our changes --- they either: a) accept them as is or b) tell us to change something (which means we add another commit that updates the same files) or c) reject it completely.

dpavlin:
We also have quite relaxed merge policy -- if it's good enough for original author, it's good enough for us :-)

I will read through all changes before merging them, but to be honest, I mostly won't test them until they land in master. Than again, if I created bug by merging code I will also try to submit another pull request which fixes it.

dpavlin:
Hm. I wasn't clear enough -- replace "original author" with "developer who submitted pull request".

+1 from me for merge

houqp:
> I will read through all changes before merging them, but to be honest, I mostly won't test them until they land in master. Than again, if I created bug by merging code I will also try to submit another pull request which fixes it.

Same here, I usually prefer merge then send fixes in new request :)

tigran123:
@NuPogodi Btw, of course it only makes sense to continue scanning the zip if crengine is actually able to access it eventually. I.e. even if inside some subdirectory there is a valid fb2 (or epub etc) file, we should only consider it valid if it is visible to crengine.

tigran123:
Ok, I'll merge this pull request now to not hold up things, but if @NuPogodi wishes to further enhance the ZipContentExt() function to deal with zip files that contain subdirectories he can do so in a separate pull request, as the change in the current fix at least fixes the crash so it is self-consistent and complete in itself.

PR #325, opened 2012-09-26
Various fixes.

tigran123:
This pull request contains a fix for the crash when trying to delete a non-existent entry in file history.

Also, the fix for the need to redraw page after displaying "Already first jump!" on pressing Back on a freshly opened (i.e. no TOC retrieved yet) file.

Also, just two minor cleanups in the Screen:fb2bmp() function, nothing major:
1. Remove unused assert() around io.open of the input device. The failure
   to open input device is already guarded by the "if inputf" code.
2. Remove unneeded assert() around io.open of the output device. This is
   unneeded because we should not crash the whole application just because
   we cannot write the screen dump (e.g. because filesystem is full, etc) ---
   instead we should exit gracefully.

### DIFF
--- a/filehistory.lua
+++ b/filehistory.lua
@@ -220,6 +220,7 @@ function FileHistory:addAllCommands()
 		"delete history entry",
 		function(self)
 			file_entry = self.result[self.perpage*(self.page-1)+self.current]
+			if not file_entry then return end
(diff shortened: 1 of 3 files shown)


tigran123:
Oops, mixed up self.toc and unireader.toc --- will fix in a minute!

Issue #275, opened 2012-09-11
Manual bbox values lost in two-column (F or Shift-F) modes on Next page.


tigran123:
Here is how to reproduce the problem.
1. Take a real 2-column book with big margins (test/2col.pdf is no good because the margins are so small that the bug is almost unnoticeable).
2. Open a typical page and do "A" to reset to "fit to page" zoom.
3. Press Shift-X and position the bbox boundaries appropriately, i.e. getting rid of all the white margins around the text.
   Then turn to the next page and do the same, thus programming bbox values for both odd and even pages.
4. Press F to start reading. The view is correctly positioned to the top left corner, in accordance with our bbox assignment. (btw, for Hebrew books it should be top right corner, but never mind, we are not talking about Hebrew books right now :)
5. To read the first column keep pressing Down on the 5-way controller until you reach the bottom of the first column.
6. To go to the top of the second column press Right on the 5-way controller. The view is correctly positioned to the top right corner. So far, so good.
7. Keep pressing Down on the 5-way controller until you reach the bottom of the second column.
8. Now, to go to the next page you have two options: either press Right on the 5-way controller or press Next page. Both do the same thing and position the view incorrectly, i.e. showing most of the first column and some part of the second column and a gap in between.

I.e. after turning the page the bbox values are lost. However, if I now press "F" again, then the view is corrected and placed in the proper place.

So, the only annoyance is having to press "F" on every page turn to reposition the view on the top part of the left column. Shouldn't this be done automatically, i.e. without the need to press "F"?

eLiNK2gl:
just checked one of my books.
When I press 'F', the page is placed to the BOTTOM left corner.
Pressing 'Right' on the 5-way controller selects the BOTTOM right corner.

A bit off-topic, on at least one of the pages the bbox is grossly off.
Do I understand correctly that if I deleted the history (using 'Del' in the list of last documents opened) there is no lua file left, which could have contained info on bbox positioning?

P.S.: just got a scrambled picture problem (dotted lines on the display and no text) on exiting kpdfviewer. It is a long time since I saw it last.

tigran123:
On the "scrambled picture" problem --- are you sure you exited in portrait mode? I always get "scrambled picture" when I exit in Landscape mode.

eLiNK2gl:
Yes, I am.
I almost never read in Landscape mode.

As far as I remember, this bug was associated with the auto-rotation feature of the KDXG.
[...]

Issue #236, opened 2012-08-30
Next-page stops working after panning in fit page width and fit content modes.


tigran123:
The file test/tall.pdf is only a single page which is probably why this bug was not noticed before, so you need to do something like "pdftk tall.pdf tall.pdf tall.pdf tall.pdf cat output tall4p.pdf" and test on tall4p.pdf to reproduce it.
1. Open tall4p.pdf on page 1
2. Press Shift-S to switch to "fit content width" (or S for "fit page width" --- results are the same)
3. Use joypad to pan the view down a bit.
4. Press "Next page" button: what you will see on the screen will be exactly the same as before pressing it (because all pages in tall4.pdf are identical), but pressing "G" you find that the page is now number 2, i.e. the page was turned instead of moving screen full down (as before panning).

I expect "Next page" to behave after panning in the same way as before panning, i.e. move down in large chunks (almost screen full). This is because panning is useful in situations where some interesting element (picture or equation) got split between screens and you pan down a bit to shift it to the visible area. After viewing the element you then desire to go on reading as before, so you press "Next page", but, unexpectedly, it starts jumping physical pages...

Also, if I patiently pan the view down to the very bottom of the page and then press "Next page" I expect to see the _top_ of the next page --- instead I see the _bottom_ thereof. And pressing "Next page" again shows me the bottom of page 3 and another "Next page" the bottom of page 4. (pressing "G" I ascertain the current page number).

tigran123:
Just wanted to add that the bug is reproducible on other tall files too, not just the above artificially constructed tall4p.pdf. I.e. I originally noticed it while reading normal books.

dpavlin:
I had similar dilemma when implementing two column mode, but I decided not to change prev/next keys meaning and leave them for page flipping always to have consistent behavior of page keys in all modes.

However, crengine is behaving in the way you are describing (next page gives you next part of document) and I can see benefits of both approaches.

To make thing even more interesting, if we had continuous page scroll described in #190 page panning alone would provide desired effect.

Simplest solution which comes to mind is to change how page keys work in zoom modes which are not fit to screen: next page should reset to top of page and previous to bottom similar to two column mode.

Does any of this make sense, or did I just wrote a huge brain dump?

tigran123:
[...]

PR #327, opened 2012-09-26
recalculate pan offsets on page change #275

dpavlin:
Store last globalzoom_mode in self.pan_by_page so we can restore it to force offsets recalculation on page change

This wasn't quite enough to trigger correct re-panning for page changes when using fiveway to move in two-column mode, issue #275

I considered adding callback to goto method and than transfering it through function calls, but this approach (allthough very questionable) seemed to provide simplier code.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1297,10 +1297,19 @@ function UniReader:setzoom(page, preCache)
 		self.offset_y = -1 * y0 * self.globalzoom * 2 + margin
 		self.globalzoom = width / (x1 - x0 + margin) * 2
 		Debug("column mode offset:", self.offset_x, self.offset_y, " zoom:", self.globalzoom);
+		self.pan_by_page = self.globalzoom_mode -- store for later and enable pan_by_page
 		self.globalzoom_mode = self.ZOOM_BY_VALUE -- enable pan mode
 		self.pan_x = self.offset_x
 		self.pan_y = self.offset_y
-		self.pan_by_page = true
+	else
+		Debug("globalzoom_mode didn't modify params", self.globalzoom_mode)
+	end
+
+	if self.adjust_offset then
+		Debug("self.ajdust_offset BEFORE ", self.globalzoom, " globalrotate:", self.globalrotate, " offset:", self.offset_x, self.offset_y, " pagesize:", self.fullwidth, self.fullheight, " min_offset:", self.min_offset_x, self.min_offset_y)
+		self.adjust_offset(self)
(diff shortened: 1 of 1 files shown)


tigran123:
Very interesting, is it supposed to fix the "panning breaks Shift-S mode" I raised in issue #236 ?

You know I wanted to tell you that _several_ people on the-ebook.org forum have been complaining about this bug (and maybe on mobileread.com too, can't remember), so there is a very high demand for this to be sorted out. I replied on the forum that the code handling all the various zoom modes is quite complicated and I am not brave enough to tackle it, so this task awaits a very brave man... :)

tigran123:
@dpavlin Wait... I have just tested reading in twocolumn mode (via Shift-F) and it is perfect --- the bbox value is NOT lost on changing pages anymore... So the issue #275 can be closed (it must have been fixed by you recent pull request). Then what is the change in this pull request about? Hmmm, I think I am missing something...

dpavlin:
Bounding box was never problem, but we didn't recalculate position on page when page changes (which takes into account bounding box).

Easiest way to reproduce this is to take any two-column paper which have title shifted a little bit down on first page. When you switch to next page (without this patch) you won't see top of column (because we are using pan from first page).

Another good test would be two column document which doesn't have columns centered on page.

tigran123:
Ah, correction to my comment --- I tested on the Bible edition where the two columns are perfectly centered on a page. But now I re-tested with Princeton Companion to Mathematics where columns are way off center and see that the problem is still there, you are absolutely right. (i.e. on changing the page I have to press Shift-F to reset the view to the correct one).

tigran123:
Ok, shall I test this change or you are not quite finished with it?

tigran123:
Testing now. Moving to the next page now works fine, i.e. the position is correct. But having just arrived at the top of the left column of the next page the user is likely to press fiveway "Left" (as opposite to fiveway "Right" by means of which he got there) to go back to the bottom of the right column of the previous page. And this does not work, i.e. one has to press Shift-F to reset the position to the correct one PLUS also move down (to the bottom of the left column) and right (to switch to the right column) and then down again all the way to the bottom of the right column --- i.e. too many extra keypresses...

dpavlin:
Render modes are mess. I really can't explain how old code worked and fact that I wrote it doesn't help.

Hopefully, it's now replaced with code which better describes what it actually does :-)

tigran123:
Wonderful, works perfectly! Merging now.

PR #329, opened 2012-09-26
remember last zoom mode when panning #236

dpavlin:
This tries to address issue #236 so that previous and next keys don't
change meaning when user go into panning mode by pressing fiveway.


tigran123:
Wonderful, thank you! I am going to test it now and if all is fine I'll put the latest build in my dev-build area.

PR #339, opened 2012-09-28
Typo: --disable-largefiles -> --disable-largefile

tigran123:
As a result of a simple typo we were continuing to include LFS
support in djvulibre unnecessarily...


tigran123:
I hope I am not contradicting my own suggestion of code freeze! :)
These changes are very simple and of course I tested them. With a slight stretch of imagination they can be counted as bugfixes. Well, the first one certainly is. The third one is too trivial to even discuss. The second one (cre optimization) is a bit more subtle, but I think it is an accidental omission (i.e. it should have been compiled like the rest, imho), so also counts as "bug"

PR #341, opened 2012-09-29
- Fixes bugs that caused *FLAGS not to be passed properly to sub-makes.

kai771:
- Fixes bugs that would cause  build process to  fail if env CFLAGS
  was defined.
- HOST:= changed to HOST?= to allow specifying HOST variable
  both in front or after make command
- added flags to CFLAGS and CXXFLAGS that allow KPDFV to be built with
  CS/MG toolchains newer than 2007q3. This flags are dependant on proper
  passing of *FLAGS for build process to work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ TTF_FONTS_DIR=$(MUPDFDIR)/fonts
 
 # set this to your ARM cross compiler:
 
-HOST:=arm-none-linux-gnueabi
+HOST?=arm-none-linux-gnueabi
 CC:=$(HOST)-gcc
 CXX:=$(HOST)-g++
 STRIP:=$(HOST)-strip
@@ -27,12 +27,13 @@ endif
 HOSTCC:=gcc
 HOSTCXX:=g++
 
-CFLAGS:=-O3 $(SYSROOT)
-CXXFLAGS:=-O3 $(SYSROOT)
+CFLAGS:=-O3 -D_GNU_SOURCE $(SYSROOT)
+CXXFLAGS:=-O3 -fno-use-cxa-atexit $(SYSROOT)
 LDFLAGS:= $(SYSROOT)
 ARM_CFLAGS:=-march=armv6
 # use this for debugging:
 #CFLAGS:=-O0 -g
+BASE_CFLAGS:= $(CFLAGS)
 
 DYNAMICLIBSTDCPP:=-lstdc++
 ifdef STATICLIBSTDCPP
@@ -58,6 +59,7 @@ ifdef EMULATE_READER
 	endif
 else
 	CFLAGS+= $(ARM_CFLAGS)
+	CXXFLAGS+= $(ARM_CFLAGS)
 endif
 
 # standard includes
@@ -90,7 +92,7 @@ LUALIB := $(LUADIR)/src/libluajit.a
 all:kpdfview
 
 kpdfview: kpdfview.o einkfb.o pdf.o blitbuffer.o drawcontext.o input.o util.o ft.o lfs.o mupdfimg.o $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) djvu.o $(DJVULIBS) cre.o $(CRENGINELIBS)
-	$(CC) \
+	$(CC) $(CFLAGS) \
 		kpdfview.o \
 		einkfb.o \
 		pdf.o \
@@ -169,12 +171,12 @@ cleanthirdparty:
 	-rm -f $(MUPDFDIR)/cmapdump.host
 
 $(MUPDFDIR)/fontdump.host:
-	make -C mupdf CC="$(HOSTCC)" $(MUPDFTARGET)/fontdump
+	CFLAGS="$(BASE_CFLAGS)" make -C mupdf CC="$(HOSTCC)" $(MUPDFTARGET)/fontdump
 	cp -a $(MUPDFLIBDIR)/fontdump $(MUPDFDIR)/fontdump.host
 	make -C mupdf clean
 
 $(MUPDFDIR)/cmapdump.host:
-	make -C mupdf CC="$(HOSTCC)" $(MUPDFTARGET)/cmapdump
+	CFLAGS="$(BASE_CFLAGS)" make -C mupdf CC="$(HOSTCC)" $(MUPDFTARGET)/cmapdump
 	cp -a $(MUPDFLIBDIR)/cmapdump $(MUPDFDIR)/cmapdump.host
 	make -C mupdf clean
 
@@ -187,20 +189,20 @@ $(DJVULIBS):
 ifdef EMULATE_READER
 	cd $(DJVUDIR)/build && ../configure --disable-desktopfiles --disable-shared --enable-static --disable-xmltools --disable-largefile
 else
-	cd $(DJVUDIR)/build && ../configure --disable-desktopfiles --disable-shared --enable-static --host=$(HOST) --disable-xmltools --disable-largefile
+	cd $(DJVUDIR)/build && CXXFLAGS="$(CXXFLAGS)" ../configure --disable-desktopfiles --disable-shared --enable-static --host=$(HOST) --disable-xmltools --disable-largefile
 endif
 	make -C $(DJVUDIR)/build
 
 $(CRENGINELIBS):
 	cd $(KPVCRLIBDIR) && rm -rf CMakeCache.txt CMakeFiles && \
-		CFLAGS="$(CFLAGS)" CC="$(CC)" CXX="$(CXX)" cmake . && \
+		CFLAGS="$(CFLAGS)" CXXFLAGS="$(CXXFLAGS)" CC="$(CC)" CXX="$(CXX)" cmake . && \
 		make
 
 $(LUALIB):
 ifdef EMULATE_READER
 	make -C $(LUADIR)
 else
-	make -C $(LUADIR) CC="$(HOSTCC)" HOST_CC="$(HOSTCC) -m32" CROSS="$(HOST)-" TARGET_FLAGS="$(SYSROOT) -DLUAJIT_NO_LOG2 -DLUAJIT_NO_EXP2"
+	CFLAGS="$(BASE_CFLAGS)" make -C $(LUADIR) CC="$(HOSTCC)" HOST_CC="$(HOSTCC) -m32" CROSS="$(HOST)-" TARGET_FLAGS="$(SYSROOT) -DLUAJIT_NO_LOG2 -DLUAJIT_NO_EXP2"
 endif
 
 thirdparty: $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) $(DJVULIBS) $(CRENGINELIBS)
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ LUALIB := $(LUADIR)/src/libluajit.a
 all:kpdfview
 
 kpdfview: kpdfview.o einkfb.o pdf.o blitbuffer.o drawcontext.o input.o util.o ft.o lfs.o mupdfimg.o $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) djvu.o $(DJVULIBS) cre.o $(CRENGINELIBS)
-	$(CC) $(CFLAGS) \
+	$(CC) \
 		kpdfview.o \
 		einkfb.o \
 		pdf.o \


dpavlin:
Thanks a lot for this, which version of compiler do you recommend?

kai771:
I see no reason not to use the latest CS/MG 2012.03-57. However, it's probably wise to use what you used up until now for the impending release, and switch after more intensive testing is done.
I would like to underline that after incorporating these fixes, using SYSROOT is not needed. It will work out-of-the-box.

To tell you the truth, I think we're relying on TC's default settings here to much. I'd actually use:

CFLAGS:=-O3 -fno-stack-protector -U_FORTIFY_SOURCE -D_GNU_SOURCE $(SYSROOT)
CXXFLAGS:=-O3 -fno-stack-protector -U_FORTIFY_SOURCE -fno-use-cxa-atexit $(SYSROOT)
ARM_CFLAGS:=-march=armv6j -mtune=arm1136jf-s -marm -mfloat-abi=softfp -mfpu=vfp

which would set all flags explicitly, and allow it to work with Linaro TC too (judging by NiLuJe's reports - as stated in my posts on MR forum, Linaro pre-built binaries aren't compatible with K3, one would need to build it himself to make it compatible). I actually hesitated with commiting changes to CFLAGS and CXXFLAGS at all, thinking maybe they should be separated from bug fixes, but eventually decided that this much won't hurt, and will simplify setting up the dev environment to possible new developers.

tigran123:
@kai771 This is a very serious change and would require testing. Have you tested that it doesn't break anything with x-tools chain? And those more specific -march, -mtune etc --- do these match Kindle 3 or Kindle DXG or what? Are these 100% accurate, i.e. we don't want a crash because of non-existent instruction...

But if you've done all this research already then looks very cool, thanks! :)

Btw, passing ARM_FLAGS to CXX_FLAGS I was going to do as well, but decided it is too dangerous as crengine and djvulibre would therefore be compiled completely differently and thought it is best to leave until after the release. But maybe we should take the plunge and test it thoroughly now...

tigran123:
@kai771 I see you have a special "makefix" branch for this, so it is easy to test. I'll do it now and let you know if all is fine (or otherwise).

kai771:
I did test with CS 2007q3 (it also works out of the box, the same as x-tools, without the changes I proposed now) and it didn't break anything (as much as I could see). I also tested 2012.03, and compiling in emu mode. All appears to work fine, but please do test it yourself too. I must admit, I didn't test with versions 2008q3 up until the latest, but I expect it to work the same, for the main problem was not passing the parameters properly. 

Main point in the fix is introduction of BASE_CFLAGS, which allows us to pass options without ARM part to portions of make process that use native compiler. I only have K3, so I could test only on K3. Please test on DXG.

Extensive changes to *FLAGS posted in previous post are NOT contained in the commit. Commit contains only this:

CFLAGS:=-O3 -D_GNU_SOURCE $(SYSROOT)
CXXFLAGS:=-O3 -fno-use-cxa-atexit $(SYSROOT)
ARM_CFLAGS:=-march=armv6

It is quite harmless, and won't break anything, so I'd suggest you accept this for now (it will allow latest CS 2012.03 to work out of the box), and the rest after more extensive testing.

tigran123:
Yes, I noticed this when I looked at the commit.

Ok, I have tested your commit on K3 and KDXG --- I even measured the page rendering times for PDF and DjVu (with ms precision) and they are practically identical (i.e. negligibly fluctuate in both directions a couple of milliseconds). I see no harm in merging this now, but I will leave to @dpavlin and @houqp to decide whether it is ok to merge now because my environment is a 64bit Fedora Core 17 (plus x-tools) and so (presumably) it differs from their environment. But I vote for merging, it is a valid change --- I only wish it actually produced some effect on the execution times :)

tigran123:
Oh, wait passing CFLAGS to the linker is meaningless, remove it.

kai771:
Hm, it might be faster for you to take a look at MR thread, where I posted my results as I made them.

To answer your question, passing CFLAGS to link stage is added in the end, while I was trying to make it work with Linaro. I went through Makefile, and added passing of CFLAGS to all gcc invocations. It's not needed in link stage, but it doesn't hurt, so I left it, just to be sure.

As I understood, a lot of people use env *FLAGS when doing their compiling (I don't mean for kpdfv exclusively, but for any build process). Decent apps take those flags into considereation.
KPDFV sets all flag explicitly, since it uses :=. It probably would've been better to use ?= for those, but I didn't want to change too much. Besides, I wanted KPDFV to compile on everyone's box, regardless of what *FLAGS they have set.

Now, if there is an env var set, make will automatically export that var to it's sub-makes. If env var is not set, make will not export those vars. That's why original Makefile didn't work when someone had env CFLAGS.

The situation here is a bit more complicated since we need to use 2 different values of CFLAGS - one with ARM part, one without. Thus introduction of BASE_CFLAGS. Since we have to use 2 different values of CFLAGS, we need to switch the value CFLAGS has. If we didn't have that, after initial assignment of CFLAGS values, all the other makes will get them from the env. Because we use 2 different values, we have to specify CFLAGS before calling sub-make.

So, in my opinion, in linking stage CFLAGS aren't needed, but in normal circumstances gcc would get them from env anyway. Adding them explicitly makes sure they are passed, althought they aren't really needed. Just for consistency.

That's how I see it anyway. I may be wrong, as I said numerous times, I'm just a noob, and don't have any experience.

tigran123:
Even if I have an exported environment variable called CFLAGS, gcc would NOT insert it into the linking stage, so my point remains valid and the addition of $(CFLAGS) to the last (linking) stage of kpdfview build should be removed. It does not introduce any "consistency" but rather this is something very unusual (using CFLAGS in place of LDFLAGS) and done for no reason (neither good nor bad, just none whatsoever).

kai771:
In that case I stand corrected. I hope it's alright now.

tigran123:
Another doubt I have is whether it is a good idea to pass "-O3" to LuaJIT, because in its own sources it says that "-O3" is no longer a good idea (due to i-cache bloat) and that "-O2 -fomit-frame-pointer" produces better code. So, I am weary to change luajit's compilation against it's own author's recommendation.

Now, concerning passing "-O3" to djvulibre --- it automatically selects this internally anyway (via ./configure process which also tests whether g++ accepts -O3).

tigran123:
On the other hand, passing -march=armv6 and -fno-use-cxa-atexit to djvulibre compilation may well be a good idea.
Here is what you could do, @kai771 if you feel the urge to do something useful: investigate what is the most accurate -march/-mtune combination for Kindle 3 and let us know. We'll gladly do some benchmarking and decide on the most optimal setting for this specific hardware. And this should be done carefully, passing to individual item's compilation (to hell the "elegancy" and "consistency", efficiency is what matters here), e.g. we may choose to pass it to djvulibre but not to luajit, for example.

Btw, it is very easy to measure the time spent in some operation --- just look in `screen.lua` file in `Screen:screenshot()` function. Only instead of showInfoMsgWithDelay() you would need to output it with Debug() to crash.log.

kai771:
Sorry, but I have neither time nor expertise to do that. My main motivation for the fix was to enable it to be compiled with recent CS toolchan in a way that would work on K3. To do that, one needs to pass  -D_GNU_SOURCE and  -fno-use-cxa-atexit. The rest of the changes were so that those settings would be passed.

-O3 wasn't introduced by me, it was already there. This is what NiLuJe said when I asked him about -O2 vs -O3:

"As for O2 vs. O3, it has the potential of starting a holy war, but basically: I'm not a huge fan of using O3 as the goto 'default' setting. Use it when you know it's gonna be useful, otherwise, it more often than not _lowers_ the performance (especially on such a tiny CPU), and almost always ends up generating larger binaries. That said, Linaro does recommend defaulting to O3/Ofast on ARMv7 with their TC."

I pretty much trust his expertise, since he helped me a lot in making this work. The full settings that he uses are:

CFLAGS:=-O2 -ffast-math -pipe -fomit-frame-pointer -fno-stack-protector -U_FORTIFY_SOURCE
CXXFLAGS:=$(BASE_CFLAGS) -fno-use-cxa-atexit
ARM_CFLAGS:=-march=armv6j -mtune=arm1136jf-s -mfpu=vfp

Changes I suggested, especially to ARM_CFLAGS are not for the sake of optimization. I think that kpdfv is pretty fast as it is, and that a couple of ms no one would notice. But feel free to differ. It is my belief that these are the settings that x-tools and CS use by default (might be wrong though). However, Linaro defaults to cortex-a8, so that's why it needs to be specified. You may notice the absence of -D_GNU_SOURCE in his version. It's because his TC is built so it defaults to that, but it's needed on CS higher than 2007q3. Otherwise it won't link to K3's glibc.

tigran123:
I take it that "TC" means "toolchain". What is "CS"?

kai771:
CS=Code Sourcery. They are known as Mentor Graphics now. They wrote that arm-none-linux-gnueabi that Makefile still defaults to :).

NiLuJe:
To recap a bit what we discussed in the thread over @ MR and answer a few questions:
- -march=armv6j -mtune=arm1136jf-s -mfpu=vfp is completely accurate for the K2/3 (that includes the DX/DXG).
- -mfloat-abi=softfp is also accurate, because that's what Amazon uses, and we can't mix hardfloat with softfp. That said, ideally, one should be using a non-multilib TC defaulting to the softfp abi (without needing to override/specify this flag)
- Same kind of deal with -marm (with the additional fact that, AFAIK, Thumb1 isn't terribly useful in terms of performance in the first place anyway).
- I stand by my O2/O3 comment quoted earlier ;). If you know O3 will help on some hot paths, or can do profiled builds, then, yeah, that's nice, but other than that I'm really not a fan of O3 as the default ;).
- The Kindle2/3 FW uses glibc 2.5 + GCC 4.1 (or 4.2, don't remember) on Linux 2.6.22/2.6.26. That's the maximum ABI we aim for with our x-tc. I happen to be using a TC based on glibc 2.9 + GCC 4.7. That means a few extra flags need to be passed to avoid pulling some newer ABI (cf. https://svn.ak-team.com/svn/Configs/trunk/Kindle/Misc/x-compile.sh for more details):
- -fno-stack-protector to kill GCC >= 4.3 stack smashing protection
- -U_FORTIFY_SOURCE to kill glibc >= 2.8 fortify (Ideally, that should be in CPPFLAGS, but a few buildsystems are tricky/handle it badly, so putting it in CFLAGS too helps).
- If you want to target FW 2.x completely safely with a Linaro 4.6/4.7 or GCC 4.7 TC:  -mno-unaligned-access
- As kai771 encountered, some TC (CS/MG) may also need -D_GNU_SOURCE in CPPFLAGS (and CFLAGS to make crappy buildsystems happy) to avoid pulling some glibc 2.7 stuff (related to the scanf family of functions and incompatibilities between C99 & G*\* in the handling of some format codes).
- -fno-use-cxa-atexit is also needed for some TC to avoid another CXX ABI backward compat issue ;).

I also happen to be a Gentoo user, so I actually really like it when buildsystems actually honor my env, so yeah, I really like the ?= defines instead of a hard override ;).

Passing CFLAGS to the linking stage is also completely accurate, since we're using the GCC driver, not calling ld manually, and it's actually needed to correctly make use of some stuff (most prominently, LTO).

FWIW, this is the patch I'm using when building kpdf with my TC: https://svn.ak-team.com/svn/Configs/trunk/Kindle/Misc/kindlepdfviewer-makefile-mangling.patch
It mostly does the same thing as this PR, albeit in a way more aimed at my own TC & system, and with mupdf built in release mode instead of the default (which happens to be debug).

tigran123:
@NiLuJe Thank you very much for detailed explanations. Btw, what are "Thumb 1" and "LTO"?

Oh, are you saying that we are building mupdf in debug mode??? That would be a critical showstopper, I'll have a look at your patch and see what's going on. I suggest we postpone the release by a day or so to investigate this, as some of these changes may be critical enough to deserve merging right now.

tigran123:
Indeed, we were building mupdf in debug mode and fixing this reduced the binary size considerably. I am testing all the other suggested changes individually (measuring performance at each step to make sure we are not degrading anything), so please wait. Thanks again to @NiLuJe and @kai771 --- you definitely provided very valuable information, just let us assimilate it slowly and act on it properly :)

kai771:
http://en.wikipedia.org/wiki/ARM_Thumb#Thumb
http://gcc.gnu.org/onlinedocs/gccint/LTO.html

tigran123:
Ok, I have tested the mupdf build -> release and concluded that it is definitely worth merging now. I am testing all the other changes as well and will be picking the needed ones one by one.

tigran123:
Ok, I have extracted what I deemed essential for the current stable release and put in the pull request #345. The rest of your suggestions are also very useful, but they mostly concern building with newer versions of toolchain and so are not critical for the current stable release. I am, however, leaving this issue open so that we remember to merge your other recommendations _after_ the release.

tigran123:
Btw, taking @NiLuJe 's makefile mangling patch "en masse" breaks mupdf ("invalid page size for page 1" error).

NiLuJe:
@tigran123 : With what kind of TC? On every PDF files, or just a few specific ones?

I'll do a rebuild of the current master on my box & check ;).

On a sidenote, I mentioned LTO, but I wouldn't necessarily recommend using it (in fact, I don't), I've had a few interesting suprises the last time I checked if it was stable enough to use... (the bad kind of surprise: bad code generation, resulting in a slew of weird runtime issues ^^).

NiLuJe:
Bleh, there was a typo in the latest refresh of my patch (broke luajit's build).

EDIT: Fixed.

NiLuJe:
@tigran123 : Nope, no issues on my end. You're welcome to try my build with your test files (it's attached here: http://www.mobileread.com/forums/showpost.php?p=2231551&postcount=685), but if it works while yours doesn't, I'm guessing something's wonky with the TC you're using (or at least doesn't like some of the flags for some reason (really old TC in 'soft' float ABI, not 'softfp'?)).

tigran123:
Hmmm, are you suggesting we disable LTO? My third patch in the pull request  #345 enables it, but if you had bad experiences with it we might as well disable it, just to be on the safe side.

tigran123:
The TC I am using is called `arm-kindle-linux-x-tools-glibc2.5-gcc4.2.4.tar.gz` and the size is 45741995 bytes.

NiLuJe:
@tigran123 : It looks like it's a hardfp TC, which might explain the issue. I never got kpdf to behave correctly with a hardfp TC, and it's a bad idea anyway, since Amazon's system is built as softfp.

It looks like it was built using ct-ng according to an old version of my x-compile script (which indeed happened to default to hardfp until I realized what a horrible idea it was [and until upstream switched the loader helper name to avoid ABI mismatch ^^]), so the march/mtune/mfpu stuff should default to the right settings...

But, it's using GCC 4.2, which _doesn't_ correctly support the hardfp ABI IIRC. So that leaves GCC using, err, I don't know which float ABI, probably the old 'soft'... I don't see any default float-abi flags (gcc -Q -v), and I don't see anything either in the attributes of a binary built with it...

You can try passing -mfloat-abi=softfp see if it helps, but I'm not sure it'll actually help, not familiar enough with the state of the float ABI in a GCC version this old...

Also, it won't run on FW 2.x. (Built to handle Linux 2.6.26 minimum, FW 2.x runs Linux 2.6.22).

(FWIW, @hwhw appears to be using CS 2011.09-70, which, being CS/MG, is usually a good choice, and probaby 777% saner ;)).

kai771:
@tigran123 I think (and it's a guess based on some old posts) that hwhw used CS 2011.09-70 with SYSROOT. If you want to try it without sysroot, and without my patch, use CS 2007q3. Should work out of the box, I think I tried using all of NiLuJe's switches with it and that it worked (I might have mixed it up - I tried so many things).

If you want newer version, without SYSROOT, you'll need to apply suggested changes. (Although, I only tested it with 2012.03)
(Or NiLuJe's Makefile, of course.. but don't forget to add -D_GNU_SOURCE)

Oh, and don't try 2008q1... that version just sucks :)

NiLuJe:
@kai771 : Ah, that explains the SYSROOT stuff, thanks! ;). (I'm obviously not using it ^^).

kai771:
@NiLuJe Neither am I (thanks to you :))

tigran123:
I just realized that in our specific case the CFLAGS and LDFLAGS on kpdfview link line won't do anything anyway, so I am going to get rid of teh third commit in my pull request, i.e. undo it.

NiLuJe:
@tigran123 : Err, yes, it does something. It honors LDFLAGS, at least. (ie. as-needed/O1).

tigran123:
@NiLuJe According to ld(1) manpage that flag is only meaningful for shared .so libraries, not executable binaries.

tigran123:
@NiLuJe Ah, actually it was only -Wl,-O1 that was meaningful for shared libraries only, but --as-needed could affect our case too, I agreed. Well, let's consider this after the release too.

NiLuJe:
Keeping in mind that I see this kind of thing through a Gentoo lense, the point is actually simply to honor LDFLAGS, whatever it might be set to ;-). 

Granted, in this specific case, with the default LDFLAGS set, it might not be terribly useful, but honoring this variable is still the 'right' thing to do ;-).

NiLuJe:
Just to mention that #348 should have fixed most of this, albeit with O2 instead of O3, and without _GNU_SOURCE, hopefully without breaking anything for the dreadful TC some of you are using ;).

EDIT: In particular, it's relying on the env for CXXFLAGS with libdjvu. That works for me (without -fno-use-cxa-atexit in my env), but please shout at me if it doesn't for you, @kai771 ;).

houqp:
Um... @kai771 , could you please merge the latest master branch into your local repo and solve the conflicts manually? Then we can merge this request :)

tigran123:
... after it has been tested not to break anything, of course :)

kai771:
@houqp
Um... I don't think so. This was supposed to be a simple bug fix. Nothing more. I deliberately kept changes to a minimum, only adding flags that I thought are harmless and that would allow it to be compiled out of the box with CS 2012.03 (namely, -D_GNU_SOURCE and -fno-use-cxa-atexit). For that purpose, to keep it as simple as possible, I left out compiling muPDF in release mode, and I left out other flags too.

If someone cares to read the discussions on MR, he will see that adding ARM_CFLAGS to CXXFLAGS, along with passing CFLAGS in the link stage was added in the end, while I tried to make it work with Linaro. I left it in the commit too, although it wasn't necessary for CS 2012.03, because I believe that correct passing of variables is the right thing to do. To me, it's a bug not to do so. Why would passing -O3 break anything? Or passing -march=armv6? Without it, it would be compiled with TC default, which god knows what they might be.

However, nobody was interested in fixing a bug. Wild goose chase of performance was more important. And yet, kpdfv was released with degraded performace. And when I reported a problem, I got preached that I should convert my pdfs to djvu in a very snobbish manner. So excuse me for not being in a very cooperative mood.

To prevent a reply preaching me how djvu is superior to pdf for scanned pages - I know that. But I don't want to convert my files. I want them to just work. There are much more computers with pdf reader installed compared to those with djvu viewer.

So please resolve conflicts (or whatever you want) yourselves. I don't want to waste any more time on this.

@NiLuJe
CS 2012.03 needs -fno-use-cxa-atexit. Otherwise, you get CXXABI_ARM_1.3.3 runtime error.

tigran123:
@kai771 Hey, calm down, my friend :) Nobody is preaching anything to you. At the moment KPV build is totally broken anyway, let's fix it first and then we'll see if there is still something to be extracted from your patch (I thought I did extract some useful bits already and THANKED you for it).

tigran123:
@kai771 Instead of complaining at your fellows' gracious attempts to educate you, please educate us instead --- where do we get these recent "CS/MG 200???" TC from and how do we install them? Do we need to copy anything from the Kindle or do we just extract some tar file to /usr/local (like with x-tools) and everything will just work?

Education is a good thing --- I always taught all my colleagues the million things which they didn't know and learnt ten million things from them as well. Static knowledge is useless (and ultimately suicidal because it tends to give one the self-conceit of superiority), only dynamic, growing knowledge has any value and is conducive of the spirit of brotherhood and cooperation.

houqp:
@kai771 OK, I see :)

The reason I did not solve the conflict by myself is I am not experienced as you do regarding the cross compiling stuff, so I was afraid that my fix might break something else.

kai771:
@houqp 
I'm not experienced at all. I literally ment when I said that I'm just a noob. I was not being sarcastic. In fact, before this [url: http://www.mobileread.com/forums/showpost.php?p=2225087&postcount=629] I never tried cross compiling in my life.

Instructions on setting up CS environment are in my guide here [url: http://www.mobileread.com/forums/showpost.php?p=2227307&postcount=658], in "Mainstream way" section. It's the same for newest 2012.03, but you'll need the changes I commited to work. So, it will work out of the box with my mainfix branch. I don't know what has been done with the master after it, and I probably wouldn't be able to help anyway.

tigran123:
@kai771 So I don't need to copy any libraries from Kindle? Perfect. I'll give it a try and see if it fixes the mupdf miscompilation (see #359). Thank you.

(after 20 minutes --- still waiting for their email... nothing yet. Maybe they only provide it for people in the USA but not in the UK? I emailed them directly asking to check or re-send the download URL to me manually).

kai771:
@tigran123 No, I'm not in the USA either... anyway, you should have those binaries hosted on DIY KeK pages too.

tigran123:
I got it eventually, just now. I'll try to compile with it and will report if mupdf lockup is fixed by this TC. So, the current recommendation is definitely 2007-q3-51, right? Or 2012.03?

kai771:
@tigran123 2007q3 worked out of the box without changes. 2012.03 needs changes. Depends what you wanna do. Both will work with my makefix branch, though.

tigran123:
Ok, thank you, I will try both.

tigran123:
@NiLuJe  Btw, why did you put popen_noshell/\* files (e.g. its Makefile) in .gitignore? I thought these are "our" files, i.e. we can modify them if needed (like you did) and so git should not ignore them, no?

NiLuJe:
@tigran123 : Because, while I initially put them in the repo, I ended up fetching them as part of fetchthirdparty, so I figured I should ignore anything that doesn't come straight from _our_ repository (right now, that means only the single patch), and not from outside sources (ie. fetchthirdparty). It was cluttering the git status output for no good reason.

I'm open to suggestions if that's not how it's usually done ;).

tigran123:
No, no, that's fine. It is the right thing to do. I assumed that Makefile did NOT come from their repository but that you specifically wrote it for our usage. But now I see it is mentioned in the -buildfix.patch. Well, the patch shows that you did actually write it from scratch for KPV. But since you include it in the patch (and are sending it upstream) it logically belongs to (the future state of) upstream, so you did the right thing to mention it in .gitignore. All is fine :)

tigran123:
@kai771 I am closing this PR because it is superseded by #363. But don't get me wrong, @kai771 --- your contribution was definitely valuable and was noted in my comments. Without it I would not have moved to the latest 2012.03 and would remain "out of sync" with @NiLuJe as we would be using different build environments and things would build fine for one but not the other. So, I thank you again!

kai771:
@tigran 123
Why thank me? I didn't do a thing. This PR was outdated the moment you started fiddling with optimization settings, so of course it was much easier to replace it with newer one. Nothing wrong about that. I wasn't dissatisfied that you didn't use my commit - I was dissatisfied with your unwillingness to address the bugs. Maybe you won't agree, but I think it should've been bugs first, optimizations later. 

To tell you the truth, I would much rateher someone else did something about it (yes, I'm disappointed that none of you seem to read your own thread on MR. It seems that you just occasionally stop by, and read just a few last posts), but since no one did, I took the trouble of learning how to do it (I never used github before, this was my first).

All the work was done by NiLuJe (who do you think taught me how to do it?), so all the credits should go to him and him only.

@NiLuJe
Thanks again for all your help and your patience teaching me stuff. I really appreciate it.

tigran123:
@kai771 I am truly sorry to have disappointed you. The truth is that we both have exactly the same opinion wrt "bug first, optimizations later" and I thought that I have demonstrated it by quickly (within minutes of your report) discovering the cause of the bug you reported (slow rendering in two column mode for scanned files) and resolving it within hours or so, proving with detailed timing measurements and code analysis that indeed pre-caching in two-column mode is beneficial, at least for the scanned files and explaining how page pre-decoding is responsible for it in DjVu case (and probably something similar is there in the PDF case). I don't understand what more did you expect or how we failed to live up to your expectations, but I hope that when you consider it both in your mind and in your heart you will then find no more cause for grievance.

PR #344, opened 2012-09-30
Restrict the values of rcountmax to the range 0 to 10.

tigran123:
Currently the user can press `Shift-R` and enter a value like -10 which is accepted and saved in the settings files. Not anymore.

(Oh, and forgive the "Merge remote-tracking branch" entries in the commit list. I do remember the --rebase advice, but I always follow good advices as soon as I understand them, i.e. I have not yet understood the meaning of --rebase switch :)


tigran123:
please don't merge it just yet, I'll simplify it to make it more readable in a minute.

tigran123:
Ok, now the fix is polished into perfection, I hope :)

PR #345, opened 2012-10-01
Build optimizations.

tigran123:
1. Thanks to NiLuJe who pointed out that we are building mupdf in debug
   mode. Switching to "release" build reduced the size of the kpdfview
   binary and did not cause any performance degradation (but no noticeable
   improvement either --- the page handling times seem to be exactly the
   same, i.e. fluctuating a couple of ms in both directions).
2. Enabled ARM-specific optimization -march=armv6j -mtune=arm1136jf-s -mfpu=vfp (also in CXXFLAGS for crengine) as per @NiLuJe 's recommendation.
3. Enabled LTO for kpdfview and crengine as per @NiLuJe 's recommendation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ HOSTCXX:=g++
 CFLAGS:=-O3 $(SYSROOT)
 CXXFLAGS:=-O3 $(SYSROOT)
 LDFLAGS:= $(SYSROOT)
-ARM_CFLAGS:=-march=armv6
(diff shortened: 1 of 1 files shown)


tigran123:
Ok, that is it for now. The rest of @NiLuJe and @kai771 suggestions concern making our build useable with other (newer) versions of toolchain and can go in _after_ the release, they are not critical for the current stable release.

Issue #346, opened 2012-10-01
Pressing fiveway in Shift-S mode makes A mode unattainable.


tigran123:
There seems to be a side-effect of the recent fixes to zoom mode handling. Namely, now if open a fresh djvu or pdf book (no history) and go through its pages (by default in A mode, i.e. fit to page) everything is fine. Now, if you switch to Shift-S (fit content width) and turn a few pages and then press A to switch back to "fit to page" again, everything is fine, the mode stays "A".

However, if during browsing in Shift-S mode you touched a fiveway down then you can never go back to A mode, i.e. every page turn will reset the mode to Shift-S. Even closing the document AND deleting the history file does not solve it --- for as long as the viewer is running, opening this OR ANY OTHER book will keep switching to Shift-S mode on each page turn, ignoring your request A to switch to fit page mode. Clearly a bug...

tigran123:
Correction: delete "or pdf" in the above description, i.e. this is observable only with djvu files, not pdf.

dpavlin:
When I was implementing it I wondered if I need to invalidate last_globalzoom_mode and It seems that I do need it. It's invalidated on each call to setglobalzoom_mode which are all user-visible shortcuts.

tigran123:
pull request with the fix tested and merged into master.

PR #347, opened 2012-10-01
reset last_globalzoom_mode on setglobalzoom_mode #346

dpavlin:


tigran123:
Tested, works fine, merging now, thank you.

PR #348, opened 2012-10-02
Fix input.closeAll & don't leave zombie lipc-wait-event processes

NiLuJe:
I was looking at kpdf's handling of lipc-wait-event, because I'm doing the same thing with the SS/Fonts hack, and I noticed that it left the process hanging on exit, and left the wrapper script to handle the mess with an ugly killall.

In the course of trying to figure out why that happened, I noticed that input.closeAll/closeInputDevices wasn't ever reaching the block of code handling the termination of the slider watcher fork, because of a simple typo ;).

That still left zombie lipc-wait-event process, though... To fix it, I opted to use a custom popen() implementation (since we have no way of knowing the PID of the process popen creates with the standard API). After a bit of testing, I ended up with this.
It seems to be working properly (the first event _sometimes_ doesn't seem to print the 'Going into screensaver' infobox, might be a buffering issue somewhere, but then I'm not sure it worked better before my mess ;p), it doesn't relaunch lipc-wait-event every event, and it properly kills lipc-wait-event on exit. It might need some more massaging to handle every error paths properly, though (might still end up with zombie processes after a crash, for example).


NiLuJe:
Well, technically it was probably reached, but since it was closing stdout & stderr instead of the correct fds, I couldn't tell ;p.

I'll fix the Makefiles to build popen-noshell as a library tomorrow...

tigran123:
@NiLuJe this looks great and the idea of using `popen_noshell` is brilliant, but do you mind if we merge it right _after_ the stable release (now long overdue), so that we test and understand everything thoroughly (along with other great changes from @NuPogodi ), ok?

NiLuJe:
Not at all, actually, I'd prefer it: I still have a few things to check/do to make this cleaner ;-).

NiLuJe:
Okay, I finally figured out what confused me about signal propagation/process groups, so I'm pretty sure this should do the trick in most, if not all, cases.

Also, probably fixed the buffering issues (can't replicate anymore, at least).

Also fixed the last few Makefile tweaks to be more accurate, and harmless for upstream, AFAICT. (I killed the SYSROOT stuff since it shouldn't be needed anymore, I can add a comment about the potential need to define _GNU_SOURCE on some TCs, though).

Fetch popen_noshell during fetchthirdparty instead of bundling it in the repo (I had to add a few more safety checks besides patch -N, because patch was being stupid).

NiLuJe:
Also switched from make to $(MAKE), to be able to use make's jobserver (parallel builds).

houqp:
Looks awesome! Is it ready for merge?

houqp:
BTW, @NiLuJe do you plan to send your popen_noshell patch upstream?

NiLuJe:
@houqp : Yup, I think I'm done ;). And yeah, I'll probably send at least the buildfix to upstream, it's really, really trivial anyway ;).

NiLuJe:
FYI, Linux build issue reported upstream: http://code.google.com/p/popen-noshell/issues/detail?id=3

PR #352, opened 2012-10-02
New InfoMessage Concept

ghost:
I hope it should not result in conflicts with main-branch
Committed:
1. Part I of new InfoMessage-concept changes (still without updated unireader, crereader, filesearcher & fileinfo).
2. kpdf.sh is moved to ./kpdfviewer/ (the paths in launchpad- & kite-scripts are modified)


houqp:
Um... this is a huge patch and difficult to review on web page. Let's just merge it and test on PC?

PR #353, opened 2012-10-03
set full screen refresh on every page turn

houqp:
And two more cherry picks from new_ui_code branch


tigran123:
I'll merge it now as without this fix it won't build for emulator at the moment so nothing can be tested anyway...

tigran123:
Oops, my fault as well --- I should have reviewed more carefully before merging :)

PR #355, opened 2012-10-03
Mention svn as a dependency.

tigran123:
When we run make fetchthirdparty it does svn co of popen-noshell


tigran123:
Obvious change, so I am merging it myself.

PR #356, opened 2012-10-03
Various cleanups.

tigran123:
I'll put into this pull request various cleanups, so please don't merge until I am finished:
1. When running on the emulator don't attempt to perform any of the TTS
   functions as this will crash the application. Note that the crashes will occur when merely pressing "I" or just exiting the program and NOT if the user attempts to do any of TTS (otherwise it would not be a bug as one would just say "don't do that in emulator") --- but one may need to change message settings AND definitely needs to exit application in the emulator too :)
2. In UniReader:goto() no need to call self.doc:getPages() twice, just save the value in the local variable.

### DIFF
--- a/dialog.lua
+++ b/dialog.lua
@@ -215,6 +215,9 @@ function InfoMessage:incrTTSspeed(direction) -- either +1 or -1
 end
 
 function InfoMessage:getTTSspeed()
+	if util.isEmulated() == 1 then
+		return 0
+	end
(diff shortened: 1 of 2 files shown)


tigran123:
Ok, I am finished, this can be merged --- no more cleanups, there are more serious issues (opening some PDF files crashes if TTS is enabled and some other strange bugs, I'll try to reproduce and track them down).

PR #362, opened 2012-10-03
New InfoMessage Concept (Part II)

ghost:
1. filehistory.lua: fixed search with empty pattern ("")
2. fileinfo.lua: i) outdated messages are replaced by
   MessageInfo:inform(); ii) some cosmetical changes to hotkeys (e.g.
   {KEY_SPACE} -> KEY_SPACE)
3. filesearcher.lua:

i) the first item in the found hits is now '..' that allows user to exit
ii) initialize self.commands only once
iii) corrections to hotkeys: added fast jumps to item at position 0%,
10%, ... 100% by Alt+Q, W, ... P; added fast jump to page entered via
inputbox (by G); some corrections to function descriptions and to
hotkeys for the helppage (e.g. "goto next item" --> "next item" and
"Back" --> "Back, '..', Home");
iv) compressed code of function to delete document by extracting the
code into separate function deleteFoundFile(full_filename)
v) corrected number of hits for the header (as item ".." is added, the
number of hits = (self.items-1))
1. Thus, only two file -- unireader.lua & crereader.lua -- still remain not updated by new MessageInfo:inform(). Coming soon...

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -2,6 +2,7 @@ require "font"
 require "unireader"
 require "inputbox"
 require "selectmenu"
+require "dialog"
 
 CREReader = UniReader:new{
 	pos = nil,
@@ -208,7 +209,6 @@ end
 
 function CREReader:gotoTocEntry(entry)
 	self:goto(entry.xpointer, nil, "xpointer")
-	--self:redrawCurrentPage()
 end
 
 function CREReader:nextView()
@@ -242,8 +242,7 @@ function CREReader:showJumpHist()
 	end
 
 	if #menu_items == 0 then
-		showInfoMsgWithDelay(
-			"No jump history found.", 2000, 1)
+		InfoMessage:inform("No jump history found! ", 2000, 1, MSG_WARN)
 	else
 		-- if cur points to head, draw entry for current page
 		if self.jump_history.cur > #self.jump_history then
@@ -306,8 +305,7 @@ function CREReader:showBookMarks()
 			.." "..v.notes.." @ "..v.datetime)
 	end
 	if #menu_items == 0 then
-		showInfoMsgWithDelay(
-			"No bookmark found.", 2000, 1)
+		InfoMessage:inform("No bookmark found! ", 2000, 1, MSG_WARN)
 	else
 		local bkmk_menu = SelectMenu:new{
 			menu_title = "Bookmarks",
@@ -349,7 +347,7 @@ function CREReader:gotoPrevNextTocEntry(direction)
 		self:fillToc()
 	end
 	if #self.toc == 0 then
-		showInfoMsgWithDelay("No Table of Contents", 1500, 1)
+		InfoMessage:inform("No Table of Contents! ", 1500, 1, MSG_WARN)
 		return
 	end
 	-- search for current TOC-entry
@@ -455,7 +453,7 @@ function CREReader:adjustCreReaderCommands()
 	self.commands:add(KEY_K, nil, "K",
 		"rotate screen counterclockwise",
 		function(self)
-			InfoMessage:show("Rotating counterclockwise ", 0)
+			InfoMessage:inform("Rotating counterclockwise...", nil, 1, MSG_AUX)
 			local prev_xpointer = self.doc:getXPointer()
 			Screen:screenRotate("anticlockwise")
 			G_width, G_height = fb:getSize()
@@ -467,7 +465,7 @@ function CREReader:adjustCreReaderCommands()
 	self.commands:add(KEY_J, nil, "J",
 		"rotate screen clockwise",
 		function(self)
-			InfoMessage:show("Rotating clockwise ", 0)
+			InfoMessage:inform("Rotating clockwise...", nil, 1, MSG_AUX)
 			local prev_xpointer = self.doc:getXPointer()
 			Screen:screenRotate("clockwise")
 			G_width, G_height = fb:getSize()
@@ -481,10 +479,10 @@ function CREReader:adjustCreReaderCommands()
 		"scroll to previous/next chapter",
 		function(self)
 			if keydef.keycode == KEY_FW_UP then
-				--InfoMessage:show("Jumping to previous chapter ", 0)
+				InfoMessage:inform("Jumping to previous chapter...", nil, 1, MSG_AUX)
 				self:gotoPrevNextTocEntry(-1)
 			else
-				--InfoMessage:show("Jumping to next chapter ", 0)
+				InfoMessage:inform("Jumping to next chapter...", nil, 1, MSG_AUX)
 				self:gotoPrevNextTocEntry(1)
 			end
 		end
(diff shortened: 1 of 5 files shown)


tigran123:
Does not KEY_BACK already exit from the search list? It does, I just checked. I see no reason for adding ".." entry. Please clarify the need for it.

tigran123:
Actually, KEY_HOME exits the search list as well. This should be fixed to only allow KEY_BACK to exit (and match the help info). Having too many unnecessary aliases is confusing because the user can't "get used" to the correct keys to be used in the given context, i.e. he tries to use all possible combinations hoping that some key will work.

tigran123:
I have just reviewed it and the rest of the changes look fine to me (in fact are very desirable features). So if you remove the ".." from the list (don't forget the check `if #self.result < 2` would become `if #self.result == 0` in KEY_S handler.) then I recommend merging your patch now. Also please remove KEY_HOME from the "break" handler while you are at it.

ghost:
> Does not KEY_BACK already exit from the search list? It does, I just checked. I see no reason for adding ".." entry. Please clarify the need for it.

It just adds another way to leave filehistory and filesearcher, besides it looks for me quite comfortable to have similar look'n'feel for whole filemanager-branch -- filechooser, filesearcher & filehistory

> Also please remove KEY_HOME from the "break" handler while you are at it.

I've added all the quitting keys for compatibility reasons. Do you have some reasons not to map quit on KEY_HOME and/or '..'?

PS. I've just prepared the last commit with the rest files -- unireader & crereader, but since this commit has the same name as New InfoMessage Concept (Part II), I can't do pull request

tigran123:
Your reason for adding ".." to the history is not valid --- the amount of space on Kindle's screen is limited and should not be wasted for duplicating existing functionality. KEY_BACK does the job nicely.

The reason for not having two (or more) quit keys I have already explained, see my comment. The "compatibility" reason is not valid either --- if something is done wrongly (in other parts of application) we should not quietly duplicate it but rather do the right thing in the new code. And the old code will be cleaned up too, eventually.

Concerning the next pull request --- ok, to help you out I suggest we merge this one and then cleanup the ".."/KEY_HOME issues later.

tigran123:
Oh, there is another bug --- the history list should have no ".." entry either --- it is a total waste of space. I don't know if it came from your commit but please remove it while you are removing the one from the search list.

And, please, please always SPOONFEED your patches. Do you know what "spoonfeeding the patches" means? It is a term invented by Linus Torvalds many years ago. It is the correct policy for free software development (i.e. where millions of eyes are scrutinizing the changes simultaneously) --- namely, each commit should have just one logically connected set of changes. Do NOT bundle so many changes into a single commit, making it impossible to analyze.

tigran123:
Also, if you simply back out of the search box in history then it will crash:

```
lua config error: ./filehistory.lua:307: bad argument #1 to 'upper' (string expected, got nil)
```

I don't know if this was introduced by your recent changes, but you may wish to fix it if you are working on that particular code. And if you do --- please don't put everything in one commit. Split it into small, logical chunks.

tigran123:
(EDIT: circular logic argument removed :)

But pressing KEY_HOME should not exit in BOTH cases. The main reason (if other reason I gave you are not sufficient) is because it is too easy to press KEY_HOME one extra time and exit the application altogether, which is very annoying.

tigran123:
@NuPogodi It would be nice to get some kind of acknowledgment, otherwise I may simply fix all these bugs myself but then my changes would conflict with yours and I don't want that --- it would be a waste of time for both of us. It doesn't cost you a penny to be sociable you know :)

Btw, yet another reason for not having ".." in the history is that the most common use thereof is to quickly get to the last book read (no, Shift-P-P doesn't work on my DXG, btw this is also a bug that needs to be investigated) and having ".." in the list adds one more unnecessary key press.

ghost:
> Also, if you simply back out of the search box in history then it will crash: lua config error: ./filehistory.lua:307: bad argument #1 to 'upper' (string expected, got nil)

AFAIK, it was already fixed https://github.com/hwhw/kindlepdfviewer/commit/25e37cf42ae74b8c27012a18ac1fa3838bddde38#diff-0 , requested to pull, but not merged

> But pressing KEY_HOME should not exit in BOTH cases. The main reason (if other reason I gave you are not sufficient) is because it is too easy to press KEY_HOME one extra time and exit the application altogether, which is very annoying.

I'm using Alt-Back to exit both any reader (unireader & crereader) & filemanager. And standalone CoolReader3, FBReader, etc If you don't use Alt-Back to quit and prefer KEY_HOME for this purpose, one may readily remove it afterwards. 

> Your reason for adding ".." to the history is not valid --- the amount of space on Kindle's screen is limited and should not be wasted for duplicating existing functionality. KEY_BACK does the job nicely.

Following your arguments, one should also delete '..' in filechooser and use Back to go one level up. Well, we have just made a release, so one has approximately one month to decide whether the one additional possibility to quit is so harmful.

tigran123:
No, the ".." in filechooser is part of FORTY YEARS of Unix history, please don't confuse the issues by mixing in irrelevant analogies. Ok, don't worry about all this, I'll fix them myself if you don't wish to.

tigran123:
And concerning the search crash fix --- you should have created a separate commit and a separate PR for this and it would have been merged easily and painlessly. But now your PR is not even mergeable anyway, because it contains conflicts. Please redo this properly. I am keeping this PR open even though it is unmergeable, in hope that you will add appropriate commits to it which address the issues brought up.

tigran123:
@NuPogodi Ok, I have just found a solution for your troubles (you don't need to thank me --- surely you would have done the same for me --- are we not all brothers?). All you need to do to be able to get proper pull requests (hereafter PRs) through is to create separate branch for each change and then send a PR from that branch into hwhw/master. This actually works, I have tested it right now, in order to assist you.

So, in summary, although one cannot pick and choose commit from the master (seems unfortunate, but there is a good reason for it) to send in a PR, one can create a branch and then send PRs from it to remote master.

Therefore, this PR can be safely closed now as unmergeable. Then this is what you should do:
1. Start from the current master and create a branch for the specific fix (e.g. for the filesearch crash OR InfoMessage:inform() transformation but NOT both!)
2. Commit the fix into that branch.
3. Push it to origin (i.e. to your own fork, NOT to upstream)
4. Open PR from the branch in your fork into hwhw/master. This should work fine. If not, please tell me and I'll gladly try to help you.

tigran123:
@NuPogodi Hey, I hope I haven't inadvertently upset you, have I? To prove my genuine willingness to help I volunteer to do the git chores for you (for this time only), passing the option `--author=NuPogodi` so all the code you wrote is properly attributed to yourself. Just let me know.

All I ask in return is your willingness to learn to spoon-feed the patches in the future. Learn from @NiLuJe @dpavlin and @houqp --- why is it so pleasant to review their patches? Because they are logically autonomous, clearly commented and SMALL. This is the right, i.e. the Unix way of doing things...

dpavlin:
smaller commits also makes it easier to revert change if we find (after release, blush!) that we broke something :-)

Issue #359, opened 2012-10-03
Can't open any PDF files


tigran123:
With the current build using x-tools we cannot open any PDF files. (this has nothing to do with TTS enabled, I just happened to have it enabled at the time, but reproducible with it disabled too. The same issue as I mentioned before "invalid page size in page 1" errors from mupdf and then lockup, so I have to hard-reset the Kindle).

So we need either:

a) clear instructions on upgrading TC to a more recent one.

b) undo the recent optimizations in the Makefile in order to work with x-tools TC.

NiLuJe:
I'll add some more explicit flags when I get back, see if it helps.

But, yes, will also provide TC instructions, because I'm pretty sure the one you're using generates binaries that won't work on a K2.

tigran123:
Ok, excellent, thank you very much! :)

tigran123:
Just for completeness let me mention the exact messages in the crash.log I get when compiling the current state of KPV with x-tools:

```
warning: cannot find page size for page 1
warning: invalid page size in page 1
```

And then total lockup: can't go to screensaver or connect via usb --- only hard reset of device will help.

These are exactly the same symptoms as what I got when blindly (apart from a typo fix) applying @NiLuJe 's makefile-mangling patch a couple of days ago. I assume that changing mupdf compilation is dangerous. I could track it down and say which changes need to be reverted, but a better solution would be to upgrade the TC to the more recent and recommended by @NiLuJe so I shall patiently wait for his instructions.

Also, if we are all building with the same TC then it would make the development much smoother.

NiLuJe:
Huh. Surprisingly enough, it does work on a K2. It shouldn't, though: 2.6.26 > 2.6.22.

tigran123:
Well, I get this when building with 2012.03:

```
./kpdfview: /lib/libm.so.6: version `GLIBC_2.15' not found (required by ./kpdfview)
```

This is without your `-mfloat-abi=softfp -marm` change. trying with it now...

Even though I added -DGNU_SOURCE (as per @kai771 recommendation) and also passing all the stuff to djuvulibre build as per one of your commits in your fork).

tigran123:
I meant `-D_GNU_SOURCE` of course. Btw, without it I would get GLIBC_2.17 error as well.

NiLuJe:
@tigran123 : I'll go get 2012.03 (and whatever the latest one might be) to check it out myself ;).

tigran123:
[...]

Issue #357, opened 2012-10-03
make cleanthirdparty broken on the emulator


tigran123:
With the recent changes to Makefile we now have to type this to do cleanthirdparty even on emulator:

```
make CHOST=arm-kindle-linux-gnueabi cleanthirdparty
```

because if we type the usual `make cleanthirdparty` then we'll get error from luajit's Makefile:

```
$ make cleanthirdparty
make -C luajit-2.0 CC="gcc" HOST_CC="gcc -m32" CROSS="arm-none-linux-gnueabi-" clean
make[1]: Entering directory `/home/tigran/Kindle/kindlepdfviewer/luajit-2.0'
make -C src clean
make[2]: arm-none-linux-gnueabi-gcc: Command not found
make[2]: Entering directory `/home/tigran/Kindle/kindlepdfviewer/luajit-2.0/src'
make[2]: arm-none-linux-gnueabi-gcc: Command not found
make[2]: arm-none-linux-gnueabi-gcc: Command not found
make[2]: arm-none-linux-gnueabi-gcc: Command not found
make[2]: arm-none-linux-gnueabi-gcc: Command not found
make[2]: arm-none-linux-gnueabi-gcc: Command not found
Makefile:251: *** Unsupported target architecture.  Stop.
make[2]: Leaving directory `/home/tigran/Kindle/kindlepdfviewer/luajit-2.0/src'
make[1]: *** [clean] Error 2
make[1]: Leaving directory `/home/tigran/Kindle/kindlepdfviewer/luajit-2.0'
make: *** [cleanthirdparty] Error 2
```

NiLuJe:
The cross TC isn't in your PATH?

I can fallback to the previous ugly fix of simply unsetting CFLAGS, but I thought this was slightly less ugly ;).

(The issue being if you happen to have CFLAGS set in your env with a -march set, all hell breaks loose).

I'll look into it, see if I can come up with something that doesn't require the cross-tc on clean ;).

tigran123:
Of course it is in my PATH. Otherwise how I would have been compiling KPV for the past three months :)
But cross TC should be _irrelevant_  for `make cleanthirdparty` in the emulator mode.

tigran123:
Btw, we need clear instructions on upgrading TC to the one that works with your recent changes, because x-tools no longer does, see issue #359 I have just opened.

NiLuJe:
Have a few errands to run, will fix when I get back.

basically,

[...]

PR #363, opened 2012-10-03
Bunch of Makefile fixes

NiLuJe:
As promised & discussed, this should fix #341, #359 and #357 ;).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,7 @@ clean:
 	rm -f *.o kpdfview slider_watcher
 
 cleanthirdparty:
-	$(MAKE) -C $(LUADIR) CC="$(HOSTCC)" HOST_CC="$(HOSTCC) -m32" CROSS="$(CHOST)-" clean
+	$(MAKE) -C $(LUADIR) CC=$(HOSTCC) CFLAGS=$(BASE_CFLAGS) clean
 	$(MAKE) -C $(MUPDFDIR) build="release" clean
 	$(MAKE) -C $(CRENGINEDIR)/thirdparty/antiword clean
 	test -d $(CRENGINEDIR)/thirdparty/chmlib && $(MAKE) -C $(CRENGINEDIR)/thirdparty/chmlib clean || echo warn: chmlib folder not found
--- a/Makefile
+++ b/Makefile
@@ -218,7 +218,7 @@ $(DJVULIBS):
 ifdef EMULATE_READER
 	cd $(DJVUDIR)/build && CC="$(HOSTCC)" CXX="$(HOSTCXX)" CFLAGS="$(HOSTCFLAGS)" CXXFLAGS="$(HOSTCFLAGS)" LDFLAGS="$(LDFLAGS)" ../configure --disable-desktopfiles --disable-shared --enable-static --disable-xmltools --disable-largefile
 else
-	cd $(DJVUDIR)/build && CC="$(CC)" CXX="$(CXX)" CFLAGS="$(CFLAGS)" CXXFLAGS="$(CXXFLAGS)" LDFLAGS="$(LDFLAGS)"  ../configure --disable-desktopfiles --disable-shared --enable-static --host=$(CHOST) --disable-xmltools --disable-largefile
+	cd $(DJVUDIR)/build && CC="$(CC)" CXX="$(CXX)" CFLAGS="$(CFLAGS)" CXXFLAGS="$(CXXFLAGS)" LDFLAGS="$(LDFLAGS)" ../configure --disable-desktopfiles --disable-shared --enable-static --host=$(CHOST) --disable-xmltools --disable-largefile
 endif
 	$(MAKE) -C $(DJVUDIR)/build
 
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,7 @@ clean:
 	rm -f *.o kpdfview slider_watcher
 
 cleanthirdparty:
-	$(MAKE) -C $(LUADIR) CC=$(HOSTCC) CFLAGS=$(BASE_CFLAGS) clean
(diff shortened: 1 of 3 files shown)


tigran123:
Cool, let me test it and get back to you.

tigran123:
Yes, perfect --- it works fine with the latest 2012.03 TC from CS/MG. However, before merging I would like confirmation of @dpavlin and @houqp as this would possibly require them to upgrade their TC as well. Are you guys ok with this? This would mean we are all "on the same page", so to speak, i.e. using the same build environment. I think there is definite benefit to be gained from it. And thanks to @NiLuJe who fearlessly fought with all the nasty alignments and make issues and won at the end! :) Thanks also to @kai771 who initiated the whole idea of moving to the latest TC!

PS. Btw, DjVu paging seems incredibly fast now... Hmmm, I should measure the timings properly before making such statements but it may be that the new way of building djvulibre did make some difference...

NiLuJe:
I actually forgot to check if it fixed the PDF issues with the old GCC 4.2/glibc 2.5 TC...

Will do, because I'm intrigued...

@tigran123: I think djvulibre forces -O3, possibly crengine too. I didn't delve into autotools/cmake to check, and since it didn't mangle the rest of the CFLAGS, I'm trusting upstream knows what it's doing ;).

tigran123:
Well, if @dpavlin and @houqp agree to move to 2012.03 TC then there is no need :)
Let's wait what they say...

But there is no harm to check of course, especially if you are intrigued.

tigran123:
Also, if we all agree to move to 2012.03 TC then @kai771 can remove the "Controversial part" of his Development Guide, which may attract more developers (hopefully) as it would be crystal clear that we are not doing anything dodgy or breaking anyone's licenses...

NiLuJe:
Hmm, it's working, but with the same kind of issues on FW 2.x as with a GCC 4.7 w/ unaligned access TC... Except it's GCC 4.2, so we can't do a thing about it. That's weird. Will do a few checks.

NiLuJe:
Yup, -mno-unaligned-access correctly 'fixes' these issues on my own Linaro TC (which is still 'broken' in the sense that it still defaults to -munaligned-access, so a good testcase for this), but I'm afraid there's not much we can do about it on the old TC, so, yeah, while it might appear to work, I'd still recommend updating ;).

NiLuJe:
Huh. The alignment trap doesn't increment/complain, but switching it to fixup instead of ignore does fix the issues. I'm stumped o_O.

tigran123:
Great, let's hope that @dpavlin and @houqp hear you. I am all for upgrading. Now (after the release, i.e. long before the next release) is the _ideal_ time for such things --- leaves us plenty of time to test everything thoroughly...

NiLuJe:
Okay, scratch my three previous messages. It's a random issue I'm seeing only on FW 2.x, every other kpdf launch, regardless of the TC used, and mostly unrelated to the unaligned accesses thing.

I'm checking with a MG build, just in case, and with my K2 in a slightly less custom state, but basically: apparently still works with the old TC.

tigran123:
With FW 3.4 I see no issues whatsoever. And djvu (both opening and paging) definitely seems a bit faster (I am just too lazy to downgrade and compare the timings accurately :)

dpavlin:
I'm all for moving to newer toolchain, so if you are happy with this changes I would love to see it merged.

tigran123:
Yes, I tested them and even provided the build for the public in my "developer's build" area.
Ok, merging now. (let's hope @houqp doesn't mind :)

tigran123:
Or shall we wait for @houqp confirmation? I would hate to upset him...

NiLuJe:
Okay! False alarm. It appears to be a bad interaction with the specific Fonts/Settings I'm using on my K2 with the Fonts Hack and/or me being cheap and running kpdfview manually instead of relying on the launcher.

Can't reproduce it anymore via the launcher, after having killed the fonts hack.

That was starting to drive me mad. :D

(FWIW, the issue was broken screensaver behaviour every other launch: the 'Going to screensaver...' popup appearing on top of the screensaver (and staying there!) and not on top of the currently viewed item, and resuming to the screensaver instead of the last viewed page/item).

And it appears to still be working with every binaries (MG/old/mine).

And, yep, the MG binary does appear to be slightly snappier.

And I need to finish a sentence without the word 'appear' in it, so, there :D.

tigran123:
You are still developing on K2??? :)
I also still like my old good Hanlin V3 (though I gave away as gifts almost all devices, keeping just one as a "memento" :), but I doubt I would ever bother developing on it again (btw, @houqp said he referred to my little libdjvu for Hanlin V3 while writing djvu support in KPV, so it wasn't a waste of time after all :)

NiLuJe:
@tigran123 : Well, I mostly try not to break anything on it when I change stuff ;). For what I'm usually doing, it's fairly similar to the K3, so that's usually not a hassle to keep supporting ;).

It was also my first Kindle/eReader, so that may come into the equation ^^.

dpavlin:
FWIW, I think it's important to support older devices because rooting kindle to install software is big enough hurdle to pass for most Kindle users, so we don't want to drop them off.

AFAIK, from software perspective K2 is on same software version as DXG so supporting it shouldn't be hard (I've seen only one K2 and nowdays it's at 3.x software, so I really can't check that claim :-)

NiLuJe:
@dpavlin : Yep, the DX/DXG uses the exact same software as the K2 ;).

houqp:
Oh, I just got up ;p Missed a long conversation. I would like to have my TC updated too and it will be better if all of us use the same one. :) 

merging for now.

tigran123:
@houqp Great, thank you!

NiLuJe:
@houqp : Well, to be perfectly honest, it was pretty much mostly me rambling to myself and running in circles because of a stupid oversight ^^.

PR #366, opened 2012-10-03
Kill getDiskSizeInfo() function and make only Back (not Home) return from Search.

tigran123:
1. Instead of making an extra function call, packing return values in key'd
   values of a table and then use two table indexing operations, just call
   util.df(".") directly from FileInfo:formatDiskSizeInfo() and use its two
   return values appropriately.
2. Make only Back (not Home) return from Search results. For more info see the commit message.

### DIFF
--- a/fileinfo.lua
+++ b/fileinfo.lua
@@ -57,17 +57,9 @@ function getUnpackedZipSize(zipfile)
 	return tonumber(res)
 end
 
-function getDiskSizeInfo()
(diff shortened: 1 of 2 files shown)


tigran123:
@houqp Ok, I understood --rebase, so from now on I'll do:

```
$ git pull --rebase upstream/master
```

Seems to work fine, i.e. the commit log after pushing to origin looks cleaner. Thank you for pointing this out :)

Issue #367, opened 2012-10-03
Minor problem with kindlepdfviewer-v2012.09-64-g6ea40b4.zip


dracodoc:
The new release is great, now the crengine cache works correctly, the opening time for epub book is fast enough after the first time.

One minor problem I found is:
When pressed menu button for status, there will be a "retrieving TOC" message, but it will not appear after the first time. No matter if there is a TOC or not, this message always pop up in first time and no further updates. It's a little bit confusing, I'm not sure what this is about.

tigran123:
@dracodoc Thank you, yes, we know about this, but it is not a bug really. When you press Menu to get the status you are going to press it again before you continue reading, right? And when you do that then the message will disappear, so nothing is obstructing your reading. Only while you are reading the status (for the first time) you can't see the text behind the "Retrieving TOC..." window.

I could easily fix this but it would mean adding an extra redraw of the page after the first press of Menu and I thought it is not necessary (i.e. it would just waste the battery).

But if you strongly believe that this is a bug, I can fix it in a minute. Think about this and let me know.

Oh btw, the reason why it is retrieving TOC (Table of Contents) is so that it can show which section the current page belongs to (at the bottom of the status). And, as you said, this will only happen for the first press of Menu --- the TOC is retrieved from the physical file only once (obviously the file cannot change :)

dracodoc:
That's all right, I don't think it is a bug, just it confused me. With the message, I expect there will be some update when the retrieving is done, but there is no change and the message is kept on, which may lead user to think kpv is still retrieving TOC and wait some time to see the change, while in fact it probably already done that.

If I remember correctly, the previous version kpv can also show the current location to TOC in status, but there is no such message. Now with this message, what's the benefit?

So I'm thinking the message should just be removed, instead of show it then redraw the page.

tigran123:
Ah, you see the reason the message must be shown on retrieving the physical TOC from disk is that for some files it can literally take _minutes_ and if there is no indication as to what is going on then the user will think that KPV is dead/locked up and will start resetting the Kindle.

On the other hand, soon you will be able to disable such messages altogether. And here is how:
1. In file manager press Alt-M and select "Expert mode"
2. In file manager press I and select "Messages"
3. Select option Q "Avoid any notifications".
4. Back to file manager, open the book and press Menu --- no "Retrieving TOC..." message is displayed.

[...]

PR #368, opened 2012-10-03
Kill "Retrieving TOC..." message and simplify code.

tigran123:
After some consideration I believe that it is best to get rid of
"Retrieving TOC..." message altogether on the following grounds:
1. The code in unireader.lua/crereader.lua that has to deal with this
   specifically can be removed.
2. The fillToc() is a data manipulation function and ought not to touch
   graphical state.
3. The files for which retrieving TOC can take a long time are
   _exceptionally_ rare (only two are known to me --- both are my own
   editions with detailed morphological tags which require massive
   hierarchical TOC for quick navigation)
4. Users are going to be confused when they press Menu for the first
   time and see "Retrieving TOC..." staying on the screen until they press
   Menu again.

This fixes issue #367 --- thanks to @dracodoc for pointing this out.

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -639,13 +639,7 @@ function CREReader:adjustCreReaderCommands()
 		"go backward in jump history",
 		function(self)
 			local prev_jump_no = 0
-			local need_refresh = false
 			if self.jump_history.cur > #self.jump_history then
-				-- addJump() will cause a "Retrieving TOC..." msg, so we'll
-				-- need to redraw the page after our own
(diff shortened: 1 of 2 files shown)


houqp:
Great, I think the right way to do is to manually pop out messagebox before calling `fillToc()` if needed. Then we can customize the message that present to the user (thought we are doing the same operation underneath). For example, when pressing `menu`, we can pop out `Retrieving current chapter titile...`, and when pressing `T`, we can pop out `Retrieving TOC` as before.

Issue #373, opened 2012-10-04
Wrong order of decoding and error checking in djvu.c?


tigran123:
In `djvu.c` we have the following code for both document and page decoding:

```
    doc->doc_ref = ddjvu_document_create_by_filename_utf8(doc->context, filename, TRUE);
    while (! ddjvu_document_decoding_done(doc->doc_ref))
        handle(L, doc->context, True);
    if (! doc->doc_ref) {
        return luaL_error(L, "cannot open DjVu file <%s>", filename);
    }
```

and

```
    page->page_ref = ddjvu_page_create_by_pageno(doc->doc_ref, pageno - 1);
    while (! ddjvu_page_decoding_done(page->page_ref))
        handle(L, doc->context, TRUE);
    if (! page->page_ref) {
        return luaL_error(L, "cannot open page #%d", pageno);
    }
```

I think that the checks `if (! doc->doc_ref)` and `if (! page->page_ref)` should be done immediately after their return by the corresponding `ddjvu_..._create_...` functions and NOT after passing their values (possibly `NULL`) to the decoding functions.

houqp:
Good catch :)  Closing for now.

PR #374, opened 2012-10-04
Display DjVu info in the status line.

tigran123:
The DjVu info shown in the status line includes:
1. Physical page dimensions.
2. Current value of gamma and (in square brackets) the value of the
display for which the page was designed.
3. Page resolution (in dpi).
4. Page type.
For the end-user probably the most useful bit is the page type as it
helps him decide which rendering mode to choose for this page (and also
explains why he can't see anything on the page --- e.g. when rendering
some COMPOUND or PHOTO pages in B&W mode). For the developer the
physical page dimensions are also interesting as they allow to estimate
the amount of time needed for page decoding and cache efficiency.

Oh, and also I changed the order of error checking and page/document decoding as mentioned in #373.

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -199,11 +199,10 @@ static int openPage(lua_State *L) {
 
 	/* djvulibre counts page starts from 0 */
 	page->page_ref = ddjvu_page_create_by_pageno(doc->doc_ref, pageno - 1);
+	if (! page->page_ref)
+		return luaL_error(L, "cannot open page #%d", pageno);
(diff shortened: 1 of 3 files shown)


eLiNK2gl:
while you're at it, could you please indent the date and the info on the battery in the upper status line (top of the page) to the right. For esthetic reasons. Looks better on KDXG.

tigran123:
@eLiNK2gl Yes, good idea.

eLiNK2gl:
@tigran123 10x

PR #381, opened 2012-10-06
Don't use kerning for monospaced fonts.

tigran123:
If you happen to have a word like "HISTORY" in the book's TOC you will know
how ugly it looks when kerning is enabled. Since we use monospaced font
(by default) for selectmenu, it makes sense to disable kerning for the
actual items but leave it on for shortcuts and "..." etc.

### DIFF
--- a/selectmenu.lua
+++ b/selectmenu.lua
@@ -332,11 +332,11 @@ function SelectMenu:choose(ypos, height)
 							own_face = cface
 						end
 						-- rendering menu items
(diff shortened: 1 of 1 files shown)


tigran123:
Btw, even WikiPedia says here:

http://en.wikipedia.org/wiki/Kerning

Non-proportional (monospaced) fonts don't use kerning, since their characters by definition always have the same spacing.

PR #383, opened 2012-10-07
Fixes to text rendering functions and their callers.

tigran123:
1. In `widget.lua:TextWidget:_render()` function the last argument (`kerning`) of renderUtf8Text() should be a boolean value.
2. In rendertext.lua the sizeUtf8Text()/renderUtf8text/renderUtf8TextWidth() functions had a flawed logic of checking whether to kern or not.

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -58,14 +58,17 @@ function BatteryLevel()
 end
 
 -- NuPogodi, 29.09.12: avoid using widgets
-function DrawTitle(text,lmargin,y,height,color,font_face)
+function DrawTitle(text, lmargin, y, height, color, font_face)
 	local r = 6	-- radius for round corners
 	color = 3	-- redefine to ignore the input for background color
+
 	fb.bb:paintRect(1, 1, G_width-2, height - r, color)
 	blitbuffer.paintBorder(fb.bb, 1, height/2, G_width-2, height/2, height/2, color, r)
+
 	local t = BatteryLevel() .. os.date(" %H:%M")
 	r = sizeUtf8Text(0, G_width, font_face, t, true).x
 	renderUtf8Text(fb.bb, G_width-r-lmargin, height-10, font_face, t, true)
+
 	r = G_width - r - 2 * lmargin - 10 -- let's leave small gap
 	if sizeUtf8Text(0, G_width, font_face, text, true).x <= r then
 		renderUtf8Text(fb.bb, lmargin, height-10, font_face, text, true)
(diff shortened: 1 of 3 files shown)


tigran123:
Before I send my next set of commits I would like to know the difference between the two functions in `blitbuffer.c`, namely `blitToBuffer()` and `addblitToBuffer()`. The thing is the way `blitFrom` (corresponding to C `blitToBuffer()` --- btw, having different names for the same function viewed in C and Lua is not always a good idea: I do understand the reason, like page:getWidth() would be getPageWidth() in C etc) is used in `renderUtf8Text()` is definitely wrong, and simply changing it to use `addblitFrom()` works fine, like this:

```
-                       if kerning and prevcharcode then
-                               local kern = face.ftface:getKerning(prevcharcode, charcode)
-                               pen_x = pen_x + kern
-                               --Debug("prev:"..string.char(prevcharcode).." curr:"..string.char(charcode).." pen_x:"..pen_x.." kern:"..kern)
-                               buffer:addblitFrom(glyph.bb, x + pen_x + glyph.l, y - glyph.t, 0, 0, glyph.bb:getWidth(), glyph.bb:getHeight())
-                       else
-                               --Debug(" curr:"..string.char(charcode))
-                               buffer:blitFrom(glyph.bb, x + pen_x + glyph.l, y - glyph.t, 0, 0, glyph.bb:getWidth(), glyph.bb:getHeight())
+                       if kerning and (prevcharcode ~= 0) then
+                               pen_x = pen_x + face.ftface:getKerning(prevcharcode, charcode)
                        end
+                       buffer:addblitFrom(glyph.bb, x + pen_x + glyph.l, y - glyph.t, 0, 0, glyph.bb:getWidth(), glyph.bb:getHeight())
```

However, I never change code without full understanding of what I am doing, so I would like to know if there is a real/legitimate need for the use of `blitFrom()` in the above (as opposed to always calling `addblitFrom()`) and if so, then I will find a way to use it correctly (the above usage (_before_ my modification) would produce the wrong background IF it was ever called, as it is, it is never called anyway so we never see the bug).

tigran123:
NB: I pushed the commit SHA1: c7b4cf7 to make this discussion more meaningful. If my understanding (and the code) is incorrect, I'll gladly revert it. All the changes have been tested both on the EMU and Kindle.

dpavlin:
They look OK to me

tigran123:
I forgot to fix the `renderUtf8TextWidth()` function as well. Will do in a minute in a separate PR.

PR #384, opened 2012-10-07
Show the program version in the Help header.

tigran123:
1. On startup read the file git-rev if it exists and initialize a new global
   G_program_version.
2. In filechooser's help page's header show the program version.
   This helps to easily identify what version of KPV the user is running without
   asking him to cat git-rev file manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,9 @@ POPENNSLIB := $(POPENNSDIR)/libpopen_noshell.a
 
 all: kpdfview
 
+VERSION?=$(shell git describe HEAD)
 kpdfview: kpdfview.o einkfb.o pdf.o blitbuffer.o drawcontext.o input.o $(POPENNSLIB) util.o ft.o lfs.o mupdfimg.o $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) djvu.o $(DJVULIBS) cre.o $(CRENGINELIBS)
+	echo $(VERSION) > git-rev
 	$(CC) \
 		$(CFLAGS) \
 		kpdfview.o \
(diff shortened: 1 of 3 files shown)


dpavlin:
We should probably generate git-rev on each make instead in customupdate so that compiled version will always show correct revision.

tigran123:
You mean to embed it in a little git-rev.c file embedded into kpdfview? That would save us a few ms reading from a file on startup. I can easily implement this, shall I?

dpavlin:
Sorry I wasn't clear enough. I though of moving line `echo $(VERSION) > $(INSTALL_DIR)/git-rev` in `Makefile` from `customupdate` target to `kpdfview`

tigran123:
Ah, so you don't think it is worth embedding it inside kpdfview binary? (it would be visible manually via `strings` as well).

dpavlin:
Back in CSV days that would be done with magic marker which would get expanded, but git doesn't have it, so creating file on make is only solution.

Creating lua file instead of plain file we have now might be better solution.

Update: or C file... I don't have clear pereference... But embedding it in C might be better...

tigran123:
I think executing an lua file (like we do for history/*.lua via dofile) would be a lot slower than just plain reading a single line from a text file.

So, I'll move the echo line in the Makefile as you suggest. This way we'll also have a valid version when running in EMU mode.

tigran123:
Initially I was enthusiastic about the C version but then realized that we would have to call it from Lua and that means all that setup with luaL_register stuff and creating an extra table (like util and we can't re-use util because overwriting util.c all the time will make git go mad :) --- all this extra overhead annuls the benefit of saving 2-3 milliseconds on reading the file from flash on startup. So, let's keep it simple --- it is just a couple of Lua lines to read that file...

tigran123:
Ok, I am done with this PR.

Issue #385, opened 2012-10-07
font size for hotkeys' labels (pdf hyperlinks) dependent on the font size of the first hyperlink


eLiNK2gl:
If your page is typeset as follows:
font size of the first row (contains a hyperlink): 20 pt
the rest of the page: 10 pt.

pressing 'L' overlays the hotkeys' labels in size 20 pt, making the text unreadable.

tigran123:
Hmmm, do you mean to say that pressing 'L' makes it unclear which shortcut belongs to which link? If so, then yes, this would be a bug. But if you merely mean that pressing 'L' makes the _other_ text unreadable (i.e. text irrelevant for selecting the shortcut) then this is expected behaviour, as you are not expected to read the main text in 'L' mode --- you are only expected to select the shortcut OR to pess `Back` and return from L-mode.

eLiNK2gl:
@tigran123 in some cases it could be not clear.
The problem would typically come up on an index page, which contains many hyperlinks close to each other.

As usual, I'm ready to provide a link.

dpavlin:
That's true. I'm calculating font size based on first link to make sure that shortcut will overwrite just link and not content around it. I wondered if better solution would be to calculate font size for each link, but decided against it because of font init overhead for each link (without benchmarking, so I might be wrong). I will probably re-visit font sizing issue, but I don't have good solution yet.

I'm wondering should we have shortcuts in varying size on page (which might look ugly) or use "most common" size on page?

tigran123:
I think the best way forward is to examine the worst case scenario, i.e. the actual file. @eLiNK2gl can you please provide this file? Thanks.

eLiNK2gl:
@tigran123 you've got an email (at MR).

tigran123:
Oh yes, you are right, it is a bug. @dpavlin I'll forward the pdf file to your email now.

tigran123:
Fix merged into master.

eLiNK2gl:
I do not know if it a bug.
So I am not opening a separate issue.
But the selection of the content (press 'X') is wrong on most of the pages in Stephens.pdf

tigran123:
Each PDF can define its own BoundingBox (it came from PostScript actually) internally, so it is not a bug. You can always override it with `Shift-X` and define your own bbox.

PR #387, opened 2012-10-07
recalculate font size for each link

dpavlin:
This fixes problems with pages full of links where first link
is so big that rest of link shotcuts overlap on page #385

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -3051,7 +3051,7 @@ function UniReader:addAllCommands()
 					if link then
 						link = shortcut_map[link]
 						if visible_links[link] ~= nil and visible_links[link].page ~= nil then
-							goto_page = links[link].page + 1
+							goto_page = visible_links[link].page + 1
 						else
 							Debug("missing link", link)
 						end
--- a/unireader.lua
+++ b/unireader.lua
@@ -2954,10 +2954,7 @@ function UniReader:addAllCommands()
 			if links == nil or next(links) == nil then
 				InfoMessage:inform("No links on this page ", 2000, 1, MSG_WARN)
 			else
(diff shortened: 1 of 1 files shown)


dpavlin:
don't merge this one yet. -2 should be after globalzoom

tigran123:
Tested --- works.

Update: tested on Zak.pdf fine, but on Stephens.pdf it is still not good if you define bbox around the text on the first Index page and then do `Shift-S` followed by `L` --- the index shortcuts are still too big (probably that is what you meant by "-2 should be after globalzoom"

tigran123:
Even with the latest patches the links on the file Stephents.pdf are too big.

tigran123:
Also, if you just position the screen in the middle of the Index page and then `Alt-Next` to zooom in a few times, then press `L` --- the shortcuts are awfully big and no way to tell which link each of them belongs to.

dpavlin:
I totally missed fact that I'm using globalzoom wrong. Since result is already in pixels I don't need to multiply it with zoom.

I would love to see some kind of hungarian notation of pixels and document points (maybe with `*_px` or `*_pt`) to prevent mistakes like this. It totally not obvious when looking in code if variables contain pixels of document points.

dpavlin:
Can we merge this?  d14d0ba is quite important fix, because all link targets are wrong if page has urls

tigran123:
The last time I checked the font sizes were still wrong (i.e. huge shortcuts obscuring the smaller ones). But if you changed something since then, then I'll retest it cleanly, but not today (it is almost midnight here) --- tomorrow morning. Or someone else can test it today, if it is urgent.

tigran123:
Ah, I just noticed that your commits were AFTER my comment, in fact a WHOLE DAY after! Sorry :)
I thought they were BEFORE!

tigran123:
Tested now --- works fine, merging...

Issue #391, opened 2012-10-08
How make kpdfv on DXG partial refresh ?


kaysin:
I try to add 'partial_refresh_count' into .reader.setting.lua .but it doesn't work..  

My setting file is following ........... Please HELP ~~

-- we can read Lua syntax here!
return {
    ["lastfile"] = "/mnt/us/documents/Prentice Hall.-.The C Programming Language(2nd Edition).pdf",
    ["version"] = 2012.05,
    ["pan_overlap_vertical"] = 30,
    ["fontmap"] = {
        ["hfont"] = "host/CJK.ttf",
        ["infont"] = "host/CJK.ttf",
        ["rifont"] = "host/CJK.ttf",
        ["hpkfont"] = "host/CJK.ttf",
        ["infofont"] = "host/CJK.ttf",
        ["ffont"] = "host/CJK.ttf",
        ["cfont"] = "host/CJK.ttf",
        ["scfont"] = "host/CJK.ttf",
        ["tfont"] = "host/CJK.ttf"
    },
    ["filemanager_expert_mode"] = 2,
    ["partial_refresh_count"] = 5
}

tigran123:
In a good application (like KPV) there is no need to edit any settings files manually.
When you open a file please press `H` to get the help and then read the help page. You will discover that the answer to your question is in the hotkey `Shift-R`. Oh, I just checked it myself and see that it doesn't actually explain `Shift-R` properly --- I will fix the help page now, thanks for bringing this up! :)

PR #392, opened 2012-10-08
Explain Shift-R function in the help better

tigran123:
1. Explain the purpose of `Shift-R` function better in the help screen.
2. Show the range of valid values when the user presses `Shift-R`.
3. Added a separate function for manual page redraw bount to `Space` key.

Thanks to @kaysin (see #391) for pointing this out.


tigran123:
Please don't merge this one yet, as I have another commit coming.

tigran123:
Ok, I am done with this PR. I added `Space` command which forces page redraw now. It is much more convenient than overloading `Shift-R` functionality (because the latter's InputBox obscures a part of the page anyway).

PR #395, opened 2012-10-08
Add dim_overlap configurable

tigran123:
On request of users (dracodoc, kai771 et al.) I have added a new command `O` which toggles the display (via dimmed box) of the page overlap areas. This setting is saved on a per-book basis, i.e. is specific to the book, which is the desired behaviour as one wants dimming on some books (math, physics, etc) but not others (fiction).


dpavlin:
I would name `dim_overlap` something like `show_overlap_enabled` which would make it obvious and remove need for comment in self structure.

tigran123:
@dpavlin Yes, good point, I'll rename it now.

tigran123:
Ok, done.

dpavlin:
thanks, much nicer diff :-)

Issue #401, opened 2012-10-09
Djvulibre stack corruption --- TC issues?


tigran123:
I am seeing some serious problems (independent of my recent DjVu optimizations) with viewing DjVu files. Consistent crashes with core files pointing to stack corruption. Not reproducible in the emulator at all, only the ARM binary is bad. I am suspecting the TC and now am building on 32bit to see if it makes a difference and also will try building with the old good arm-kindle TC (but I'll have to revert changes to Makefile to match arm-kindle TC).

Update: replaced the word "random" with "consistent" :)

tigran123:
The binary generated on 64 bit shows this stacktrace:

```
(gdb) bt
#0  0x0013d8c8 in ?? ()
Cannot access memory at address 0x0
#1  0x0013dae0 in ?? ()
Cannot access memory at address 0x0
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

The binary generated on 32 bit shows a different stracktrace:

```
(gdb) bt
#0  0x001238ac in lj_str_new ()
#1  0x00123f88 in lj_str_fromint ()
#2  0x00000064 in ?? ()
Backtrace stopped: frame did not save the PC
```

However, there is no guarantee that they both crashed in the same place (though what I was doing was exactly identical). You can reproduce the crash easily, just download this binary:

https://github.com/downloads/tigran123/libdjvu/libdjvu-test.djvu

and then goto page 15 (`G` 15) and press `Shift-X` to define the bbox. It will either crash immediately or allow you to set the top-left part and will crash on setting the bottom-right. Works fine in the emulator, crashes on Kindle 3.

tigran123:
Update, the 28th September version works fine --- that was the time just before we switched to the new TC:

https://github.com/downloads/tigran123/kindlepdfviewer/kindlepdfviewer-496fcc0.zip

I'll try to narrow down to see which is the first build that introduced these crashes.

tigran123:
Update: all the builds that I have made available work fine:

https://github.com/tigran123/kindlepdfviewer/downloads

The last build is dated 7th October and by then we have already switched to the new TC, so it doesn't look like a TC issue. We just need to check carefully what went into the master between the build `v2012.09-114-gcb25029` and now that broke DjVu.

Looks like a good idea to have regular daily builds --- helps to narrow down issues like this one.

tigran123:
I didn't realize that Kindle has gdb installed :)
[...]

PR #405, opened 2012-10-09
changed page number reading

hwhw:
this is far from complete since in fact, we read page references
which follow a more complex syntax. while specifying page numbers
is one option for them, they might also specify page names. we
currently do not yet have code for this case and we will return
"-1" for all page references we can not parse.

preliminary fix for issue #401

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -17,6 +17,7 @@
 */
 #include <math.h>
 #include <string.h>
+#include <errno.h>
 #include <libdjvu/miniexp.h>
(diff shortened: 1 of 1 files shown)


tigran123:
I'll test it now and get back to you asap.

tigran123:
Yes, confirmed that it does fix _one_ of the issues discussed in #401, namely crash when retrieving a TOC with more elaborate entries than just plain "#numbers". Merging now.

PR #411, opened 2012-10-10
Picture viewer (currently supports only grayscale JPEG files)

tigran123:
Here is a simple picture viewer module for KPV.
At the moment it is restricted to 1-component (grayscale) images only, but could be naturally extended to support the colour ones too. Zooming, panning seems to work fine.

Note that this PR also contains the change to rendertext.lua which fixes the crashes in the current master tree.

For those who wish to try it out I put the build here:

https://github.com/downloads/tigran123/kindlepdfviewer/kindlepdfviewer-v2012.09-164-ga68407a.zip


hwhw:
That's quite nice! However, shouldn't this be "jpegviewer"?

tigran123:
Thanks. I did call it that originally, but then renamed to more generic "picviewer" for the following reasons: support for other formats could be added to the same module pic.c --- no need to implement the whole Lua infrastructure for each format individually as some of the functions would be shared e.g. scaling the images. (of course they can be moved out to a separate C module and shared that way, but keeping the whole thing in pic.c is easier --- it is a "Simple Picture Viewer" with the stress on "Simple" :)

Also, it would avoid namespace clash with libjpeg functions --- they start with "jpeg_". So, I still think "picviewer" is a better, more generic/neutral name.

tigran123:
I reverted my "undo changes to rendertext.lua" commit, so this PR does not touch rendertext.lua anymore.

dpavlin:
nice work @tigran123

tigran123:
@dpavlin Thank you, I will try first to add support for colour (components=3) jpeg files and then will look at other formats (png seems to be the easy one).

houqp:
Finally we have this module :) Good job @tigran123

PR #412, opened 2012-10-10
Use the local copy of the global G_width variable.

tigran123:
Trivial cleanup: subject says enough.

### DIFF
--- a/djvureader.lua
+++ b/djvureader.lua
@@ -116,7 +116,7 @@ function DJVUReader:_drawReadingInfo()
 		math.ceil( self.cache_current_memsize / 1024 ).."/"..math.ceil( self.cache_max_memsize / 1024 ).."k, "..
 		math.ceil( self.doc:getCacheSize() / 1024 ).."/"..math.ceil( self.cache_document_size / 1024 ).."k", true)
 	local txt = os.date("%a %d %b %Y %T").." ["..BatteryLevel().."]"
-	local w = sizeUtf8Text(0, G_width, face, txt, true).x
+	local w = sizeUtf8Text(0, width, face, txt, true).x
 	renderUtf8Text(fb.bb, width - w - 10, 15+6, face, txt, true)
 	renderUtf8Text(fb.bb, 10, 15+6+22, face,
 		"Gm:"..string.format("%.1f",self.globalgamma).." ["..tostring(page_gamma).."], "..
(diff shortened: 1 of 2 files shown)


dpavlin:
Should we just remove width and use G_width everywhere (other way around)? It might be cleaner since G_width and G_height are always framebuffer size while width can be width of anything.

tigran123:
@dpavlin No, the Lua book advises against that, because access to a local variable is always faster than to global.

tigran123:
Btw, while we are at it --- do you know why crereader.lua uses `fb.bb:getWidth()` in the context where everything else uses `G_width`? Is there any difference? If not, then it should be replaced with `G_width` (or, better, a local copy thereof if more than one access is needed).

houqp:
@tigran123 , following are some of my thoughts:
- The safest way to do is use `fb.bb:getWidth()`, which makes sure that you always get the right screen width. Because screen rotation will change the screen width and you want to apply that update to UI drawing ASAP.
- But calling `fb.bb:getWidth()` everywhere is not efficient and ugly, so we introduce `G_width`, which gets updated when screen is rotated (hopefully). So when another function uses `G_width`, it is very likely `G_width` represents the right screen width. 
- You are right that we should use local var is a good practice. But in this case, `width` is indeed referring to the screen width, which should be shared with and consistent for all the UI drawing methods. So if you can make sure that screen will never be rotated during uses of local `width`, then it's OK. Otherwise, we should handle it with care.

tigran123:
@houqp Thank you for the explanation. I assume that by "this case" you mean this particular PR. Ok, let's look at, for example `djvulibre.lua:DJVUReader:_drawReadingInfo()` instance. In this specific instance the value of the global variable `G_width` (whose value is not going to change while we are in this function) is used 5 times, of which 4 times it is used by referring to the local reference saved in the name `width`. My patch simply tidied this up to include the 5th reference as well. In the light of your explanation I see nothing wrong with this, do you?

And similarly for the other part of the commit, i.e. the function `unireader.lua:UniReader:_drawReadingInfo()`. Again, it was used there 5 times of which 4 times via the local `width` and my patch tidied it up to include the 5th access.

I think my patch handled both instances with sufficient care, do you not?

houqp:
Oh, yes. I am not saying your patches introduce any bug. Just to make sure we come to the same agreement of how `G_width` should be used ;)

tigran123:
Yes, I realized that after sending my reply. Usually, I read my own emails only from the "Sent Mail" folder :)

PR #414, opened 2012-10-10
Fix for blitbuffer.c / boundary checking

hwhw:
Should fix yet another bug leading to issue #401

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -105,6 +105,63 @@ static int blitFullToBuffer(lua_State *L) {
 	return 0;
 }
 
+/**
+* check/adapt boundaries for blitting operations
+*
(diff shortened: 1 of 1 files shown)


tigran123:
Wow, cool. With this fix we would be able to keep my corrections to rendertext.lua, right? Obviously they will have to be slightly adjusted since there is a difference between blit and addblit :)

I will test it now and will get back to you asap.

tigran123:
Tested with `#define DEBUG 1` in `blitbuffer.c` and confirm that the master tree is no longer crashing, i.e. I could not break it in 10 minutes of intensive testing :)

The kerning is still broken though --- I broke it, I'll fix it.

tigran123:
Actually, why did I say that kerning is broken? It seems to work in the current state of master tree, does it not?

PR #416, opened 2012-10-10
Implements a toggle that allows enabling/disabling links underlining

kai771:
Implements a toggle that allows enabling/disabling links underlining

I commented earlier that new versions of KPV are sluggish with turning pages in cbz files compared to earlier versions. After some analysis,  I concluded that the culprit is this line:

local links = self:getPageLinks( no )

which was confirmed with a test replacing that line with:

local links = nil

It seems that getPageLinks() takes some time to return for cbz files, which of course, don't have links in them. It is called 
from UniReader:show(), so because of that, even panning is sluggish. This PR implements a toggle that shows/hides the links on the page (skipping calling getPageLinks if show_links_enable is false).

Btw, it seems that saving settings is broken in current dev builds.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -67,6 +67,7 @@ UniReader = {
 	pan_overlap_vertical = 30,
 	show_overlap = 0,
 	show_overlap_enable = true,
+	show_links_enable = true,
 
 	-- the document:
 	doc = nil,
@@ -987,6 +988,7 @@ function UniReader:loadSettings(filename)
 
 		self.rcountmax = self.settings:readSetting("rcountmax") or self.rcountmax
 		self.show_overlap_enable = self.settings:readSetting("show_overlap_enable")
+		self.show_links_enable = self.settings:readSetting("show_links_enable")
 
 		-- other parameters are reader-specific --> @TODO: move to proper place, like loadSpecialSettings()
 		-- since DJVUReader still has no loadSpecialSettings(), just a quick solution is
@@ -1410,7 +1412,10 @@ function UniReader:show(no)
 	end
 
 	-- draw links on page
-	local links = self:getPageLinks( no )
+	local links = nil
+	if self.show_links_enable then
+		links = self:getPageLinks( no )
+	end
 	if links ~= nil then
 		for i, link in ipairs(links) do
 			if link.page then -- skip non-page links
@@ -2966,6 +2971,19 @@ function UniReader:addAllCommands()
 			else
 				unireader:goto(unireader.pageno)
 			end
+		end
+	)
+	self.commands:add(KEY_L, MOD_SHIFT, "L",
+		"show/hide links on page",
+		function(unireader)
+			unireader.show_links_enable = not unireader.show_links_enable
+			if unireader.show_links_enable then
+				InfoMessage:inform("Links on page ON", nil, 1, MSG_AUX)
+			else
+				InfoMessage:inform("Links on page OFF", nil, 1, MSG_AUX)
+			end
+			self.settings:saveSetting("show_links_enable", unireader.show_links_enable)
+			self:redrawCurrentPage()
 		end
 	)
 	self.commands:add(KEY_L, nil, "L",
--- a/unireader.lua
+++ b/unireader.lua
@@ -2987,7 +2987,7 @@ function UniReader:addAllCommands()
 		end
 	)
 	self.commands:add(KEY_L, nil, "L",
-		"page links",
+		"page links shortcut keys",
 		function(unireader)
 			local links = unireader:getPageLinks( unireader.pageno )
 			if links == nil or next(links) == nil then


tigran123:
Very good idea, thank you!

The only slight issue that the user will get "No links on this page" if the links are disabled, but I do NOT think it is a serious issue and there is no way to overcome it anyway (short of retrieving the links, but that would defeat the whole purpose :)

I will test your patch now and will merge it if everything is fine (unless someone raises objections).

kai771:
It might not be a bad idea to also rename "page links" (on shortcut L) to "page links shortcut keys" in help.

tigran123:
Yes, please do that and add a commit by pushing it to the same branch --- it will automatically go into this PR. I am remotely tracking this branch from my end.

tigran123:
Hmmm, wait, something is wrong. If I disable the links with `Shift-L` and then press `L` I still get the shortcuts...

tigran123:
Ah, because the handler for KEY_L calls getPageLinks() anyway...
I suppose this is what we want, isn't it? I thought this function is only called in one place (the one controlled by your new configurable)...

kai771:
Yes, I wanted to still be able to see links on pressing L. I didn't want to change that. This only removes initial underlining.

tigran123:
I will now measure the timings to see how much time do we save (relative, of course, not absolute)

tigran123:
Do you have a cbz file I can test on?

kai771:
Take this one, for example:
http://www.mediafire.com/?15o1sg825azfwh5

rename zip to cbz.

tigran123:
Actually, even on a PDF file (on a TOC page with 77 links) your patch saves 15% off the page rendering time (I measured it just now), so this alone justifies merging it. So, don't worry about providing cbz file --- I am convinced that it is useful even for PDF.

tigran123:
Ok, will test on cbz as well, for completeness :)

tigran123:
Oh yes, on cbz files it saves 40% of page rendering time!

kai771:
It wasn't so noticeable on pdfs I tested (correct ones, not scanned pages), but it's really noticeable on cbz. Lag from pressing the button till something starts happening is really annoying :).

kai771:
Btw, what happened to saving settings? It doesn't seem to work.

tigran123:
Cool. Testing completed. Merging into master now. And thank you very much, @kai771 for this useful patch! :)

tigran123:
And I also mentioned it in the Wiki Download page in the ChangeLog as it is a useful user-visible feature.

tigran123:
Settings? Does it not? Hmmm, I thought it did work. Let me double-check again...

tigran123:
It does work. I have retested it again.

tigran123:
But I tested on the emulator, let me re-test on the real Kindle now...

kai771:
It doesn't even save last page on my kindle.

tigran123:
Yes, tested on Kindle as well, works absolutely fine.

Ah... maybe you didn't realize that this is a per-book setting, not global. I.e. it is saved in that book's specific history/book.lua file and is reset to default when you open another book. But when you open the first book again the setting will be loaded from the history file --- and it works correctly, I've just tested it.

And the last page (just saw your latest message) is loaded correctly too.

Something is wrong with your settings. I suggest you remove settings.reader.lua file and history/\* files and try again.

kai771:
Something must've been wrong earlier when I pulled master, since I have a folder named {history,screenshots} on my kindle, instead two separate dirs, history and screenshots. I'll pull it again and rebuild.

tigran123:
I also uploaded my build so you can try it if you wish (it has your feature in):

https://github.com/tigran123/kindlepdfviewer/downloads

It is called "11 Oct", i.e. I put it as tomorrow's build...

tigran123:
Hmmm, I have put the command `mkdir -p $(INSTALL_DIR)/{history,screenshots}` to our Makefile's "customupdate" target. Surely you didn't try to run "make customupdate" directly on a Kindle, did you?

I presume busybox's "shell" is not as intelligent as bash, so it may have tried creating a single directory "{history,screenshots}" instead of doing the right thing :)

kai771:
Of course I didn't do it on Kindle, lol. I did:

$ git pull upstream master
$ make clean cleanthirdparty
$ make fetchthirdparty (had to do it, because of the popen_noshell - it's been a while since I pulled)
$ make thirdparty kpdfview customupdate

all of it on VirtualBox Ubuntu. I'll do clone again if this build doesn't work. It might also be that Ubuntu uses dash for scripts... we'll see.

tigran123:
Hmm, yes that is what I do normally as well, except that I prefer:

```
$ git pull --rebase upstream master
```

instead of plain "pull". This is to keep the commit log tidy, i.e. free of "Merging remote branch blah-blah" type of entries.

Try this

```
$ mkdir {a,b}
```

Does it create two empty directories "a" and "b" or a single one "{a,b}"?

tigran123:
Yes, you are absolutely right, it is the `dash` problem, look:

```
$ mkdir {a,b}
$ l
total 8
drwxrwxr-x 2 tigran tigran 4096 Oct 10 20:59 a/
drwxrwxr-x 2 tigran tigran 4096 Oct 10 20:59 b/
$ rmdir a b
$ /bin/dash
$ mkdir {a,b}
$ ls -al
total 12
drwxrwxr-x   3 tigran tigran 4096 Oct 10 20:59 .
drwx------. 54 tigran tigran 4096 Oct 10 20:47 ..
drwxrwxr-x   2 tigran tigran 4096 Oct 10 20:59 {a,b}
```

kai771:
Yep, in script, it creates a dir {a,b}. You might want to change that in the makefile. As I understood, a lot of distros use dash for scripts, since it's supposed to be lighter and faster than bash. Interactive shell is bash, though, and it works corectly.

tigran123:
Done, see PR #418.

Interestingly, `make(1)` ignores the value of SHELL environment variable but honours its own macro called SHELL too...

PR #418, opened 2012-10-10
Force the use of bash in the Makefile.

tigran123:
On some distributions (e.g. Ubuntu) the shell used for scripts is dash, not bash and we explicitly rely on bash's features in the Makefile, so we must guarantee that we are running with bash (not dash or anything else).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ TTF_FONTS_DIR=$(MUPDFDIR)/fonts
 
 # set this to your ARM cross compiler:
 
+SHELL:=/bin/bash
 CHOST?=arm-none-linux-gnueabi
 CC:=$(CHOST)-gcc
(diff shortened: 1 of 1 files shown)


kai771:
Hm, I would rather use 2 lines than force bash, but... it's your call. It's not like bash features are absolutely necessary, imno (imno=in my noob opinion :))

tigran123:
Well, in my opinion, it is much safer to force the use of bash instead of going around and hunting down anything that might be bash-specific. E.g. we use logical or "||" --- does it work the same on other shells? Who knows. But forcing bash we know that we can safely use any feature. It is not like bash is some exotic thing that is passing away (like Korn Shell ksh --- though 20 years ago I couldn't imagine life without ksh :)

Issue #320, opened 2012-09-25
CoolReader3 vs kindlepdfviewer+crengine


tigran123:
I have tried using builtin crengine today [seriously] for the first time. On the good side, I was pleased to find that integrating external fonts (Minion Pro and Garamond) was trivial. But, alas, the bad sides were very serious:
1. No footnotes. I.e. they are not displayed where they should be (at the bottom of the page).
2. No hyphenation. This makes the page look very unprofessional.
3. Much slower opening (for the first time) of the files than in CoolReader3
4. No way to disable formatting for text files. I fiddled with css, but still couldn't disable it entirely.

Other issues were much smaller like no status bar with chapter marks, etc. For books (i.e. PDF/DjVu) I am vehemently against any kind of permanent "status bar" because the space is very precious. But ebooks are not books, i.e. they have no fixed concept of "page" and being just a stream of characters they can be resized and reflowed in any shape or manner, so a permanent status bar is quite acceptable and even desirable.

PR #423, opened 2012-10-11
crengine link support

dpavlin:
This is pull request from individual commits. This is not ready for merge.

I would welcome review of my C++ code so I'm submitting this as-is.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -390,6 +390,72 @@ static int cursorRight(lua_State *L) {
 	return 0;
 }
 
+static int getPageLinks(lua_State *L) {
+	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+
+	lua_newtable(L); // all links
+
+	int pos = doc->text_view->GetPos();
+
+	printf("## pos=%d\n", pos);
+
+	ldomXRangeList links;
+	ldomXRangeList & sel = doc->text_view->getDocument()->getSelections();
+	doc->text_view->getCurrentPageLinks( links );
+	int linkCount = links.length();
+	if ( linkCount ) {
+		sel.clear();
+		for ( int i=0; i<linkCount; i++ ) {
+			sel.add( new ldomXRange(*links[i]) );
+
+			lString16 txt = links[i]->getRangeText();
+			lString8 txt8 = UnicodeToLocal( txt );
+
+			lString16 link = links[i]->getHRef();
+			lString8 link8 = UnicodeToLocal( link );
+
+			ldomXRange currSel;
+			currSel = *links[i];
+
+			lvPoint start_pt ( currSel.getStart().toPoint() );
+			lvPoint end_pt ( currSel.getEnd().toPoint() );
+
+			printf("# link %d start %d %d end %d %d '%s' %s\n", i,
+				start_pt.x, start_pt.y, end_pt.x, end_pt.y,
+				txt8.c_str(), link8.c_str()
+			);
+
+			lua_newtable(L); // new link
+
+			lua_pushstring(L, "start_x");
+			lua_pushinteger(L, start_pt.x);
+			lua_settable(L, -3);
+			lua_pushstring(L, "start_y");
+			lua_pushinteger(L, start_pt.y);
+			lua_settable(L, -3);
+			lua_pushstring(L, "end_x");
+			lua_pushinteger(L, end_pt.x);
+			lua_settable(L, -3);
+			lua_pushstring(L, "end_y");
+			lua_pushinteger(L, end_pt.y);
+			lua_settable(L, -3);
+
+			const char * link_to = link8.c_str();
+
+			lua_pushstring(L, "section");
+			lua_pushstring(L, link_to);
+			lua_settable(L, -3);
+
+			lua_rawseti(L, -2, i + 1);
+
+		}
+		doc->text_view->updateSelections();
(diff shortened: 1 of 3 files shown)


tigran123:
I see that footnotes are also treated --- excellent! This was one of the major items on my list of crengine's useability mentioned in #320.

However, when I tested it I see that I cannot go back via `Back` if I follow a shortcut to the footnote in an fb2 file. Pressing `Back` just gives "Already first jump" message, so looks like it is not added to the jump history list.

Oh, and a more serious bug --- closing an fb2 book, exiting KPV and then starting it again and re-opening the same fb2 book via History caused the kpdfview address space to grow to 8.5GB with RSS of 6.3GB forcing my machine (with only 8GB physical RAM) to swap heavily, so I had to login via ssh from the netbook and kill -9 the kpdfview process. The book is Herodotus' History (in Russian) which has a lot of footnotes, here it is (I can't remember where I got it from, probably lib.rus.ec):

https://github.com/downloads/tigran123/libdjvu/Gerodot-Istorija.fb2.zip

dpavlin:
Footnotes are really not my improvement - if they are links in original document crengine will support it.

I added origin point to history, but can't reproduce problem with fb2. I recently moved to x86_64 system so I think that our configuration are similar enough. OOH, I don't think I could fix that anyway.

I also contacted Vadim (crengine author) about margins since I wasn't able to find any hints in crengine source itself but haven't received any answer yet. I'm thinking about adding fixed number to x coordinate so we can merge this PR.

tigran123:
I'll try to reproduce the problem with address space growth.

tigran123:
Ok, I've tested the latest version and it works. I can't reproduce the "infinite growth of AS" issue (must have been cr3cache corruption). My only suggestion is to disable debug printf() for the actual committed version. Btw, margins are a separate issue from links, aren't they? So, why don't we merge this PR now, so that it doesn't get out of sync too far from the master tree?

tigran123:
Actually, another suggestion would be to delete the `Shift-L` command in crereader.lua as it is meaningless there (the links are shown internally by crengine anyway). And also set the show_links_enable = false by default (via new :setDefaults() method).

tigran123:
Oops, following a link on a chm file coredumps:

```
(gdb) bt
#0  0x0000000000668e97 in ldomXPointer::getRect(lvRect&) const ()
Cannot access memory at address 0x100000008
```

Looks like memory corruption.

dpavlin:
I have a mobi tile which also crash, but my backtrace is different:

```
# command to execute: [46]L: page links shortcut keys
## pos=2310

Program received signal SIGSEGV, Segmentation fault.
0x00000000006651cd in ldomXRange::ldomXRange(ldomNode*) ()
(gdb) #0  0x00000000006651cd in ldomXRange::ldomXRange(ldomNode*) ()
#1  0x00000000006bab0d in LVDocView::getCurrentPageLinks(ldomXRangeList&)::LinkKeeper::onElement(ldomXPointerEx*) ()
#2  0x0000000000667847 in ldomXRange::forEach(ldomNodeCallback*) ()
#3  0x00000000006bf269 in LVDocView::getCurrentPageLinks(ldomXRangeList&) ()
#4  0x0000000000638b43 in getPageLinks (L=0x40000378) at cre.cpp:404
#5  0x00000000005124c3 in lj_BC_FUNCC ()
#6  0x00000000005013f0 in lua_pcall ()
#7  0x0000000000408c5e in main (argc=<optimized out>, argv=0x7fffffffe008) at kpdfview.c:76
```

dpavlin:
I will play a bit more with it today, and if I don't manage to crash it on Kindle, I will merge it into master later today

tigran123:
Ok, some of the crashes may not be directly caused by your changes but due to the fact that your changes open access to more of crengine's functionality and therefore to more of its existing bugs.

tigran123:
The previous commit (about getSelections()) seems to have completely broken links --- they don't work anymore.

tigran123:
I forgot to mention, specifically the footnotes in fb2 files no longer work, i.e. I get "No links on this page" message. Before that commit they worked fine, I've tested.

tigran123:
Yes, confirmed, after doing `git revert edd347eb555a97a9b3d2743b1cc6ac36d39f9b5b` and recompiling kpdfview the links work fine again.

dpavlin:
Yes I noticed that after I merged it to master, but #436 fixes that. However, we again have problem with segfault inside crengine.

tigran123:
Well, if the segfault happens with non-fb2 files then it is perfectly understandable. I am sure you know that crengine properly supports only fb2 format --- all the others are, mildly speaking, "experimental" and, more accurately described as "too unstable to be useable". But if it happens with fb2, then it may well be something new or due to our invalid usage or something like that.

I am very surprized that someone is using crengine to read epub files on a Kindle. It is much easier and more comfortable to send the epub to @free.kindle.com address and have it converted to mobi by Amazon. But fb2 they cannot convert so the appropriate option is to use CoolReader3 or our builtin crengine. (one can convert fb2 to mobi via calibre but this is silly because the file size will double, compared to fb2.zip --- so fb2 files should be read on Kindle natively, without conversion)

dpavlin:
I track it down to fix in PR #437

dpavlin:
For rebuild will all changes one needs to apply patch file manually or run `make cleanthirdparty && make fetchthirdparty`

Issue #420, opened 2012-10-11
Moving the last file to clipboard leaves cursor in the wrong position.


tigran123:
If you have just one file in the current directory and you pres `Shift-X` then the cursor is positioned on an empty "phantom" entry just below `..`

tigran123:
The fix for #419 will address this bug as well, hopefully...

tigran123:
Btw, this is not a cosmetic bug --- if you press `Shift-X` on that phantom position again then KPV will crash, like this:

```
lua config error: ./filechooser.lua:621: attempt to concatenate a nil value
```

tigran123:
The fix was merged to master, closing.

Issue #419, opened 2012-10-11
Very funny bug in filechooser.


tigran123:
If you have two directories, one called `00-test` and another `test` and you enter the directory `00-test` and then come out of it by pressing on `..` entry you will get positioned on the `test` directory instead of `00-test`. If you do the same with any other directory, e.g. `Books` (and there is no directory containing the string "Books" in it) then coming out via `..` positions you on `Books` of course.

tigran123:
I think it is due to the use of `string.find()` at line 175 in filechooser.lua, i.e. in the function `FileChooser:setPath(newPath)` --- we need _exact_ match, not just any match. I'll look at Lua string manipulation functions to see how to do exact matching in Lua.

Update: ... or maybe not. Anyway, I'll look at it since I already assigned myself to it :)

tigran123:
That function is actually full of bugs --- the one mentioned above is the most innocent. E.g. if you enter a directory `mupdf` it positions you to the subdirectory `pdf` because the string "mupdf" happens to match "pdf". Anyway, I'll try to fix all these bugs by re-writing it in a more readable Lua fashion.

tigran123:
The fix was merged into master, closing.

PR #424, opened 2012-10-11
Two bugfixes: issues #419 and #420

tigran123:
1. Rewritten the function `FileChooser:setPath()` to handle returning to the parent directory via ".." entry correctly. Issue #419.
2. Fixed the crash when moving the last file from a directory to the clipboard via Shift-X and then pressing Shift-X on the phantom entry that is left. Issue #420

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -152,37 +152,60 @@ function FileChooser:readDir()
 end
 
 function FileChooser:setPath(newPath)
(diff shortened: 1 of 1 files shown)


tigran123:
vfat doesn't support symlinks, so my trick with parent directory name being always shorter than the child's will always work :)

PR #425, opened 2012-10-11
Add support for colour JPEG images.

tigran123:
I used the luminance match algorithm for converting RGB colour images to grayscale as it gives the best visual results (but at the cost of speed).


tigran123:
Ok, I think I am done with commits to this one, at least for today :)

tigran123:
Oops, a tiny memory leak commit missed just one minute after you merged this PR. Ok, I'll send a new PR now...

PR #426, opened 2012-10-11
Free raw_image if it is unsupported.

tigran123:
If the number of colour components is neither 1 nor 3 we should remember to free the buffer holding the raw decoded image before returning the error to Lua.

### DIFF
--- a/pic.c
+++ b/pic.c
@@ -142,8 +142,10 @@ static int openDocument(lua_State *L) {
 			return luaL_error(L, "Cannot convert to grayscale");
 		else
 			doc->image = gray_image;
-	} else
+	} else {
+		free(raw_image);
 		return luaL_error(L, "Unsupported image format");
+	}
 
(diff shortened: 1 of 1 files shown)


tigran123:
Wait, don't merge yet. Yet another last minute fix coming...

tigran123:
In one of the error paths there is a memory leak. Not sure why, I'll debug it tomorrow. The commit in this PR is correct but we might as well merge it when I've fixed this memory leak, so please leave it until tomorrow.

tigran123:
No, false alarm --- there is no memory leak --- it was leaking only in my debug version as I simulated failure to convert to grayscale but forgot to free the gray_image (which was successfully allocated by conversion). But there was a need for the last commit which initializes doc->image in the openDocument() to guard against passing junk to free() in closeDocument() (I didn't realize that closeDocument() can be called even when openDocument() failed --- very strange).

PR #432, opened 2012-10-12
Settings etc

tigran123:
This PR contains two bugfixes, one message text cleanup and one enhancement.

1.When loading boolean settings we have to explicitly check for the return of `nil` from the `self.settings:readSetting()` function and only modify the default if the return value is NOT `nil`.
1. When closing the document we have to reset the settings to defaults, otherwise their values are re-used on the next open.
2. The messages "Links on page ON/OFF" changed to "Link underlines ON/OFF" because the old version gave two false impressions: a) as if the action is pertinent only to the current page (but in fact it is global for the entire document) and b) that we are disabling the links whereas we are only disabling the show of underlines under the links. The new message addresses both issues.
3. For the picture viewer we want to disable showing grey boxes on page overlaps and underlining of the links (because there aren't any). With the first commit in this PR we can override the UniReader's defaults very easily and elegantly.
4. Removed unneeded page redraw from `UniReader:showToc()` which was there from the days of "Retrieving TOC..." message which no longer exists.

### DIFF
--- a/djvureader.lua
+++ b/djvureader.lua
@@ -1,6 +1,8 @@
 require "unireader"
 
-DJVUReader = UniReader:new{}
+DJVUReader = UniReader:new{
+	show_links_enable = false
+}
 
 -- check DjVu magic string to validate
 function validDJVUFile(filename)
--- a/picviewer.lua
+++ b/picviewer.lua
@@ -1,6 +1,9 @@
 require "unireader"
 
-PICViewer = UniReader:new{}
+PICViewer = UniReader:new{
+	show_overlap_enable = false,
+	show_links_enable = false,
+}
 
 function PICViewer:open(filename)
 	ok, self.doc = pcall(pic.openDocument, filename)
--- a/unireader.lua
+++ b/unireader.lua
@@ -1862,8 +1862,7 @@ function UniReader:showToc()
 	end
 
 	if #self.toc == 0 then
-		InfoMessage:inform("No Table of Contents ", 1500, 1, MSG_WARN)
-		return self:redrawCurrentPage()
+		return InfoMessage:inform("No Table of Contents ", 1500, 1, MSG_WARN)
 	end
 
 	self.toc_curitem = self:findTOCpos()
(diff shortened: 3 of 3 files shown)


kai771:
I like the new "Links underline" message, but I think that you should change the text on help page too, to be consistent with this.

tigran123:
@kai771 Good point, thank you.

tigran123:
@kai771 No, it wouldn't work --- it wouldn't even run because there is no `unireader` in that context.

You have a good point that we now have two places where the default value is set, but I couldn't think of an easy way to overcome it other than to have yet another variable that actually holds the "unchangeable default value", but this is messy, unreadable and worse than having to remember to change the default value in two places. Logically, it is actually _natural_ to do this (change in two places) because we are re-using the same object between close/open of a book so we do a lot of re-initialization in the inputLoop() function for this very purpose.

tigran123:
However, you are right, there is probably still a bug lurking here, namely for the objects with the default value of `false` (like DjVu and Pic) it would be overwritten with `true` on closing the document UNLESS the new document of the same type had an explicit change of default, in which case the changed value will be loaded from the settings file.

Let's think about the best way to fix this... (but first test that the above scenario is correct, i.e. I only thought of this "in theory", but haven't yet tested that the bug actually exists :)

tigran123:
Yes, bug is there --- confirmed.

To fix it we can provide a function `UniReader:setDefaultSettings()` and allow different objects set their own defaults and reset them in inputLoop() too.

kai771:
@tigran123 I haven't tested anything either, I was just thinking about it... I just thought that it would be nice for users to be able to set their own defaults. (I, for example, would like both overlay and links to default to false on my kindle. And I don't mean it only for DjVu nad Pic - I want it to be default for all files). The way it is now (if I understood it correctly - I'm unable to test it atm), even if I set them to false, when I open the pdf file, then close it, then open another  new pdf file  - they would turn on. I would have to change it for every new pdf file I open - then it would be saved in history. I'd like to be able to avoid all that.

tigran123:
@kai771 Yes, but that is a separate issue. One thing at a time, please. I've tested my solution and it works, so I'll do a PR in a minute.

kai771:
NP, take your time. I'm not in a hurry. I was just stating my opinion.

tigran123:
Ok, done, see #434

PR #437, opened 2012-10-13
fix getCurrentPageLinks segfault on links without target

dpavlin:
It's also reported to upstream at:

https://sourceforge.net/tracker/?func=detail&aid=3577117&group_id=191284&atid=936756

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,7 +185,8 @@ fetchthirdparty:
 	# CREngine patch: change child nodes' type face
 	# @TODO replace this dirty hack  24.04 2012 (houqp)
 	cd kpvcrlib/crengine/crengine/src && \
-		patch -N -p0 < ../../../lvrend_node_type_face.patch || true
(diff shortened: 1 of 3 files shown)


houqp:
nice catch :) Seems that Vadim has stopped working on coolreader for a long time. Did he reply your mail?

tigran123:
What I didn't understand is whether this fix will be automatically picked up if we do "make fetchthirdparty"? If yes, then applying the patch would fail, right?

tigran123:
Ah, I understand now, ignore my question :)

tigran123:
@dpavlin nice catch indeed, well done!

dpavlin:
@houqp no, Vadmin didn't reply but there are recent commits at http://crengine.git.sourceforge.net/git/gitweb.cgi?p=crengine/crengine;a=summary so I'm still hopeful :-)

PR #439, opened 2012-10-14
add bookmark jumping feature

houqp:


houqp:
Oh, yes, I will fix that, thanks :)

houqp:
Oh, wait, I was too sleepy last night when committing this feature, now it turns out that `Shift + UP/DOWN` is already used by other features... so don't merge it yet.

houqp:
Key binding changed now :)

tigran123:
@houqp I think there is still a little bug there --- you forgot to pass the "true" value to "is_ignore_jmp" to goto() and as a result the jump will be added to the jump history. But surely we don't want that --- the purpose of Jump history is to keep the info about the jumps that is otherwise not available, but in this case the list of bookmarks is precisely such "jump history", so there is no need to duplicate it in the real jump history.

houqp:
Yeah, that's a good point. So let's discussion the implementation of jump history here :)

Originally, I decided that history should be a list that keep track of all the non-normal page turn, i.e. page offset that is larger than 1 or less than -1. That makes it cleaner and easier to implement.

But you are right that this implementation leads to dulplication. However, if we want to completely avoid the dulplication, we need to check every jump target with the bookmark list. For instance, when jumping to a TOC entry, we also need to go throught the bookmark list to see if the target is there. This applies to other jump targets like links.

So if we decide we need to write jump history in that way, we also need to change other places. Which implementation do you think is better?

tigran123:
Hmmm,  yes, let's think....

I think that if you are jumping via a TOC entry then the possibility that the target page also happens to be present in the bookmark list is merely "accidental", i.e. we still want such jumps (via TOC) to be added to the jump history (regardless of the target being present in the bookmark list). I.e. I don't think we really need to check such jumps against the bookmark list. And likewise for links --- the link may be pointing to a page you have bookmarked but the fact that you arrived there via link indicates that it was in the course of "random browsing", i.e. it should be recorded in the jump history.

But jumping via bookmarks list is not "random browsing" it is "purposeful use of bookmarks list as a separate autonomous jump history", so such jumps should not be recorded (in fact duplicated) in the jump history.

So, to make myself clearer, I am not focusing on the fact of duplication as such, but rather am trying to distinguish "random browsing" (via direct `G` goto page jumps or TOC or links) from the cycling through "bookmark list" which are two logically different things.

tigran123:
Sorry, when I said "jumping via bookmarks list" I meant ONLY jumping by CYCLING the bookmarks list.
But jumping via bookmarks list in the sense of selecting an entry there should probably be treated in the same way as selecting a TOC entry or a link or direct Goto via `G`.

tigran123:
I.e. cycling through bookmarks (your new feature of this PR) is like having an autonomous jump history and should not be duplicated in the jump history.

But selecting a bookmark from the list (and jumping there) is logically the same thing as selecting a TOC or a link or Goto.

houqp:
Ok, that "random browsing" explanation makes sense :) 

patch added.

tigran123:
@houqp Yesterday when I already went to bed I thought of a much better way to clarify my point, but was too lazy to get up and switch the computer on again :)

Namely, this:

When you are cycling through bookmarks (e.g. pressing `Alt-Down`) you can use the "opposite direction" key (in this case `Alt-Up`) as a `Back` key. And if the jumps are added to history then you really have TWO `Back` keys (`Alt-Up` and `Back`) --- clear case of duplication. But for all other jumps we discussed (choice from bookmarks list, choice of TOC entry, choice of link and direct `G` goto) there is no other `Back` key, so if you don't add them to the jump history then the information about the user's choice is lost --- there is no way to retrace it.

tigran123:
@houqp Thank you for adding that patch. Btw, the change to `picviewer.lua` still deletes the Shift-versions of the commands, but the real change adds the Alt-versions.

tigran123:
@houqp Also, another tiny comment: you are overwriting a global variable `bm` --- safer to make it local, imho.

tigran123:
And, btw, as a sister of your bookmark cycling feature I would like to implement a similar "TOC cycling" feature (similar to Shift-Up/Down we have in crereader, but we can't use the same keys as they are used for setting Y-panning step) bound to Alt-Left/Alt-Right keys.

So, Alt-Up/Down would cycle through bookmarks and Alt-Left/Right would cycle through TOC entries of the same level. Both would not add the jumps to the history.

eLiNK2gl:
This is a potentially very useful enhancement.
Except that I personally do not quite like the current implementation of the bookmarking feature.
Because there's  no way at the moment to recognize why I set a bookmark (as I mentioned once it would be desirable to add some kind of note to the bookmark instead of the current 'TOC stamp' + 'date stamp').

So maybe it would make sense to either cycle through bookmarks only OR pages bookmarked and pages containing highlights (excluding double counting).
One could specify in settings what one is interested in.

What do you think?

P.S: I completely agree that cycling though bookmarks should not be added to the jump history.

tigran123:
@eLiNK2gl Maybe it is better to add a new feature (but do we have keys left? I am reserving Alt-Left/Right for the TOC cycling PR #444 :) to allow cycling through the highlights list. Then you could cycle either through the bookmarks list OR highlights list. But you want to cycle through the union of these two lists... Hmmm, I am not sure it is more useful than cycling through two separate lists. But I don't feel strongly about this --- I realize that maybe the particular way you use it indeed makes it more convenient to have a single bookmarks+highlights list for cycling through...

Btw, in the bookmarks list we should definitely shorten "Page N" to "p.N" to allow three more characters to come from the TOC section name or date/time stamp.

eLiNK2gl:
@tigran123 I thought as well that we're running out of shortcuts.
That's why I suggested to use the same shortcuts but allow the user through settings to decide whther he wants to cycle through bookmarks, bookmarks & highlights or highlights only (thanks to God, the number of combinations is limited :))

tigran123:
Actually, we have a bigger problem --- Alt key doesn't work as a modifier, so neither my new TOC cycling nor @houqp 's new bookmarks cycling works at the moment... I'm trying to fix it now.

They worked fine on the emulator, of course ;)

PR #440, opened 2012-10-14
one bug fix and add paperwhite detection

houqp:


frontend/ui/device.lua
```diff
@@ -3,6 +3,35 @@ Device = {
 	charging_mode = false,
 }
 
+function Device:getModel()
+	local std_out = io.popen("grep 'MX' /proc/cpuinfo | cut -d':' -f2 | awk {'print $2'}", "r")
```
dpavlin, line 7: This will work on K3 but won't work on DXG because there is no MX in /proc/cpuinfo. We have Hardware line like:

K3:
[root@kindle root]# grep MX /proc/cpuinfo 
Hardware        : Amazon MX35 Luigi Board

DXG:
[root@kindle root]# grep Hardware /proc/cpuinfo 
Hardware        : Mario Platform

houqp, line 7: Cool, I will fix that :)

houqp:
OK, see if that works :)

PR #447, opened 2012-10-15
Increase the max value of rcountmax to 200

tigran123:
On user @kaysin 's request increasing the maximum value for the number of pages before a full refresh to 200. I see no harm in doing so and clearly it has some use as there is someone who needs this.


kaysin:
@tigran123  i see ..thanks to my friend~

PR #449, opened 2012-10-15
Fixes to bookmarks and TOC cycling features

tigran123:
1. Use local variable `bm` in the command handler instead of overwriting the global `bm`
2. Delete the correct command group in picture viewer module
3. Use `Alt-H/J` and `Alt-K/L` keys because Alt+fiveway keys don't work on a Kindle (only on the emulator).

(this relates to PR #439)


tigran123:
Although the changes in this PR are valid, we have a bigger problem --- Alt does not work as a modifier to fiveway keys on Kindle. As a result we now have two cool features that work only in the emulator :)

We'll need to select two pairs of close keys (e.g. Alt-H/Alt-J + Alt-K/Alt-L) which are currently unused and use them instead of Alt-Up/Down and Alt-Left/Right.

tigran123:
Ok, now everything tested on Kindle and it works fine, so it can be merged.

PR #452, opened 2012-10-16
add PDF reflow 

chrox:
using code from K2pdfopt

### DIFF
--- a/pdf.c
+++ b/pdf.c
@@ -20,6 +20,7 @@
 #include "blitbuffer.h"
 #include "drawcontext.h"
 #include "pdf.h"
+#include "k2pdfopt.h"
 #include <stdio.h>
 #include <math.h>
 #include <stddef.h>
@@ -511,6 +512,96 @@ static int closePage(lua_State *L) {
 	return 0;
 }
 
+static int reflowPage(lua_State *L) {
+	fz_context *ctx;
+	fz_device *dev;
+	fz_pixmap *pix;
+	fz_rect bounds,bounds2;
+	fz_matrix ctm;
+	fz_bbox bbox;
+
+	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
+	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
+
+	double dpi = 200;
+	double dpp;
+	dpp = dpi / 72.;
+	pix = NULL;
+	fz_var(pix);
+	bounds = fz_bound_page(page->doc->xref, page->page);
+	ctm = fz_scale(dpp, dpp);
+	//    ctm=fz_concat(ctm,fz_rotate(rotation));
+	bounds2 = fz_transform_rect(ctm, bounds);
+	bbox = fz_round_rect(bounds2);
+	//    ctm=fz_translate(0,-page->mediabox.y1);
+	//    ctm=fz_concat(ctm,fz_scale(dpp,-dpp));
+	//    ctm=fz_concat(ctm,fz_rotate(page->rotate));
+	//    ctm=fz_concat(ctm,fz_rotate(0));
+	//    bbox=fz_round_rect(fz_transform_rect(ctm,page->mediabox));
+	//    pix=fz_new_pixmap_with_rect(colorspace,bbox);
+	pix = fz_new_pixmap_with_bbox(page->doc->context, fz_device_gray, bbox);
+	printf("bbox:%d,%d,%d,%d\n",bbox.x0,bbox.y0,bbox.x1,bbox.y1);
+	fz_clear_pixmap_with_value(page->doc->context, pix, 0xff);
+	dev = fz_new_draw_device(page->doc->context, pix);
+#ifdef MUPDF_TRACE
+	fz_device *tdev;
+	fz_try(page->doc->context) {
+		tdev = fz_new_trace_device(page->doc->context);
+		fz_run_page(page->doc->xref, page->page, tdev, ctm, NULL);
+	}
+	fz_always(page->doc->context) {
+		fz_free_device(tdev);
+	}
+#endif
+	fz_run_page(page->doc->xref, page->page, dev, ctm, NULL);
+	fz_free_device(dev);
+
+	if(dc->gamma >= 0.0) {
+		fz_gamma_pixmap(page->doc->context, pix, dc->gamma);
+	}
+	int width, height;
+	k2pdfopt_mupdf_reflow_bmp(page->doc->context, pix, 0);
+	k2pdfopt_mupdf_rfbmp_size(&width, &height);
+
+	lua_pushnumber(L, (double)width);
+	lua_pushnumber(L, (double)height);
+
+	fz_drop_pixmap(page->doc->context, pix);
+
+	return 2;
+}
+
+static int drawReflowedPage(lua_State *L) {
+	static unsigned char *bmptr = NULL;
+
+	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
+	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
+	BlitBuffer *bb = (BlitBuffer*) luaL_checkudata(L, 3, "blitbuffer");
+ 	k2pdfopt_mupdf_rfbmp_ptr(&bmptr);
+
+	uint8_t *bbptr = (uint8_t*)bb->data;
+	uint8_t *pmptr = (uint8_t*)bmptr;
+
+	int x_offset = 0;
+	int y_offset = 0;
+
+	bbptr += bb->pitch * y_offset;
+	int x, y;
+	for(y = y_offset; y < bb->h; y++) {
+		for(x = x_offset/2; x < (bb->w/2); x++) {
+			int p = x*2 - x_offset;
+			bbptr[x] = (((pmptr[p + 1] & 0xF0) >> 4) | (pmptr[p] & 0xF0)) ^ 0xFF;
+		}
+		bbptr += bb->pitch;
+		pmptr += bb->w;
+		if (bb->w & 1) {
+			bbptr[x] = 255 - (pmptr[x*2] & 0xF0);
+		}
+	}
+
+	return 0;
+}
+
 static int drawPage(lua_State *L) {
 	fz_pixmap *pix;
 	fz_device *dev;
@@ -657,6 +748,8 @@ static const struct luaL_Reg pdfpage_meth[] = {
 	{"getPageLinks", getPageLinks},
 	{"close", closePage},
 	{"__gc", closePage},
+	{"reflow", reflowPage},
+	{"rfdraw", drawReflowedPage},
 	{"draw", drawPage},
 	{NULL, NULL}
 };
--- a/pdf.c
+++ b/pdf.c
@@ -522,8 +522,9 @@ static int reflowPage(lua_State *L) {
 
 	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
 	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
+	double zoom = (double) luaL_checknumber(L, 3);
 
-	double dpi = 200;
+	double dpi = 250*zoom;
 	double dpp;
 	dpp = dpi / 72.;
 	pix = NULL;
@@ -572,15 +573,14 @@ static int reflowPage(lua_State *L) {
 }
 
 static int drawReflowedPage(lua_State *L) {
-	static unsigned char *bmptr = NULL;
+	uint8_t *pmptr = NULL;
 
 	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
 	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
 	BlitBuffer *bb = (BlitBuffer*) luaL_checkudata(L, 3, "blitbuffer");
- 	k2pdfopt_mupdf_rfbmp_ptr(&bmptr);
 
-	uint8_t *bbptr = (uint8_t*)bb->data;
-	uint8_t *pmptr = (uint8_t*)bmptr;
+	uint8_t *bbptr = bb->data;
+	k2pdfopt_mupdf_rfbmp_ptr(&pmptr);
 
 	int x_offset = 0;
 	int y_offset = 0;
--- a/pdfreader.lua
+++ b/pdfreader.lua
@@ -1,7 +1,17 @@
 require "unireader"
 require "inputbox"
 
-PDFReader = UniReader:new{}
+PDFReader = UniReader:new{
+	-- "constants":
+	REFLOW_MODE_ON = 1,
+	REFLOW_MODE_OFF = 0,
+	
+	-- "pdf page":
+	src_page_dpi = 300,
+	
+	-- "reflow state":
+	reflow_mode_enable = true,
+}
 
 -- open a PDF file and its settings store
 function PDFReader:open(filename)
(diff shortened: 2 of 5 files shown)


houqp:
OMG, this is really a big surprise for me! I will test it now.

Just one comment before I start testing on it: I think it is better to download k2pdfopt in fetchthirdparty target instead of adding its whole source into our tree :)

pdfreader.lua
```diff
@@ -1,7 +1,17 @@
...
+PDFReader = UniReader:new{
+	-- "constants":
+	REFLOW_MODE_ON = 1,
+	REFLOW_MODE_OFF = 0,
+	
+	-- "pdf page":
+	src_page_dpi = 300,
+	
+	-- "reflow state":
+	reflow_mode_enable = true,
```
houqp: I would suggest turn reflow off by default

houqp:
OK, moving around with fiveway keys works fine so far, but there are problems with PG_FWD and PG_BCK key for every pdfs that I tried. Probably a bug in the cache system. I will take a deeper look tonight :)

Overall, it's awesome, good work @chrox !

chrox:
The original k2pdfopt source relies on modified libpng, libjpeg and libz which are incompatible with third party libraries used by kpdfview. I'm afraid it need to be patched a lot to include the whole k2pdfopt source. But I will try later.

pdfreader.lua
chrox: Most PDFs need to be reflowed to have a barely satisfactory reading experience in K3. So I think it is wise to turn reflow mode on by default.

tigran123:
@chrox Thank you, I will test it as well when I am back at home.

@houqp Actually, I had a look and it is a small function that may be easier to maintain if we consider it our own code rather than doing the "fetchthirdparty" thing and patching it heavily, so in this issue I agree with @chrox. Ah, no, I take this comment back --- I thought it was all in that small function in pdf.c, but actually it is a HUGE (6000+ lines of code) module k2pdfopt.c (which was btw imported with those strange MS-DOS ^M carriage return line-terminators added). So, yes, fetching it and patching it may be a better idea. Well, depending on how heavy is the patching...

pdfreader.lua
tigran123: @chrox No, @houqp is right. It should be turned OFF by default. 100% of PDFs which are properly made for Kindle 3 (or generally for 6" eReaders) do not require reflow. That covers all the Bibles and Epochal Revelation (as I am the one who publishes them and I normally make Kindle-optimized PDF file for each edition, not just the paper ones). And of the remaining PDF files (i.e. physics, mathematics, etc) about 90% of PDFs which are NOT specifically made for Kindle do not require it either because we have a very good manual bbox assignment function (`Shift-X`) which makes them readable, except the remaining 10% which either have a too big page size or too small font size or some other abnormality. But, for example, all the two-column PDF files do NOT require reflow, because we have a special two-column mode (`Shift-F`) and the Kindle's screen is wider than half of the width (minus margins) of even the biggest page size used for books (i.e. A4). Try folding a leaf of A4 paper in half vertically and compare to the Kindle's screen (mentally deducting the margins from paper as well). And if that is not big enough then in landscape mode it is definitely wider than the widest column ever printed.

pdfreader.lua
```diff
@@ -58,3 +68,184 @@ function PDFReader:getPageLinks(pageno)
...
+	-- #1 goal: we must render the needed area.
+	-- #2 goal: we render as much of the requested page as we can
+	-- #3 goal: we render the full page
+	-- #4 goal: we render next page, too. (TODO)
+	
+	if not self.reflow_mode_enable then
+		return UniReader.drawOrCache(self, no, preCache)
+	end
+	
+	local pg_w = G_width / ( self.doc:getPages() )
```
tigran123: Please get rid of this and all the calls to `page_indicator()` as they are meaningless in reflow mode anyway.

pdfreader.lua
```diff
@@ -1,7 +1,17 @@
...
 require "inputbox"
 
-PDFReader = UniReader:new{}
+PDFReader = UniReader:new{
+	-- "constants":
+	REFLOW_MODE_ON = 1,
+	REFLOW_MODE_OFF = 0,
+	
+	-- "pdf page":
+	src_page_dpi = 300,
```
tigran123: This seems to be unused. Instead the value of 200dpi is hardcoded in `pdf.c`

tigran123:
@chrox So far, on all the PDF files I've tried it didn't work. For example, this file reflows correctly on Hanlin V3 but not on KPV with your patch:

https://github.com/downloads/tigran123/urantia-study-edition/Reflow-test.pdf

pdfreader.lua
```diff
@@ -58,3 +68,184 @@ function PDFReader:getPageLinks(pageno)
...
+	self.commands:add(KEY_R, nil, "R",
+		"toggle reflow page",
+		function(pdfreader)
+			pdfreader.reflow_mode_enable = not pdfreader.reflow_mode_enable
+			if pdfreader.reflow_mode_enable then
+				InfoMessage:inform("Turning reflow ON", nil, 1, MSG_AUX)
+			else
+				InfoMessage:inform("Turning reflow OFF", nil, 1, MSG_AUX)
+			end
+			self.settings:saveSetting("reflow_mode_enable", pdfreader.reflow_mode_enable)
```
tigran123: You save the setting correctly, but you do not load it. So, as a result, the saved setting is unused and the flag resets to the default value.

pdfreader.lua
```diff
@@ -58,3 +68,184 @@ function PDFReader:getPageLinks(pageno)
...
+		"previous/next page",
+		function(unireader,keydef)
+			self.offset_y = 0
+			unireader:goto(
+			(keydef.keycode == KEY_PGBCK or keydef.keycode == KEY_LPGBCK)
+			and unireader:prevView() or unireader:nextView())
+		end)
+		
+	self.commands:add(KEY_R, nil, "R",
+		"toggle reflow page",
```
tigran123: Two comments on this line:
1. Please call it "toggle reflow mode" rather than "page" because: a) the mode affects the whole document, not just the current page, b) that is what Adobe calls it and c) that is what the variable itsef is called (what is good for programmers is good for users) 
2. It is unfortunate that the `R` command is at the end (even after Exit commands), but I understand the reason. The only way to overcome it would be to add it in unireader and delete (or overwrite in DjVu) in all modules other than PDF, but this is really ugly.... (the proper place for the command would be to place it before the `Shift-R` command).

tigran123:
I have tested on a dozen of other files and it still did not work, not even on one of them...
All it did was just to change the margins (cutting some text, i.e. losing information not restorable by neither panning nor paging). But no actual text reflow happened. I can provide lots of test pdf files with this behaviour, if you wish. For example, all of these:

http://world-destiny.com/pdf.html

(I should disclaim that these ones were NOT typeset by me --- from typographical standpoint they are awful :)

The only noticeable change (though far from satisfactory reflow) was on the two-column documents, but for these we do not need reflow --- they are much better read in our existing two-column mode.

So, there is still a lot of work to do on this feature, but I must congratulate you, @chrox --- very nice beginning and I am sure you will bring it to a usable state very soon! :)

Remember to test only on SINGLE-COLUMN documents, for that is where PDF reflow on KPV is desperately needed. By "desperately" I meant to include those 6-7% of PDF files where neither manual bbox (`Shift-X`) nor twocolumn mode (`Shift-F`) produce satisfactory results. These are the single-column books with extremely wide page sizes (normally used for programming books because they need to show code fragments) and small font size.

tigran123:
@chrox Here is the _ideal_ candidate to be used for testing PDF reflow:

https://github.com/downloads/tigran123/urantia-study-edition/Ideal-pdf-reflow-test.pdf

This is because this is a very typical representative of the class of books for which we need KPV reflow, because without reflow we can't really read it on KPV (even in landscape mode the type is too small).

On this file enabling PDF reflow definitely makes some text reflow changes, but with the following bugs:
1. Some text is lost at the bottom of the page and you can neither scroll/pan to it nor page-down to see it.
2. Pressing `L` in reflow mode positions the link shortcuts in the wrong place. Even without pressing `L` you can see the underlines of the links positioned wrongly (`Shift-L` to toggle them)
3. No ability to increase the text size (I don't say "to zoom in" because "zoom" is meaningless in reflow mode, we are effectively extracting the text and retypesetting it ourselves in real-time, so the more proper term would be "to increase the type size" than "to zoom").

tigran123:
Also, the text is cut sometimes on the right and sometimes on the left side (and not restorable in any way). But pressing `X` to see the actual bbox shows no problem, i.e. the text is included in the bbox on both sides.

dpavlin:
@chrox thanks a lot, this looks very interesting. I'm wondering what is user workflow when using this feature? Do we expect of users to use fiveway to scroll on current page and use prev/next page keys to change pages? Would it be better to behave like `shift+f` so that prev/next page keys would move down the page?

tigran123:
Another suggestion: looking at k2pdfopt.c code I see that it knows internally how many columns the document has (very cool!) so, it would be nice if we extracted that info and at least tried to prevent the user from using reflow on a two-column document, i.e. displayed an info message suggesting to use the proper two-column mode (`Shift-F`) instead. This would discourage the lazy users from learning how to navigate `Shift-F` mode properly (ah, what we do desperately need is the updated KPV User's Guide!!!) and used reflow mode instead (which would produce far inferior reading experience than the proper two-column mode). Alternatively, this could be described in KPV User's Guide (anyone out there willing to write it?)

@dpavlin In `Shift-F` mode it is the fiveway keys that move down the page, I assume that is what you meant. The page next/prev behave in a very strange (I always considered it "undefined" :) way in `Shift-F` mode, definitely not used for reading.

pdf.c
```diff
@@ -511,6 +512,96 @@ static int closePage(lua_State *L) {
...
+	lua_pushnumber(L, (double)width);
+	lua_pushnumber(L, (double)height);
+
+	fz_drop_pixmap(page->doc->context, pix);
+
+	return 2;
+}
+
+static int drawReflowedPage(lua_State *L) {
+	static unsigned char *bmptr = NULL;
```
tigran123: why not define `uint8_t *bmptr` and get rid of (dummy anyway) typecasting below?
Also, why did you make it `static`? In fact, it is best to get rid of it altogether and let `pmptr` be initialized by `k2pdfopt_mupdf_rfbmp_ptr()` directly...

pdf.c
```diff
@@ -511,6 +512,96 @@ static int closePage(lua_State *L) {
...
+
+	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
+	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
+	BlitBuffer *bb = (BlitBuffer*) luaL_checkudata(L, 3, "blitbuffer");
+ 	k2pdfopt_mupdf_rfbmp_ptr(&bmptr);
+
+	uint8_t *bbptr = (uint8_t*)bb->data;
```
tigran123: `bb->data` is already of the right type, no casting is needed.

pdf.c
```diff
@@ -511,6 +512,96 @@ static int closePage(lua_State *L) {
...
+	uint8_t *pmptr = (uint8_t*)bmptr;
+
+	int x_offset = 0;
+	int y_offset = 0;
+
```
tigran123, line 587: Presumably you are going to do something with `x_offset` and `y_offset` (like what we do in `djvu.c:drawPage()`) and not just keep zeros stored in them?

pdfreader.lua
```diff
@@ -58,3 +68,184 @@ function PDFReader:getPageLinks(pageno)
...
+		dc:setGamma(self.globalgamma)
+	end
+	
+	return dc
+end
+
+-- command definitions
+function PDFReader:addAllCommands()
+	UniReader.addAllCommands(self)
+	self.commands:addGroup("< >",{
```
tigran123: This is definitely the wrong way to add PDF-specific commands. The correct way (see, e.g. picviewer.lua or djvureader.lua) is to define the :init() method which calls self:addAllCommands() followed by self:adjustCommands() and also define PDFReader:adjustCommands() method which does the actual adjustments (deletion and addition). Also, you shouldn't add a command group which already exists --- you have to delete the existing one first (in this case the "< >" group). When you do this properly (via init/adjust) this will be very natural.

pdfreader.lua
chrox: @tigran123 Agreed. I'm not very familiar with kindlepdfviewer. Actually I just came to know this project in last weekends. So I will try the advanced features you mentioned above.

pdfreader.lua
chrox: @tigran123 After I remove this line reflow mode still can be restored when reopening the document. But I found that in EMULATOR mode kpdfview cannot remember last reading position. Is it a bug?

tigran123:
@chrox You are definitely doing GREAT for just a few days of familiarity with KPV, so please don't consider my comments as too harsh, just work on them gradually, one thing at a time...

pdfreader.lua
tigran123: @chrox The problem is not with the saving but with the loading of the value of `self.reflow_mode_enable`. Have a look at how `show_links_enable` is loaded in `UniReader:loadSettings()` and you can define `PDFReader:loadSpecialSettings()` which does the same for `self.reflow_mode_enable`. Remember to do the `tmp ~= nil` check or it won't work (because of boolean nature of `reflow_mode_enable`). E.g. you can't use the shorter version of the code as is done, for example, with `self.rcountmax`.

houqp: I guess it is because kpv only remembers last read page for pdf files. But for pdfreflow, each original page is typeset into multiple "virtual pages", so in order to remember the last reading position, we need to remember the offset with the last read page, so we can jump to the right "virtual page" according the offset

tigran123: ... or do what reflow in Adobe Digital Editions does (as included in Hanlin V3 firmware), i.e. jump to the top of the physical page, i.e. to the very first virtual page.

houqp:
Has anybody thought of implement the reflow in another approach? What about writing a separate reader for pdfreflow? Reasons as follows:
- the reflowed pdf is totally different from pdf file, it is in deed more like files supported by crengine. So it kind of make sense to write a standalone reader for it
- this can greatly reduce conditional codes in pdfreader.lua, because we need to treat reflow and normal mode differently in many many places, that will make our code cleaner

Or we can stick to current approach, then we will need to factor out format-unrelated code in the cache system and other places in order to reuse code as much as possible. But if we take this approach, refactoring can be ignored for now. I think it is better to make it work first, refine the code later. ;p

tigran123:
@houqp Actually, having a separate reader is a very nice idea and your reasons are valid, imho.

However, a simple question: how do you decide whether to pass a ".pdf" file to `PDFReader` or a hypothetical `PDFReflowReader`? I suppose we can make a toggle switch on filechooser.lua level, right?

hwhw:
When I thought about PDF reflow in the past, in my mind I was converting
the PDF to a stream of markup hypertext, i.e. HTML (or something like it).
I even more thought so when we got CREngine support.

Then, muPDF has a pretty decent "text" output device. It will even
introduce some "flow" information. On the other hand it's not so easy to
properly interweave this with images & Co, and tables are a matter on its
own... But that would be where I would have started.

Have a "create flow version" tool invocation in the file browser and then
hand over to CREngine. It's nice though to have something, but I'm not sure
if we should cludder the PDF output either. Reflowed PDF is an entirely
different beast than "pure" PDF.

houqp:
> However, a simple question: how do you decide whether to pass a ".pdf" file to PDFReader or a hypothetical PDFReflowReader? I suppose we can make a toggle switch on filechooser.lua level, right?

Well, we just need to read the document configuration file before opening the document to decide which reader to use :P

houqp:
> When I thought about PDF reflow in the past, in my mind I was converting the PDF to a stream of markup hypertext, i.e. HTML (or something like it). I even more thought so when we got CREngine support.

Same for me, that's why even though I know k2pdfopt project for long, but never come to the idea of using it for reflow support...

Anyway, I guess non of us will have the time and enthusiasm to implement the reflow in stream approach. ;p I am very interested in how far can we go with the k2pdfopt approach :)

tigran123:
@houqp I thought about your idea of moving reflow to a separate reader (while walking picking my daughter from school) and I agree even more now, because I see that you are wisely looking into the future rather than thinking merely of the present state. Namely, in the present state, @chrox can probably tidy everything up to everyone's satisfaction using the current approach. However, in the future (I am talking about very near future, like tomorrow or by the end of the week the latest) when one expects all the features to work properly (like highlights) we will face exactly what you are predicting, i.e. many MANY conditionals to check for reflow mode and behave differently. So I agree with your suggestion (and I see that @hwhw also agrees that reflown pdf is an entirely different beast than "pure pdf") of shifting it to a separate reader.

As for @hwhw 's idea (if I understood it rightly), yes, very nice, but like @houqp says --- who has the time and the enthusiasm to do things in the most perfect/ideal way you describe.... "A good enough" thing is always a killer of "the best" :)

houqp:
Great, let's see what @chrox think about the separate reader approach.

tigran123:
@houqp and if we are opening the file for the first time, i.e. there is no configuration history yet? chicken and egg situation :) I suppose we can open it in the standard PDF reader, which provides a dummy reflow command which advises the user to close and re-open the document? And vice versa for disabling reflow when opened in the PDFReflowReader object? Or, can we do such a mighty thing as to automatically close and re-open the document without user's intervention?

houqp:
If there is no config history then we use the default reflow mode, which I prefer to standard PDF reader. And as you mentioned, I think we can just automatically close and re-open the document to hide this fact from user ;P

tigran123:
@houqp I see @chrox is tidying up the code in the current approach. Well, how about we wait and see --- if he manages to make everything work without cluttering the code with too many `if reflow_mode` conditionals, then maybe this is "good enough"? Or do you predict that it will 100% definitely have too many such conditionals in the final "actually working" version? You know TONS more about KPV internals, so I'll just abide by what you decide.

dpavlin:
I (for one ;-) quite like the ability to turn reflow on or off in current pdf file. This is especially useful if you are reading magazine content, so you want to switch from reflow to normal page and back.

tigran123:
@chrox the latest patch makes things much better, zoom works now --- thank you! Here are the current bugs:
1. Enabling reflow positions the view to the middle of the page, should be to the top.
2. The bottom of the text is still cut off and lost (see page 8 of the "ideal example" I provided).
3. The link underlines are still in the wrong place
4. On initialization you shouldn't add "< >" command group if it already exists --- delete the existing one first
5. On initialization you are calling :addAllCommands() twice (and also the second call should be self:addAllCommands() not PDFReader:addAllCommands() btw).
6. Enabling reflow on a page with highlights (use `N` to create, `Shift-N` to see the list) moves the highlights. But disabling reflow positions them correctly again.
7. Adding highlights with reflow enabled doesn't work.

Also, I noticed that paging through the sample PDF file (the one I provided above) in the emulator requires RSS of kpdfview process to be about 10-12MB but enabling reflow makes it jump to 60MB, i.e. memory consumption increases five times... (on the ARM, being 32bit, the absolute numbers would be smaller, of course, but still I'd expect the same proportion, which is a bit worrying).

tigran123:
Btw, all those long key-by-key assignments of tables... doesn't Lua support complex assignments for tables like

A = B

implying

A.key1 = B.key1
A.key2 = B.key2

I am going to test it and see if it works :) Yes, this trivial program works:

```
local tmp = {
    global = 5,
    offset = 10
}

local t = {}
t.global = tmp.global
t.offset = tmp.offset
print("t.global=", t.global, " t.offset=", t.offset)

tmp.global = 6
tmp.offset = 11
t = tmp
print("t.global=", t.global, " t.offset=", t.offset)
$ ./test.lua
t.global=   5    t.offset=  10
t.global=   6    t.offset=  11
```

So, why do we need those key-by-key assignments in the function `PDFReader:restoreReaderStates()`? Can't we just assign like this:

```
self.last_mode_states = tmp_states
```

houqp:
I guess most of the features will need to include condition codes. Bookmarks, jump history, all the links and goto page feature will be different because of the virtual pages. And you also want to delete all the zoom mode and cropping command. Even screen rotation will be different because we need to typeset for the new screen width.

But I think we should not give too much pressure to @chrox ;p give him sometime to get familiar with KPV's code base. We can always switch to separate reader implementation without much effort if needed. It's just code copy/delete and paste anyway. Let @chrox first finish the basic content display part and clean up the thirdparty dependency. There is already lots of work waiting for him now ;)

``` lua
self.last_mode_states = tmp_states
```

Yes, I think this should work :)

dpavlin:
Should we make a new branch for this feature? I think that @chrox might benefit from our help in getting this integrated, and new branch would allow us to submit pull requests against it.

tigran123:
@houqp I agree --- and I repeat (@chrox you hear me?) that the current version is much much closer to being usable than the original, i.e. I am very impressed with how quickly @chrox got it from "practically not working with any of my files" to "working on most files with some slight glitches".

@dpavlin I thought about this yesterday too. Yes, it is a good idea, because we could all contribute (e.g. you probably know already how to position link shortcuts properly in reflow mode but it would take @chrox some time to study the code you've written in that area). So, please create a reflow branch and merge the current PR into it.

tigran123:
@hwhw I thought about the pdf reflow some more yesterday and I am afraid my current opinion is slightly different from yesterday's. Namely, it appears that both you and I had what I now call "naive view of PDF reflow", whereas actually it is a bit more subtle than that. Specifically, we both seem to have viewed PDF reflow to be limited to what the output of `pdftotext` can possibly produce, i.e. a stream of text. This text could then be taken and re-typeset using a different font than what was included in the PDF file itself (e.g. as you suggested, handed over to crengine).

However, the reflow apparently can do more than that. E.g. if you have a page with not only text but also with formulas then the text stream would substitute the formulas with EMPTY SPACE: not only because individual glyphs may have no Unicode definition (most of them do have some equivalent, as they must if the font has Unicode encoding) but more importantly because there is no Unicode standard on how to position them properly. This is why we have some markup languages (MathML) to be able to "typeset" formulas in an html-like text stream. But with PDF reflow we are getting that formula almost intact (well, in some cases it is split apart, but in many cases it is intact) --- depending on the underlying reflow engine's smartness to scan in the appropriate object as a whole into a bitmap, instead of picking apart individual glyphs and scanning them in separately.

Running the current implementation (of this PR) on a typeset PDF with lots of formulas I was very impressed with how the formulas were kept unbroken, i.e. scanned in as a whole. This is actually better than Adobe Digital Editions (at the time of Hanlin V3 and Sony PRS-505 firmware, i.e. 2009) which always broke the formulas into pieces.

I do realize that what you and I thought was _fundamental_ whereas the "tricks" of the above paragraphs are, no doubt, _heuristical_, but if the heuristics is good enough to produce good visual results then I prefer "uncertain heuristics" to "solid fundamental approach".

hwhw:
Yes, I agree. I had a look into muPDF's code base for text extraction and
it's pretty much as you said - it will only extract text. And indeed, I was
thinking about something like the output of pdftotext or poppler's
pdftohtml. Also, I just wanted to contribute a fast thought - don't
consider this as a last word ;-)

After all it's sooo great to see improvements on this item! @chrox, this is
great work!

dpavlin:
I can't really merge this pull request to new branch, so I created new branch from @chrox master.

I will also close it to prevent accidental merge to master until it's ready. Have in mind that we have to issue pull requests for this feature against reflow branch, not master.

chrox:
@tigran123 

# about bugs
1. This can be fixed by setting minimum page height to bb:getHeight() in K2pdfopt.
2. Maybe it's introducted by wrong offset_y in rfNextView when dealing with the 30 pix pan_overlap_vertical. I will test it later.
3. This is a bug in K2pdfopt. Right now we'd better not touch it.
4. Doesn't it overwrite the existing "< >" command? That is what I mean to do.
5. Not familiar with Lua syntax. I first tried self.addAllCommands() but failed. It seems that there are two ways to call functions either colon syntax or dot syntax. And It's confusing at first sight.
6. The reflowed page is a bmp image written into blitbuffer. All highlights and links information of the original page are lost after the reflowing process. 
7. There is no way to highlight text in reflow mode except the original typesetting info is restored. So we should delete 'N' command in reflow mode.

# about memory consumption

Memory consumption is largely relate to pdf page size and global zoom value. We can set an upper limit bbox size before reading the entire page.

And @houqp @tigran123 @dpavlin @hwhw thanks all for bringing to us such a wonderful reader. And It's my honor to have a change of contributing to this project.

houqp:
`foo:bar(aa, bb)` is just a syntactic sugar of `foo.bar(foo, aa, bb)`, so you can use `foo:bar()` as class method call that automatically pass `self` into it. Actually Lua does not have the concept of class buitin, everything is faked by `table`, so we are using Prototype-based OOP here. ;P

I notice that this is a good chance to test whether KPV's code base is friendly to a new developer. So @chrox , if you find anything that's does not make sense at all at first glance, please let us know so we can see if it needs to be refactored :)

dpavlin:
I started playing with reflow and noticed that it has hardcoded K3 resolution. I could add two params to `k2pdfopt_mupdf_reflow_bmp` and pass it that way, but I'm wondering if that is right approach.

Also, it seems to crash DXG (with 128Mb) as soon as I turn it on, and I assume it's out of memory. I will have to investigate that a bit more to be sure.

chrox:
@tigran123 I just misunderstood what you mean by refering to link underlines in bug #3. If the underlines are added by KPV in run time rather than typesetted in PDF document, the positional information should always been wrong. A quick fix is disabling showing link underlines in reflow mode.

chrox:
@houqp @tigran123 @dpavlin @hwhw I suggest adding another reader named ReflowReader. In file browser user is promoted to select PDFReader or ReflowReader when opening a PDF document if ReflowReader is registered with PDF file type. And with just a few lines ReflowReader can support DJVU reflow.

houqp:
Go ahead, I am quite interested in how DJVU reflow will be supported, sounds yummy :)

tigran123:
@chrox Yes, I've just said (on another thread) that DjVu reflow is a fundamentally impossible thing, except for the primitive text extraction facility that would only work on those few files which embed the OCRed (usually full of typos!) version of the document for searching.

However, I am also known to say (search the fortune cookie database on Linux, they included some of my "wise sayings" :) something along the lines of "we are used to doing impossible things every day" (or something like that, I forgot already :)

So, I am very curious to see how you do the impossible... Please go ahead. Remember to submit your PRs against the "reflow" branch, not "master".

Also, concerning overwriting the existing commands (and command groups) by simply adding a new one with the same name --- it may be that the current implementation in KPV actually does this, but I would say it is dangerous to rely on it (and first you would have to check to make 100% sure that it really does) because in the future the adding function may have side-effects, so adding the same command twice (without deleting the old one explicitly) may lead to subtle bugs. So, I suggest being tidy wrt adding/deleting commands and explicitly deleting the existing one if you need to add one of the same name (or group).

tigran123:
Here it is:

http://fortune.smithsforge.com/fortunes/6305.html

"Existence of programs that do the impossible is
 not a proof that that "impossible" is now possible."

```
- Tigran Aivazian
```

chrox:
@houqp @tigran123 If you guys look into the K2pdfopt source, you will find it rather straightforward to add DJVU support. K2pdfopt processes both PDF and DJVU files in the same steps as follows: 

i. read one page of PDF/DJVU file into a master bmp image
ii. break the master bmp image into several image regions
iii. reconstruct all regions into a bmp image and write back to PDF file

The author of K2pdfopt uses only image processing method to extract the typesetting information and there is no need to extract any word from DJVU files.

tigran123:
@chrox Yes, I gathered that it does this (without looking at the code) by observation that it preserves mathematical formulas intact, i.e. scans them in as a single object into a bitmap image. If it was extracting text character by character and then re-rendering them using the glyph data then it would totally mess up the mathematical formulas, just like Adobe Digital Editions does. That is why I value your work very highly --- when it is ready to be merged into KPV master we will have a better reflow than even Adobe does! :)

houqp:
@tigran123 , wow, fortune looks pretty cool, I have added it to my zshrc :)

The processing algorithm must be very interesting, I guess I should take a look on it when I have time ;p

PR #455, opened 2012-10-16
Get rid of unused variable in getCacheSize()

tigran123:
Trivial cleanup of `pic.c`

### DIFF
--- a/pic.c
+++ b/pic.c
@@ -53,7 +53,7 @@ static uint8_t *rgbToGrayscale(uint8_t *image, int width, int height)
 
 	if (!buf) return NULL;
 
-	for (x = 0; x<width; x++)
+	for (x=0; x<width; x++)
 		for (y=0; y<height; y++) {
 			int pos = 3*(x+y*width);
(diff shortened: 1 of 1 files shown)


pic.c
```diff
@@ -53,7 +53,7 @@ static uint8_t *rgbToGrayscale(uint8_t *image, int width, int height)
 
 	if (!buf) return NULL;
 
-	for (x = 0; x<width; x++)
+	for (x=0; x<width; x++)
```
houqp, line 56: Ok, I guess we need to come to an agreement on coding style for spaces in for conditions ;P Many of our code add spaces in deed. But since you are the C expert here, so I will leave this decision to you :)

tigran123, line 56: Actually, my view is that one doesn't have to be very _strict_ about it, namely the common sense always suggests the "correct" answer depending on the context: for example, if it is a complicated enough `for()` like this:

```
/* standard macro for walking the list while skipping its head */
#define list_for_each(p) \
    for (p = &pcaps_head; (p = p->next) != &pcaps_head ; )
```

then spaces are in order. But if it is a trivial one like `for (x=0; x<width; x++)` then, imho, there is no need for spaces.

houqp, line 56: Got it, agree :)

PR #459, opened 2012-10-18
code cleanup

chrox:


tigran123:
If you are deleting `Shift-L` and `L` then you should also provide PDFReader:setDefaults() method with these lines:

```
function PDFReader:setDefaults()
    self.show_overlap_enable = true
    self.show_links_enable = false
end
```

i.e. set show_links_enable = false to prevent the link underlines from being displayed, as they will be displayed in the wrong place anyway.

And the other line: show_overlap_enable = true assumes that your page views overlap, so it maintains the same behaviour as in non-reflow mode. And if they don't overlap then you must set it to false, obviously.

tigran123:
Ok, thank you. One other minor comment: on initialization you are currently adding ALL commands twice: once from :init() and once from :adjustPDFReaderCommand(). In the `else` part of the function `adjustPDFReaderCommand()` it should not add ALL commands, but only adjust as appropriate for non-reflow mode. So, it should work like this:
1. On the initialization, call `self:addAllCommands()` and then `self:adjustPDFReaderCommand()` as you currently do, but the latter shouldn't add ALL commands under any condition.
2. Every time the user presses `R` call some function (whether you fix the `adjustPDFReaderCommand()` for this purpose or write  a new one is up to you) to add/delete the commands as appropriate for the chosen mode.

tigran123:
actually, on initialization you don't need to adjust any commands as the default value of reflow is false. You only need to do this on every press of `R` command.

Correction: it is slightly more complex than that. Initialization of the PDFReader object happens on startup, but you will need to adjust the commands if the user opens a PDF file with the reflow=true in its history .lua settings file.

tigran123:
Ok, I am merging this PR but in the next PR please begin tidying up the command initialization as described above.

chrox:
@tigran123 Thanks. I'm now working with the separate reader approach. And the so called KOPTReader handles both DJVU and PDF files. The koptreader can display the first page of a pdf file but after that it crashes when I press any key  saying that lua config error: ./unireader.lua:2234: attempt to index field 'commands' (a nil value). Could you repeat that bug? I just have no idea how does it come out. The commit is in the koptreader branch of  git://github.com/chrox/kindlepdfviewer.git .

tigran123:
Such error indicates that no commands were added (the "addAllCommands()" was not called).

Ok, so the KOPTReader will handle both PDF and DjVu files, good. Then, we will need to toggle between the normal (PDFReader/DJVUReader) and KOPTReader whenever the user presses the `R` key and also from filechooser.lua if the history file exists (i.e. examining how the file was read the last time it was opened).

chrox:
@tigran123 I do have called self:addAllCommands() in KOPTReader:init() method. It's really weird.

chrox:
@tigran123 Btw, should I request a pull for apparently buggy code like this so that conversation about the bug could be easier?

houqp:
@chrox , you need to call `KOPTReader:init()` also in reader.lua(around line 158). It turns out that all the init method for readers are not called automatically. I think it should be called automatically when user opens a document...

houqp:
@chrox yes, you can send a pull request even the code is not finished, but make sure that every commit message is relevant to the content of the commit. So afd97e0d43802ad1b89d4d2082c75de08a5135e4 is a bad example ;P

chrox:
@houqp Thanks. I got it.

tigran123:
@houqp I don't think that `:init()` should be called automatically on `:open()` unless you carefully ensure that it is done only on the _first_ open.  There is no point in adding the same commands again and again every time the user opens a new document of the same type.

tigran123:
concerning pull requests --- if @chrox is rewriting the whole thing as a separate KOPTReader then, I think we should put it in a separate branch because the changes against `reflow` branch would be meaningless, it is the changes against `master` branch that really matter and that should be carefully reviewed, right?

houqp:
> @houqp I don't think that :init() should be called automatically on :open() unless you carefully ensure that it is done only on the first open. There is no point in adding the same commands again and again every time the user opens a new document of the same type.

Yes this is what the new_ui_code branch is doing, we use a `is_initialized` local var to make sure the init method is only called once :)

But anyway, i guess we can leave it untouched for now, I think we have already added most of the reader, and mistake like this is not likely to happen in the future :)

tigran123:
Ok, great, I agree.

(we only have `MP3Player` reader still missing :P

chrox:
@tigran123 @houqp The DJVU reflow code is already out. It seems working just like PDF reflow. And I will send a PR after you create the koptreader branch.

houqp:
@chrox branch created, just give it a shot :)

houqp:
> (we only have MP3Player reader still missing :page_facing_up: 

LOL, then we might also need AVI reader....

tigran123:
actually, the mplayer binary compiled for Kindle does support video playback and framebuffer fb device --- so I was going to copy some movie to my Kindle and see how it looks at about 2-3fps :) I was just too lazy to do it, as it would probably look awful if anything at all.

Update: tried to play video on Kindle --- mplayer dies with fatal error "can't initialize video". But with "-vo null" option it plays fine, but you only get audio stream, no video :)

PR #460, opened 2012-10-18
Koptreader: just another pdf/djvu reader with reflow support

chrox:


houqp:
Works like a charm :)

tigran123:
Congratulations, well done @chrox ! :)

So we now need to urgently figure out a way to switch back and forth between kopt and normal pdf/djvu readers, otherwise at the moment the build of the master tree cannot be made available to the public. This (I mean koptreader) is a most excellent feature that makes the remaining 5% of PDF and DjVu files useable on Kindle 3, but we must still maintain support for the 95% of PDF/DjVu files out there...

(I am definitely against making to the public a build with koptreader as _default_ but I hope nobody is suggesting that :)

tigran123:
Hmmm, we are under DDOS attack, still we continue to work even under bullets fire, so here are some comments on the patch:
1. In adjustCommands() you don't need to add the "< >" group as it is identical to the one created by unireader.lua anyway. (see PR #462)
2. Why did you leave the `S/Shift-S` zoom modes? Are you planning to use them? If not, then you need to delete the zoom mode menu selector command `M` as well.
3. I think the most urgent problem to sort out now is the switching between standard and kopt readers. What thoughts does everyone have on this? Or shall we wait for your next PR and then discuss it?

tigran123:
Also, another comment: as we are not running "MPW's Mr. C on the Macintosh" perhaps we can uncomment the 50-digits precision constants like PI, SQRT2 etc in `k2pdftopt.c` and comment out the 20-digit ones? Also, did you keep MS-DOS ^M carriage returns to make it easier to compare with the original?

tigran123:
Do we really need the `R` command and the ability to switch between the standard and kopt readers without leaving to file manager level? I suppose there are those who use `Shift-P-P` (read the last file) feature who would be annoyed by the need to run `Shift-P-D`, set the reader and then continue reading. But, ignoring these, how about we implement a simple toggle (or a menu, though menu for two items seems an overkill) bound to one of the few remaining spare keys that allows the user to select between "standard" and "koptreader" on filechooser level?

chrox:
@tigran123 I prefer with the filechooser implementation. There could be a file association mechanism to remember which reader the user chooses to open a specific document. The filechooser should promote the user with three options namely "Remember for this file type" "Remember for this document" and "Just for once" when one file type is associated with more than one reader.

tigran123:
@chrox Ok, so, if I understand you correctly it should be context sensitive, i.e. if the cursor is positioned on a PDF or DjVu file then if you press, say `R` in filechooser then it would present you with the menu:

Title = "Select the PDF/DjVu reading mode:"
choices = 
    "STANDARD for this file type"
    "STANDARD for this file"
    "STANDARD for all files"
    "KOPTReader for this file type"
    "KOPTReader for this file"
    "KOPTReader for all files"

There are lots of slight issues here, do you see them? I am not saying they are not solveable (of course they are), but just have to be careful. Namely, what do we do with the user's choice? Save it? Where? History file? But what if it doesn't exist? And what if the user never chooses to open that file at all and keeps making these decisions for many files, never opening them? Then these decisions would be lost if he exits KPV.

Update: and if the cursor is positioned on a non-PDF/DjVu file then it should still respond with a menu consisting of just two items:

"STANDARD for all files"
"KOPTReader for all files"

Correct?

chrox:
@tigran123 Of course not. We are talking about different things. I mean when opening file a reader is selected not only according to the file ext name but also with the full file name which has a record elsewhere marked the reader selected last time. If the file type is registered with more than one reader like PDF and there is no record for that document a dialog is popped-up promoting the user to select a reader. If a file with extension epub is opened, a crereader is returned as there is only one reader registered with epub file type.

chrox:
@tigran123 This file association mechanism should be transparent to individual readers. A new reader should simply register the file formats it supports in its init method. So that if someone provide a new epub reader it will be much easier to integrate it into the existing system. And readers could be plug-and-play.

tigran123:
@chrox Hmmm, that sounds quite complex. Are you volunteering to implement all this in a simple-enough way to be bug-free?

Let's move the discussion to #465 specifically created for this purpose.

PR #461, opened 2012-10-18
Minor optimization of UniReader:drawOrCache()

tigran123:
Cut down one access to a global, one table dereference, one call to C level getPages() and one division on each call to `UniReader:drawOrCache()` function, i.e. on each page draw including panning.

I personally would prefer to get rid of `page_indicator()` altogether (and save 6 function calls + 3 conditional checks), but if @dpavlin thinks it is needed during debugging then at least we can shift the calculation of `pg_w` local variable inside `page_indicator()` function.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1067,9 +1067,9 @@ function UniReader:drawOrCache(no, preCache)
 	-- #3 goal: we render the full page
 	-- #4 goal: we render next page, too. (TODO)
 
-	local pg_w = G_width / ( self.doc:getPages() )
 	local page_indicator = function()
 		if Debug('page_indicator',no) then
+			local pg_w = G_width / ( self.doc:getPages() )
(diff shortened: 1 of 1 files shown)


dpavlin:
I quite like page_indicator (since it shows page render and location in the document at same time), but since it does show down page refresh it's not best possible solution.

I would love to have some way to indicate to users that page is still rendering. Maybe we can change led color to indicate that page is rendering?

tigran123:
@dpavlin we probably can (@NiLuJe probably knows this sort of Kindle internals stuff) but I think it would be just as slow (if not slower) as rendering rectangles on the page.

I think page rendering should be optimized to be as fast as possible. Sometimes you want to quickly page through some section of the book (just like you do with a paper book) and even with rcountmax=200 it is still too slow...

NiLuJe:
I've never played with the LED, but, yeah, that might be doable. geekmaster on MR might have more details about that, though ;).

PR #463, opened 2012-10-19
Bugfix: update offset_y when global zoom value is changed

chrox:
When zoom out at the bottom of a page upper part of the page will not displayed because offset_y is less than the new min_offset_y. With this patch offset_y is set to offset_y*zoom/last_zoom making the first line of the page stay roughly still in the new view.

### DIFF
--- a/koptreader.lua
+++ b/koptreader.lua
@@ -240,14 +240,4 @@ function KOPTReader:adjustCommands()
 	self.commands:del(KEY_N, MOD_SHIFT, "N")
 	self.commands:del(KEY_L, nil, "L")
 	self.commands:del(KEY_L, MOD_SHIFT, "L")
-	--self.commands:delGroup("< >")
-	self.commands:addGroup("< >",{
-		Keydef:new(KEY_PGBCK,nil),Keydef:new(KEY_LPGBCK,nil),
-		Keydef:new(KEY_PGFWD,nil),Keydef:new(KEY_LPGFWD,nil)},
-		"previous/next page",
-		function(koptreader,keydef)
-			koptreader:goto(
-			(keydef.keycode == KEY_PGBCK or keydef.keycode == KEY_LPGBCK)
-			and koptreader:prevView() or koptreader:nextView())
-		end)
 end


chrox:
Sorry for mis-committing the gitignore file. How could that been reverted?

tigran123:
@chrox no problem, just use `git log` to find out the SHA1 number of the commit and then do `git revert` passing it the number. This will revert the specified commit.

chrox:
@tigran123 I will try it right now. Thanks. Btw, will the reverted commit mess up this PR? There are too many tricks that I don't know about the git tools. So in the case that the master has already had a patch that my branch provides is it proper to send a PR directly from my branch to the master?

tigran123:
No, in such case it is not proper, i.e. the PR will be shown to us (those with write access to master) as "not mergeable". We don't merge PRs by actually picking your commits, no, we simply click on this BIG GREEN button "Merge" and if the button is not green, then we can't merge it :)

You would have to `git revert` that commit about the command group + `M` command and then add another commit that just removes the `M` command. Then this grey button will turn green and we'll be able to merge your commit.

At a minimum we read your commits and review the code before merging the PR. Sometimes we even test the PR before merging. But in this particular bugfix you are changing something strictly koptreader-specific and I consider you more knowledgeable than myself in this (as you wrote this code), so I wouldn't test this specific PR and would merge it if it is "green".

Now, a word of advice for the future PRs. I strongly suggest that you don't work in your own master tree, but always do `git checkout -b branchname` where `branchname` is the name of the branch that you choose for the specific fix or enhancement. Then you work in that branch and commit into it and then send a pull request from that branch to hwhw/master branch. Anyway, it is up to you how you work, I am just trying to help you make things easier and smoother, but feel free to ignore my advice if you are more comfortable doing things differently, of course.

tigran123:
I forgot to mention something else. Once your PR (from the branch `branchname`) is merged into master, you do this:

```
$ git checkout master
$ git status
$ git pull --rebase upstream master
```

The `git status` is to show you that there are no modified files in the master. If there are then get rid of them BEFORE doing `git pull --rebase`!!!

```
$ rm reader.lua
$ git checkout reader.lua
```

I used `reader.lua` as an example because that is the most frequently modified file for the change `kpdfview` -> `kpdfview.emulator` for running the emulator binary.

Then you can safely delete the branch `branchname`

```
$ git branch -d branchname
```

Also, after you have committed your changes to `branchname` locally, you still have to create the remore branch `branchname` on github.com in order to send a PR from it. This is how you do it:

```
$ git push origin branchname
```

(this assumes that `branchname` is your currently checked out branch).

Also, the above assumes that you have already created remote `upstream` pointing to hwhw/master. If you haven't, here is how to do it:

```
$ git remote add upstream https://github.com/hwhw/kindlepdfviewer.git
```

chrox:
@tigran123 Thanks for your advice and help.

tigran123:
I also forgot to show how to delete remote a remote branch (AFTER your PR from that branch was merged):

```
$ git push origin :branchname
```

For example, in your fork tree there are lots of old useless branches that have been merged into master --- they can be safely deleted to keep your own tree tidy --- have a look at how my fork tree looks :)

tigran123:
Another advice I would like to give is to make PRs as SMALL AS POSSIBLE, i.e. do not put too many commits into it. It is better to send 10 PRs with 1 commit each than 1 PR with 10 commits in it. The only situations where there should be more than one commit in a single PR are:
1. when commits are logically related, but not so tightly related as to go into a single commit. Also, you cannot put some changes in PR A that rely on PR B that is not merged yet. I.e. all PRs should be mergable individually.
2. when your PR is dedicated to the "main change" but you also add some trivial commits (like whitespace changes or comment updates) for which it seems silly to create a separate branch.

Remember what Linus said -- always spoonfeed me with patches! :)
What is right and applicable to the Linux kernel is right and applicable to all other software.

Using branches with git is so easy and convenient (and there is no disk space overhead, as long as you remember to delete them afterwards) that it is a very good idea to create a short-lived branch for every change you make.

chrox:
@tigran123 Your advice is really helpful. Thank you again.

houqp:
Good tutorial @tigran123 :)

I think we should copy them to WIKI so other new developers can learn from it too.

PR #470, opened 2012-10-20
PaperWhite support for einkfb module

houqp:
Rewrite eink update method to support PaperWhite eink update.

### DIFF
--- a/einkfb.c
+++ b/einkfb.c
@@ -75,16 +75,17 @@ static int openFrameBuffer(lua_State *L) {
 	}
 
 	/* mmap the framebuffer */
-	fb->buf->pitch = fb->finfo.line_length;
 	fb_map_address = mmap(0, fb->finfo.smem_len,
 			PROT_READ | PROT_WRITE, MAP_SHARED, fb->fd, 0);
 	if(fb_map_address == MAP_FAILED) {
 		return luaL_error(L, "cannot mmap framebuffer");
 	}
-	if (fb->vinfo.bits_per_pixel != 4) {
-		/* for 8bpp K4, we create a shadow 4bpp blitbuffer 
-		 * K4 uses 16 scale 8bpp framebuffer, so we still cheat it as 4bpp */
-		fb->buf->pitch = fb->buf->pitch / 2;
+	
+	if (fb->vinfo.bits_per_pixel == 8) {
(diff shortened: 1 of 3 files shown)


dpavlin:
@houqp eventually you will push me over the edge, and I will get PaperWhite just to see how our reader works on it. Fortunately Amazon doesn't seem to want to ship it to Croatia still, so I'm safe for now :-)

tigran123:
@dpavlin I was tempted with exactly the same thoughts yesterday... We can already order it in the UK and I thought "ah, perhaps I should just get the WiFi version, just to test @houqp 's work on the paperwhite" :)
But it is still £139 for WiFi and £200 for 3G version (don't look at Amazon's prices as published --- they are deceptive, you have to include the price of the proper leather cover as well, of course). 

@houqp please tell me, is the screen really as bad as I predicted (by observing the "glowlight" feature of Nook Touch Glowlight I own)? Or is the use of the better screen underneath this stupid "light guiding layer" compensates enough to make the final result bearable, i.e. not much worse than Kindle 3?

dpavlin:
http://www.youtube.com/watch?v=phO4QkHWsVE seem to show that it's different enough from alternative solution to be good

houqp:
For anyone who is curious how the new ui branch looks like in PaperWhite, I uploaded a photo taken by my smartphoe ;P

http://www.flickr.com/photos/houqp/8109333525/in/photostream

@tigran123 , well, I have my kindle3's screen broken 1 month ago, so I cannot do a comparison here. So far I haven't read any book on PW yet, because the native reader sucks. The final result is bearable for me, but I might be wrong, because I cannot remember how good my kindle3's screen is ;p 

The only thing that I am not satisfied with is the built-in light. It is not evenly distributed as amazon's AD claimed... Most of the part is even, but the bottom line(about 10px) is not. But I won't use that feature often anyway ;p

hwhw:
Just a note from a still too busy developer from the corner of the
playground: I've pre-ordered a PW - and it should arrive here on November,
22nd. There's hope for me getting my hands dirty when the winter arrives
here in Germany :-)

NiLuJe:
I also have one (along with a K4 & Touch), so if you need one more guinea pig, I'm up ;).

houqp:
@hwhw  cool! That should make the porting process much faster ;P

@NiLuJe , that will be great, we definitely need some one to test it on KTouch when the new UI framework is done :)

PR #471, opened 2012-10-20
add page size limit in reading pdf/djvu files

chrox:
Occasionally koptreader will crash the KPV when reading a particularly large page in PDF/DJVU file. In this patch the maximum page size is set to 3000 x 4000, that is allocating no more than 12MB memory in reading one page. In most cases, this page size should be enough. In the future the maximum page size could be user configurable.

### DIFF
--- a/k2pdfopt.c
+++ b/k2pdfopt.c
@@ -90,18 +90,10 @@
  ** Constants from the front of the CRC standard math tables
  ** (Accuracy = 50 digits)
  */
-/* The 50 digits cause problems with MPW's Mr. C on the Macintosh, */
-/* so I've truncated to 20 digits.                                 */
-/*
- #define PI      3.14159265358979323846264338327950288419716939937511
- #define SQRT2   1.41421356237309504880168872420969807856967187537695
- #define SQRT3   1.73205080756887729352744634150587236694280525381039
- #define LOG10E  0.43429448190325182765112891891660508229439700580367
- */
-#define PI      3.1415926535897932384
-#define SQRT2   1.4142135623730950488
-#define SQRT3   1.7320508075688772935
-#define LOG10E  0.4342944819032518276
+#define PI      3.14159265358979323846264338327950288419716939937511
+#define SQRT2   1.41421356237309504880168872420969807856967187537695
+#define SQRT3   1.73205080756887729352744634150587236694280525381039
+#define LOG10E  0.43429448190325182765112891891660508229439700580367
 #define DBPERNEP    (20.*LOG10E)
 
 #define SRC_TYPE_PDF     1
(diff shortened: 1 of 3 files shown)


chrox:
@dpavlin Could you do me a favor to test this patch on DXG and see if it crashes like before? I heard that DXG has a  smaller memory than K3. Thanks.

tigran123:
@chrox I suspect there is a memory leak somewhere because when using koptreader and paging through a large file the memory requirement (both the total and RSS) keep growing.

As for DXG, I have it too, so I can test it. But I thought that on DXG koptreader wouldn't even start, i.e. it would crash immediately. But I will test to see if with this patch it actually starts up and will let you know.

chrox:
@tigran123 I noticed that too. And I have almost located the bug in K2pdfopt source. When the master bitmap is not large enough to contain the reconstructed small regions, it will grow automatically. But some memory is not freed in the growing process. I will look into it and fix the bug later.

tigran123:
@chrox It still crashes on KDXG (with nothing in crash.log) when you try to zoom beyond 1.3 on page 7 of this file:

http://urantia-book.org/downloads/The-Urantia-Book-Uversa-Press.zip

Also, it is a bit confusing that the zoom value is changing but nothing actually changes. When it restricts the zoom the user should get some sort of indication that increasing it further is useless, some kind of warning message.

tigran123:
Also, could you explain what do you expect to happen when the user opens a document with the physical page size greater than 3000x4000? I thought that the restriction is really on the zoom level, right? I.e. pressing `Alt-Next` key will have no effect.

Also, I hope we are not restricting functionality on Kindle 3 for the sake of not crashing on KDXG, because it is Kindle 3 that koptreader will be most useful on, not DXG.

tigran123:
Having said all that, the current PR seems fine, i.e. you can fine-tune the limitation and also cleanup the zoom variable handling (to make the fact of limiting visible on Lua level) later. Merging now.

chrox:
@tigran123 As far as I understand, there is no such "physical" page size for pdf/djvu document. When rendering a pdf page, one should specify a transform matrix which is generated by fz_scale function with dot-per-pixel as the parameter to get a pixmap. So the actual size of the rendered pixmap depends on both page dimension in inch and the transform matrix the user provides. Similar mechanism is applied for djvu page rendering. In this patch the rendered pixmap is restricted within 3000*4000 pixels. That is if we open a pdf page with a width of 20 inches, the dpi will be decreased to 150 so that the width of the rendered pixmap is no more than 3000 pixels.

tigran123:
@chrox For PDF files, yes, your understanding is correct. For DjVu files, no, there certainly is such a thing as physical page size, namely the numbers returned by `getOriginalPageSize()` function in `djvu.c`. These are the same as `iw` and `ih` values used by `k2pdfopt_djvu_reflow()` function in `k2pdfopt.c`.

chrox:
@tigran123 The function k2pdfopt_djvu_reflow in K2pdfopt uses the same size for pagerect and renderrect. So no matter how large or small the rects are set, full page is rendered. The restriction on page size only affects the maximum font size in reflowed view.

tigran123:
@chrox Ok, thank you, I understand now.

PR #476, opened 2012-10-20
remove unused variable reflow_mode_enable in pagehash

chrox:
There is no such variable as reflow_mode_enable but why the Lua interpreter just run this line as happy as there were? Is this a language feature?


tigran123:
In Lua you don't need to define a variable in order to use it. And in that particular expression `nil` (value of a non-existent variable) works in the same way as `false`. But if you tried to concatenate this variable to a string or convert it to number then Lua would throw out an error.

PR #479, opened 2012-10-21
Switch between standard and koptreader

tigran123:
This PR completes my implementation of the switch between the standard and koptreader.

PR #480, opened 2012-10-21
add readerchooser to prompt users to select alternative readers

chrox:
The behavior of readerchooser is simple. If only one reader in the registry is associated with the file type of the to be opened document, the associated reader is returned immediately without prompting the users anything. If more than two readers are registered with the file type of the to be opened file, local setting will be retrieved to see if the user has saved the reader preference previously. If the user has saved it earlier, return the saved reader immediately without prompting. If there is no saved preference with this file, a reader choosing dialog will be displayed to prompt the user with options to save preference of this document and not save it. Note that only the name of the reader is saved as a string in settings. More tests are needed. And the GUI needs to be polished certainly.


chrox:
Still has bugs. The reader choice could not be stored or retrieved in the setting even the user has selected "Always" option in the reader choosing dialog.

Updates: This bug has been fixed in commit 34d8c32 of this PR.

tigran123:
I am back now. Very interesting. I haven't tried it yet but looked very briefly at the code. You got rid of `ext:getReader()`, but did you notice the callers thereof in crereader, filechooser, fileinfo, filesearcher? Note that these do NOT really need to get the full reader objects --- all they need is just a YES/NO answer to the question "is this file type supported?" so you can satisfy them with a very simple function that scans the concatenation of all supported extensions.

Also, your comment says "more than 2 readers". Did you really mean this or did you mean "2 or more readers". I hope you meant "2 or more readers"... And the check `if readers[2] then` is a bit non-intuitive... Is the `readers` table guaranteed to have no holes?

I didn't have time to review it properly. It is time for dinner here and after that I'll be reading. I will review and will actually try running it tomorrow.

chrox:
I didn't notice there were other modules that required the extension module. I ran a grep 'require "extension"' \* command and found only reader.lua requires it. So I thought  deleting the extension module should be safe. And I tested it partially and knew that opening epub files is OK.

houqp:
I will be very busy in following four days, so will take a break from KPV's dev. Feel free to ignore me and merge the PR when you guys decided which implementation is better.

Happy hacking :)

chrox:
Now the `readerchooser.lua` behaves as expected except that some document settings such as `globalzoom`, `globalzoom_mode`, `bbox` are shared between `koptreader` and `pdfreader`/`djvureader`. There settings should be differentiated by each reader.

dpavlin:
I merged a PR which removes `alt_getopt.lua` so this PR doesn't merge any more. Can you please rebase it on top of current master? After rebase you will have to use `git push origin readerchooser -f` (notice `-f` at end) otherwise it won't allow you to push changes because all commits will be changed after rebase.

dpavlin:
Is this PR intended to be merged togehter with #479 of does it replace it?

chrox:
@dpavlin This PR is not compatible with #479 . Both #479 and #480 aim to provide an approach to switch between `pdfreader`/`djvureader` and `koptreader`. We should make sure to choose a better one.

tigran123:
@dpavlin Yes, it would replace it. Although I have not tested @chrox implementation, from looking at the code it appears to be more elegant solution. So, if it is tested and found to work (of course we believe you, @chrox but it is good if more than one person confirms that it works before merging) then I suggest we merge his solution instead of #479.

tigran123:
@chrox btw, if `readers` table is guaranteed to have no holes then the code `if #readers > 1` is more readable than `if readers[2]`.

dpavlin:
I tested this PR using @chrox branch and it seems to work well, but I don't like it's user visible workflow. Let me try to explain my objections:
1. every pdf open now asks me additional question (this could be solved by global default for all pdf files -- this is different for always because that's for always for this file). In 99% of cases, I would like to use old pdf viewer so additional prompt for each file is total overkill.
2. there is no information for user what is difference between KOPTReader and PDFReader (we should display information that first option reflows document, while second doesn't)
3. I don't like that reflow is default (first) option. While I do understand why @chrox made it that way (he seems to prefer reflow) it's change from "old" kindlepdfviewer and I can imagine user confusion when they just select KOPTReader and get different view of pdf file

Having said all that, I really like reflow and appreciate @chrox effort to make it work, so I would like to include it in master because it's very useful for some documents.

chrox:
@dpavlin @tigran123 So the options in the prompting dialog should be "Always with this type(T)" and "Always with this file(F)". And by default neither of them is selected which means don't remember this choice.  As for the reader position I swear it's not my intention. You can look into the code in `readerchooser.lua`. There is no bias in favor of one certain reader. I even put the `KOPTReader` in the third place after `PDFReader` and `DJVUReader` in the registry. I was also curious on why every time `KOPTReader` is returned before `PDFReader` and `DJVUReader`. Maybe `table.insert(readers,key)` just add the key at the beginning of the table. But I had no idea of that when writing the code. I should try to change `KOPTReader` to the first place in the registry and see if it will be better. Thanks.

Updates: I use a priority element in the reader registry. And the original `PDFReader`/`DJVUReader` comes at the first place now.

chrox:
@dpavlin @tigran123 Please tell me if it conflicts when merging with this PR.

tigran123:
I agree with everything @dpavlin said above and would like to add the following:
1. those former usages of ext:getReader() in filechoose, fileinfo, etc do NOT need the full reader object --- all they need is to validate whether the given extension is supported or not. For this you need to provide a simple function that does a search on concatenated string containing all supported extensions. Oh, and please don't even think about concatenating at runtime --- it should be done by hand, manually and statically hardcoded (to avoid having to handle duplicates and generally to avoid over-engineering).
2. code like `if readers[1]` and `if readers[2]` should be replaced with more conventional `if #readers > 0` and `if #readers > 1` because strictly speaking it means "is the first element of the table not nil and not false?" which is not what you intend. Code should be as readable as possible.
3. why do you keep the number of readers in self.n_readers? isn't #self.readers good enough?
4. Please remove KEY_HOME from the list of keys that returns from the dialog. KEY_HOME should be reserved for quitting the application only.

tigran123:
@chrox there are no conflicts, so if you address the things @dpavlin and I mentioned I see no reason why it shouldn't be merged into master.

chrox:
@tigran123 At `crereader.lua` Line 73 and `fileinfo.lua` Line 150 and `unireader.lua` Line 1010 the return of `ReaderChooser:getReaderByType(ftype)` is compared with reader object `CREReader`. That's why reader object is returned in `ReaderChooser:getReaderByType(ftype)`.

dpavlin:
"Always with this type(T)" and "Always with this file(F)" are great solution so please just ping us when you are ready to merge this pull request. I would love to make it work on fill screen width of DXG but we have to first merge this changes in master :-)

tigran123:
@chrox Ah, I see, you've done the right thing, i.e. your getReaderByType() function does exactly what I want --- scans the local data in memory. I thought you were reading settings from disk (like the *ByName() function) and that would definitely slow things down :)

Although I still didn't have spare time to test it, I cannot spot anything wrong in the code itself, so perhaps I'll test it after it's been merged into master already :)

tigran123:
Hmmm, in the latest patch you decided to return KEY_HOME to KEY_BACK handler? Don't you think that pressing `Home` key accidentally twice and thus closing the whole KPV is going to annoy the users?

chrox:
@tigran123 All dialogs in KPV use this two commands namely KEY_HOME and KEY_BACK to return including "Go page" dialog in filechooser. So make reader chooser dialog return with KEY_HOME and KEY_BACK is reasonable.

tigran123:
@chrox No, certainly not "all dialogs" --- the "Last documents", HelpPage and "Search" dialogs use only KEY_BACK to return. I thought that I've fixed selectmenu dialogs to do the same, but looks like I forgot to do this (will do tomorrow, together with others I find). I don't think it is reasonable to use the same key to quit a dialog that also quits the application because, as I said, you can very easily press `Home` key twice and instead of exiting the current dialog you exit the whole application.

tigran123:
Also, fileinfo and directory info use only `Back` to return, not `Home`.

Thanks for pointing out that some dialogs still use KEY_HOME for exit. The fix for selectmenu is in the PR #483. I will deal with the others later.

chrox:
@tigran123 Yes I find that not all dialogs use `Home` key to return now. If you plan to unify the return behavior of all dialogs, eg. disable the `Home` key of returning dialog. Then I will disable it in readerchooser dialog immediately. Btw, the search file dialog also can return with `Home` key.

tigran123:
@chrox by "search" I meant "search results". As for "search" --- not anymore, as long as PR #483 is approved and merged into master.

tigran123:
@chrox Ok, thank you :)

chrox:
@dpavlin @tigran123 You can merge this patch now after you test it of course.

chrox:
@tigran123 @dpavlin Now the `readerchooser` dialog will show `PDFReader` along with `PDFReflow` for users to choose. And `PDFReader` always comes at the first position because the priority of `PDFReader` is higher than `KOPTReader` in the `registry`. Default reader for certain file type is also available once the user selected a reader with the option `"Remember this type"`.

chrox:
@tigran123 @dpavlin And `SHIFT+C` command in reading will reset the default reader preferences. `ALT+C` command will clear reader association with the document being read.

tigran123:
I am testing it now. `Alt-C` does not seem to work.

Also, once the selection "for this file type (T)" is made, how does one revert it? It seems irreversible...

chrox:
@tigran123 You can revert "Remember this type" option by `SHIFT+C` command when reading.

tigran123:
Ok, the problem is due to your choice of very long text for messages. If the message is too long then InfoMessage:inform() will just not print it at all instead of truncating it. This is the quirk of our architecture :)

So, shorten them like this:

Reseting default reader preferences => Reseting to default reader

This doc has not been associated with a reader yet => No reader associated

Also, please make sure that any other :inform() messages are short enough to be visible. I only checked the above two.

Wonderful patch, works like a charm! :)
After you have tidied up the long messages it can be merged.

tigran123:
Btw, a very useful enhancement in the future would be to improve :inform() to handle long messages by breaking them up into multiple lines and displaying appropriately.

chrox:
@tigran123 I have shortened the info messages as this:

Reseting default reader preferences => Reseting reader preferences
This doc has not been associated with a reader yet => No reader associated

Thanks for your suggestion.

tigran123:
Ok, great, merging now.

Thanks again for your wonderful contributions. I actually like this reader-switcher even more than reflow itself, although the whole point of the former is subservient to the latter :)

chrox:
@tigran123 Thanks. My next work would be much harder in digging into the k2pdfopt source to get ride of the memory leak bug and fix the glitch you mentioned earlier that some bottom part of a page is never displayed.

tigran123:
Excellent stuff, thank you!

Btw, if you (or anyone else) noticed the "br" branch in my local repository which doesn't have koptreader, crereader and even my own picviewer --- let that not worry anyone or give any hints that I am not happy about anything being done in KPV master. NOT AT ALL! I am very happy the way we all work as a friendly (I would say brotherly even) team, with respect for one another and unselfish dedication to the service to all mankind. And I continue to consider myself a part of this team. However, I also know from experience that big programs (and with the addition of k2pdfopt source I think we just tipped the border between "medium" and "big" size) have inevitable instabilities and the thing is that (standard) PDF/DjVu support in KPV for me is something very much "mission-critical", so I decided to isolate the core of KPV into what one might call "KPV Light" and maintain it in a separate branch, merging only bugfixes and important improvements relevant to core PDF/DjVu functionality. I called it "BR" = "Book Reader" to emphasize that it is not for ebook formats (epub,mobi, fb2, etc) but for equivalent-to-printed books only (i.e. PDF and DjVu).

But of course I shall continue to use KPV on my Kindle alongside BR. E.g. for the sake of reflow.

dpavlin:
@chrox and @tigran123 thanks a lot for your effort to make this feature land in master.

dracodoc:
@chrox Sorry I don't quite understand about this feature. 

How many file types are available? I assume there are only pdf or maybe djvu. If user chose to open pdf with kpv by default, then how can he open certain pdf with reflow? I tried ALT+C, but it doesn't seem to change the default with kpv behavior with certain pdf, so that pdf will still be opened with kpv. Unless I use Shift+C to clear all default, there is the dialog again to select, but user will need to select for every file which defeat the purpose of the default/specific design.

It seemed to me the ALT+C only cleared the association build by "select reader for this file", but it will not override the default reader by file type on this specific file. 

Another method would be add a method to switch reader by a command, I remembered see it in issues but it wasn't built into current build.

chrox:
@dracodoc The story is that at the beginning I implemented a "R" command in original `pdfreader.lua` to toggle Reflow Mode on and off. But later we found it will mess something up in `pdfreader`. So I was suggested to move reflow mode to a separate reader which finally became the `koptreader`. And people always have their own tastes and requirement, as for choosing pdf readers in KPV, I prefer the `koptreader` for all my PDFs and @tigran123 prefers the original `pdfreader`. So we made an agreement to set a global reader preference which is associated with certain file type. And the global preference has the highest priority which means it will override the preference associated with specific document.

dracodoc:
If the global preference has the highest priority, then specific association does not exist anymore as long as global preference has been set. If user want to try specific association, he have to clear the global preference first, so these two options cannot coexist.

Why don't make the global preference as a default, and let the specific association work as exceptions to default? The scenario I'm thinking is like this:
1. user can select a default reader for file type. And any file with that type will be opened directly.
2. With certain file opened, pressing ALT+C should make that pdf to exception of default (just a "Not default" mark for that file), and quit current reader (User must want to change reader for that file, so it's better to setup exception mark and quit reader).
3. User open that file again, this time with the reader selection dialog. The specific selection will be memorized and the file will be opened directly next time. Now the file have a mark "Not default" and "reader selected".
4. If user want to change his mind on this specific file, press ALT+C again to quit and open, the "reader selected" mark is cleared, so he will get the reader selection dialog to choose.

This design should work with everybody's preference, and not too hard to implement.

chrox:
@dracodoc That's a great idea. I like that. So this time everybody should be happy. :) I will give it a patch on this issue. Thanks.

PR #482, opened 2012-10-22
Two bugfixes for link shortcuts

tigran123:
1. We can't use "/" shortcut for links because `KEY_SLASH` key is not present on Kindle 3 but we might want to read files on Kindle 3 with many links on the page.
2. When I fixed "last select menu item shortcut not reachable on Kindle" issue #467 bug I switched "/" and "Sym" positions, so the link shortcuts code needs to be made aware of the new values.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -3164,7 +3164,7 @@ function UniReader:addAllCommands()
 					local shortcut_nr = 1
 					shortcut_map = {}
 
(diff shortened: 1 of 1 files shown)


tigran123:
Ah, this fix is not complete, because if there are lots of links and we move down then one link between the two groups is skipped. I'll try to fix this now.

tigran123:
Ok, bug fully fixed now, I hope.

PR #483, opened 2012-10-22
Use only Back not Home to exit dialogs

tigran123:
It is not a good idea to use the key that exits the whole application for exiting a dialog, because accidentally pressing that key twice will exit the application and annoy the user.

### DIFF
--- a/inputbox.lua
+++ b/inputbox.lua
@@ -475,7 +475,7 @@ function InputBox:addAllCommands()
 	-- NuPogodi, 02.06.12: inputmode-dependent commands are collected
 	self:ModeDependentCommands() -- here
 
-	self.commands:add({KEY_BACK, KEY_HOME}, nil, "Back, Home",
+	self.commands:add(KEY_BACK, nil, "Back",
 		"back",
 		function(self)
 			self.input_string = nil
--- a/selectmenu.lua
+++ b/selectmenu.lua
@@ -238,7 +238,7 @@ function SelectMenu:addAllCommands()
 		HelpPage:show(0, G_height, sm.commands)
 		sm.pagedirty = true
 	end)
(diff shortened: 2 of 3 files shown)


tigran123:
Here is the real world situation (occurred 10 minutes ago here) showing how `Home` should be handled in selectmenu (specifically in TOC). I was explaining various properties of prime numbers to my daughter and needed to lookup "Goldbach Conjecture" (as there are several weak/strong forms thereof and I always forget which is which :) in Weisstein's CRC Encyclopedia. Obviously I used the PDF file on Kindle using KPV. So, pressing `T` key got me to the collapsed Table of Contents tree and I moved down to the letter "G" under which there were all the entries starting with "G". Ok, so I got where I wanted and encountered a reference to "TWIN PRIMES CONSTANT" (unfortunately, it was not a link, otherwise @dpavlin 's link code would have solved the problem). So, I pressed `T` again to go back to the TOC in order to look up the entry for "TWIN PRIMES CONSTANT". But now I am in the middle of the fully expanded series of entries starting with "G". What should I do to get to the "T" entries quickly? There is at present nothing I could do except to patiently page up many many times to get to the root of "G" and press fiveway `Left` to collapse it and then move down to "T" and expand it and get to the entry I want.

But if we implemented `Home` to actually mean "go back to the very first entry" (in global sense, not necessarily visible at present) then my navigation problem would be solved in far fewer key presses.

Tomorrow I will have a go at implementing this feature as it would obviously be very useful for books with large TOCs like Dictionaries and Encyclopaedias.

Or, even better and more generally, we could implement a "jump to position 0%, 10%,... 100%" like the one present in filechooser and also inside unireader. And why wait until tomorrow when this can be done today, see PR #484

PR #488, opened 2012-10-24
Extract attachments from PDF file via Alt-S

tigran123:
This PR contains the following commits:
1. Makes file info function more resilient, i.e. we shouldn't crash the whole KPV application just because we cannot discover the correct unpacked size of a zip file. This commit is unrelated to the main theme of this PR, so I sin against my own advocated rule, but this particular commit is trivial enough to be forgiven, I dare hope :)
2. The Space command in FileChooser should actually re-read the directory and not just redraw the current (possibly stale) data. This is especially useful in two situations: a) You have extracted some attachments in the current directory b) You wish to see the current state of the Last Documents list. The reason the Alt-S command does not do this automatically is because the file may have been opened via "open last file" (Shift-P-P) command and then the FileChooser object may not even exist.
3. Adds Alt-S command to PDFReader which invokes the external utility "extr" passing it the full pathname of the open PDF file and the current page number. The utility extracts all attachments on this page (if there are any) and saves them in the same directory as the PDF file. The file names given to attachments are decoded from within the PDF file itself, i.e. they are the same as the original file names of the files embedded in the PDF.

While you are reviewing this PR I'll write a little shell script which generates a PDF file with the given files embedded into it as attachments, as promised.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ LUALIB := $(LUADIR)/src/libluajit.a
 
 POPENNSLIB := $(POPENNSDIR)/libpopen_noshell.a
 
-all: kpdfview
+all: kpdfview extr
 
 VERSION?=$(shell git describe HEAD)
 kpdfview: kpdfview.o einkfb.o pdf.o k2pdfopt.o blitbuffer.o drawcontext.o input.o $(POPENNSLIB) util.o ft.o lfs.o mupdfimg.o $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) djvu.o $(DJVULIBS) cre.o $(CRENGINELIBS) pic.o pic_jpeg.o
@@ -143,6 +143,12 @@ kpdfview: kpdfview.o einkfb.o pdf.o k2pdfopt.o blitbuffer.o drawcontext.o input.
 		-lm -ldl -lpthread -ljpeg -L$(MUPDFLIBDIR) \
 		$(EMU_LDFLAGS) \
 		$(DYNAMICLIBSTDCPP)
+
+extr:	extr.o
+	$(CC) $(CFLAGS) extr.o $(MUPDFLIBS) $(THIRDPARTYLIBS) -lm -o extr
+
+extr.o:	%.o: %.c
+	$(CC) -c -I$(MUPDFDIR)/pdf -I$(MUPDFDIR)/fitz $< -o $@
 
 slider_watcher.o: %.o: %.c
 	$(CC) -c $(CFLAGS) $< -o $@
@@ -203,7 +209,7 @@ fetchthirdparty:
 	cd popen-noshell && test -f Makefile || patch -N -p0 < popen_noshell-buildfix.patch
 
 clean:
-	rm -f *.o kpdfview slider_watcher
+	rm -f *.o kpdfview slider_watcher extr
 
 cleanthirdparty:
 	$(MAKE) -C $(LUADIR) CC="$(HOSTCC)" CFLAGS="$(BASE_CFLAGS)" clean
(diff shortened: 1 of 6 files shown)


chrox:
@tigran123 Probably we need a Calibre plugin to share the wrapped djvu documents via Email.

tigran123:
@chrox why? I have tested sending a djvu file wrapped in a pdf via gmail and it was successful. Why would we need a special plugin for this? Seems to work without it. I only tested with embedded djvu but it should also work with mp3 and any other files. Ah, which reminds me, my shell script needs to set the proper mimetype for each file it embeds...

chrox:
@tigran123 I use Calibre to manage my ebook library. In the library view one can share a specific book via Email by one click. I usually send a PDF or MOBI book to my Kindle device. You know what I mean. With a Calibre plugin DJVU documents could be sent to Kindle devices by just one click.

tigran123:
@chrox Ok, yes, I understand. Actually, is Calibre scalable enough (like has hierarchical directory/collection or whatever management) to handle 500,000 books distributed in many directories? If so (and if the Linux version actually works) then maybe I should try using it as well...

I use the old-fashioned (but very fast and reliable) Unix rsync command to manage my book collections and sync them between multiple machines and storage devices.

chrox:
@tigran123 I have never tested that large collection of books in Calibre. So I just don't know. But I really like the convenience of sharing by email and the ability to download metadata of a book from Internet.

tigran123:
Ah, now I see what you mean by Calibre plugin! :) It won't import djvu files so it would be useless for my collection as the majority of books in it are in DjVu format...

However, I could wrap them all into same size PDF files and import them like that. I guess I should make my shell script wrapper-generator very intelligent, i.e. try to embed as much metadata into the PDF as possible.

chrox:
@tigran123 No, Calibre can import most common document formats like PDF,DJVU,EPUB,MOBI,TXT,HTML,PRC and so on. There must be something wrong you deal with Calibre.

tigran123:
Yes, true it does import DjVu files fine. But I would have to edit metadata manually for each book and even then it would not be hierarchical, i.e. I can set "Series" to "Bibles" but I need  subseries like "Bibles/Hebrew/Printed/Early", "Bibles/Hebrew/Printed/Modern" and "Bible/Hebrew/MSS". I don't see how one can do hierarchical management in Calibre.

tigran123:
The `extr` utility must be enhanced to handle the absolute pathnames encoded in the attachment. I didn't realize that these are possible... So, I'll fix it now to extract just the base part (i.e. the filename) of the absolute pathname and use it when saving the attachment. Otherwise it may try to save attachments like "/home/tigran/kindlepdfviewer/Books/File.djvu" which is obviously a non-existent path on a Kindle.

Btw, evince doesn't handle this situation correctly either. But Adobe Acrobat Reader does :)

tigran123:
Btw, the memory leak (in fact four memory leaks) in extr utility is intentional because it is an utility, not a library function.

tigran123:
Ok, as far as I am concerned this PR is completed. Please review.

tigran123:
Actually, I might as well add the check for non-existent input files instead of crashing in a messy way inside pdfLaTeX.

Also, perhaps instead of the interface "pdfattach -o file.pdf -i file1.djvu -i file2.djvu" a more intuitive interface would be "pdfattach -o file.pdf file1.djvu file2.djvu ..." so that one can just say "pdfattach -o file.pdf *.djvu *.mp3 *.fb2" and embed lots of files in one go, instead of specifying "-i" for each input file.

tigran123:
Another improvement to pdfattach would be to print the filename of the embedded file on the page. This is so that if one does something like "pdfattach -o Fiction-fb2.pdf fb2/Fiction/*.fb2" then he knows what stuff he is getting when he extracts the files on the Kindle.

tigran123:
Even more enhancements to pdfattach are desirable:
1. put just one file on a page, so the user gets complete control over what files to extract.
2. beautify the text a bit more (e.g. use monospaced font for filenames etc)

tigran123:
Ok, I am done.

dpavlin:
@tigran123 nice work :-) Should we think about 2012.10 release?

tigran123:
I don't see why not? Well, @hwhw was saying something about bugfixes that went into the LuaJIT version that we are using, so perhaps it makes sense to upgrade LuaJIT. I didn't figure out at the time how to do it using git submodules infrastructure. The way I see it when we do "make fetchthirdparty" the latest luajit is automatically cloned, but obviously this is NOT the case because if I clone it manually the result is different from the content of our luajit-2.0 directory.

So, should we upgrade luajit before 2012.10 release or not?

tigran123:
On the other hand, if there is no obvious need to upgrade then perhaps we shouldn't... I would hate to see some instabilities coming from the latest luajit --- the current state of KPV seems very nice and stable: just what one needs for a stable release! :)

hwhw:
Hm. LuaJIT did only receive bug fixes. We seem to do a good job not
triggering the bugs, so you _might_ be right. But some of these bugs were a
bit nasty IIRC - I'm monitoring the LuaJIT mailing list.

A short try did not unveil build errors and I guess it wouldn't get too
unstable. But I don't insist. Another big item is a muPDF update - MuPDF
too has received some bug fixes over the last months. Maybe both should
wait until right after release.

tigran123:
@hwhw Yes, I agree, let's wait until after release.

PR #489, opened 2012-10-24
bugfix: isolate doc settings of koptreader from pdfreader/djvureader and feedback zoom value used by page reflow

chrox:
`koptreader` interprets `self.globalzoom` differently from `pdfreader`/`djvureader`.So change the variable in one reader will have unexpected effect in another. I isolated the variable `self.globalzoom` in `koptreader` and use another variable `self.kopt_zoom` so that when switching between `koptreader` and `pdfreader`/`djvureader` zoom values are stored independently in doc settings. I also introduced a new variable `self.kopt_gamma` which will have different visual effect from `self.gamma` in the future. Because modifying gamma costs very much in `koptreader` (full page will be re-reflowed) I used a different modification algorithm multiplying the original gamma with square of factor. So the changing step will be bigger.

Updates: in commit 564f73d `reflow` function will return zoom value used in K2pdfopt. Then `self.kopt_zoom` is set with the returned zoom value so that when zoom value exceeds the upper limit of reflowable page size user will notice that zoom value cannot be increased.

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -476,16 +476,16 @@ static int reflowPage(lua_State *L) {
 	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
 	ddjvu_render_mode_t mode = (int) luaL_checkint(L, 3);
 
-	double dpi = 250*(dc->zoom);
-
 	int width, height;
-	k2pdfopt_djvu_reflow(page->page_ref, page->doc->context, mode, page->doc->pixelformat, dpi);
+	k2pdfopt_djvu_reflow(page->page_ref, page->doc->context, mode, page->doc->pixelformat, dc->zoom);
 	k2pdfopt_rfbmp_size(&width, &height);
+	k2pdfopt_rfbmp_zoom(&dc->zoom);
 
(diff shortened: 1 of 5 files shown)


chrox:
I'm sorry but commit 564f73d solved another issue and I forgot to create another branch so it is appended to this PR.

tigran123:
No problem, these are all koptreader-specific things, so they are logically related, sort of.

tigran123:
I have just tested the zoom feedback thing and on some PDF files it doesn't work, i.e. the zoom value shown when you press Alt-Next keeps growing until I begin to get a steady stream of messages like this in the log:

```
reading page:2147483647,2147483647,2147483647,2147483647 dpi:-nan
```

 But on other PDF files it seems to work. Here is the PDF file on which it fails --- it is actually the simplest PDF file one can think of, the "Hello, world" PDF file:

https://github.com/downloads/tigran123/libdjvu/hello.pdf

chrox:
The bug has been fixed. And the default margin of reflowed page is changed to 0.06 which is approximately 10 pixels.

tigran123:
Yes, seems to work now, thanks.

PR #492, opened 2012-10-25
pass framebuffer size to reflow

dpavlin:
This allows us to use full screen size on DXG or in emulator when emulating resolution different than K3 600*800

### DIFF
--- a/pdf.c
+++ b/pdf.c
@@ -515,8 +515,9 @@ static int reflowPage(lua_State *L) {
 
 	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
 	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
-	int width  = luaL_checkint(L, 3); // framebuffer size
-	int height = luaL_checkint(L, 4);
+	// render_mode is 3rd param and unused
+	int width  = luaL_checkint(L, 4); // framebuffer size
+	int height = luaL_checkint(L, 5);
 
 	k2pdfopt_mupdf_reflow(page->doc->xref, page->page, page->doc->context, dc->zoom, dc->gamma, 0, width, height);
 	k2pdfopt_rfbmp_size(&width, &height); // overwrite with generated size
--- a/pdf.c
+++ b/pdf.c
@@ -515,10 +515,11 @@ static int reflowPage(lua_State *L) {
 
 	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
 	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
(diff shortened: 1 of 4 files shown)


kpvcrlib/crengine
```diff
@@ -1 +1 @@
-Subproject commit a9516b455e811b9f57f2805d6c08b6744f797083
+Subproject commit b62d054cca57366bc1ba2fa249c5bbf5c5234346
```
tigran123: does this change to `kpvcrlib/crengine` imply we are switching to a newer crengine version?

dpavlin: Good question. I'm not quite sure how that ended up in my repository, so I'm removing it until I investigate it.
It shouldn't be part of this PR anyway.

pdf.c
```diff
@@ -515,10 +515,11 @@ static int reflowPage(lua_State *L) {
 
 	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
 	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
+	int width  = luaL_checkint(L, 3); // framebuffer size
+	int height = luaL_checkint(L, 4);
 
-	int width, height;
-	k2pdfopt_mupdf_reflow(page->doc->xref, page->page, page->doc->context, dc->zoom, dc->gamma, 0);
-	k2pdfopt_rfbmp_size(&width, &height);
+	k2pdfopt_mupdf_reflow(page->doc->xref, page->page, page->doc->context, dc->zoom, dc->gamma, 0, width, height);
```
chrox, line 522: `k2pdfopt_mupdf_reflow(page->doc->xref, page->page, page->doc->context, dc->zoom, dc->gamma, 0, width, height);`
should be 
`k2pdfopt_mupdf_reflow(page->doc->xref, page->page, page->doc->context, dc->zoom, dc->gamma, 0.0, width, height);` 
or bb_width will be 0. I don't know why. Shouldn't an integer instant be cast automatically to a double when calling function? I only test on emulator on Ubuntu 12.04 AMD64.

pdf.c
```diff
@@ -515,10 +515,11 @@ static int reflowPage(lua_State *L) {
 
 	PdfPage *page = (PdfPage*) luaL_checkudata(L, 1, "pdfpage");
 	DrawContext *dc = (DrawContext*) luaL_checkudata(L, 2, "drawcontext");
+	int width  = luaL_checkint(L, 3); // framebuffer size
```
chrox: This should be `int width  = (int) luaL_checkint(L, 4);` Because the third parameter is `render_mode` for DjvuReflow. And accordingly the next line is `int height = (int) luaL_checkint(L, 5);`

dpavlin: You are right. Why are we passing render_node which isn't used?

dpavlin:
obsoleted by #493

PR #494, opened 2012-10-26
some UI tweaks on koptconfig and readerchooser to accommodate DXG screen

chrox:
The `readerchooser` uses fixed width and height of window size because window of readerchooser is almost full all around if we enlarge the window according to DXG screen it will be too sparse at some corners. While I use horizontal margin layout for `koptconfig` as option names and option items are aligned to fixed proportions of the window width.

### DIFF
--- a/koptconfig.lua
+++ b/koptconfig.lua
@@ -29,10 +29,10 @@ KOPTConfig = {
 	OPTION_PADDING_T = 50, -- options top padding
 	OPTION_PADDING_H = 50,  -- options horisontal padding
 	OPTION_SPACING_V = 35,	-- options vertical spacing
-	VALUE_PADDING_H = 150,  -- values horisontal padding
-	VALUE_SPACING_H = 10,   -- values horisontal spacing
+	ITEM_PADDING_H = 150,  -- items horisontal padding
+	ITEM_SPACING_H = 10,   -- items horisontal spacing
 	OPT_NAME_FONT_SIZE = 20,  -- option name font size
-	OPT_VALUE_FONT_SIZE = 16, -- option value font size
+	OPT_ITEM_FONT_SIZE = 16, -- option item font size
 	
 	-- last pos text is drawn
 	text_pos = 0,
@@ -70,7 +70,7 @@ function KOPTConfig:drawOptionItem(xpos, ypos, option_index, item_index, text, f
 		self.text_pos = 0
 	end
 	
-	local xpos = xpos+self.OPTION_PADDING_H+self.VALUE_PADDING_H+self.VALUE_SPACING_H*(item_index-1)+self.text_pos
+	local xpos = xpos+self.OPTION_PADDING_H+self.ITEM_PADDING_H+self.ITEM_SPACING_H*(item_index-1)+self.text_pos
 	local ypos = ypos+self.OPTION_PADDING_T+self.OPTION_SPACING_V*(option_index-1)
 	
 	if KOPTOptions[option_index].text_dirty or refresh then
@@ -94,24 +94,35 @@ function KOPTConfig:drawOptionItem(xpos, ypos, option_index, item_index, text, f
 	end
 end
 
-function KOPTConfig:drawOptions(xpos, ypos, name_font, value_font, refresh)
+function KOPTConfig:drawOptions(xpos, ypos, name_font, item_font, refresh)
 	local width, height = fb.bb:getWidth()-2*self.MARGIN_HORISONTAL, self.HEIGHT
 	for i=1,#KOPTOptions do
 		self:drawOptionName(xpos, ypos, i, KOPTOptions[i].option_text, name_font, refresh)
 		for j=1,#KOPTOptions[i].items_text do
-			self:drawOptionItem(xpos, ypos, i, j, KOPTOptions[i].items_text[j], value_font, refresh)
+			self:drawOptionItem(xpos, ypos, i, j, KOPTOptions[i].items_text[j], item_font, refresh)
 		end
 		KOPTOptions[i].text_dirty = false
+	end
+end
+
+function KOPTConfig:makeDefault()
+	for i=1,#KOPTOptions do
+		KOPTOptions[i].text_dirty = true
+		for j=1,#KOPTOptions[i].items_text do
+			KOPTOptions[i].marker_dirty[j] = true
+		end
 	end
 end
 
 function KOPTConfig:config(callback, reader)
 	local kopt_callback = callback
 	local koptreader = reader
+	
+	self:makeDefault()
 	self:addAllCommands()
 	
 	local name_font = Font:getFace("tfont", self.OPT_NAME_FONT_SIZE)
-	local value_font = Font:getFace("cfont", self.OPT_VALUE_FONT_SIZE)
+	local item_font = Font:getFace("cfont", self.OPT_ITEM_FONT_SIZE)
 	
 	-- base window coordinates 
 	local width, height = fb.bb:getWidth()-2*self.MARGIN_HORISONTAL, self.HEIGHT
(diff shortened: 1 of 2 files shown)


eLiNK2gl:
great feature and implementation but unfortunately there're frequent framework reboots

P.S.: 2012-10

chrox:
@eLiNK2gl Could you please send me a link of the book that you are testing? And It will be more helpful if you tell me on which page you were reading and with what configuration you reflowed the page. Thanks.

eLiNK2gl:
@chrox the last time it was the following book -> http://szemke.math.univ.gda.pl/zemke2003PhD.pdf
Framework reboots happen on different pages, with different files.
Pointing out to memory handling problems (?).

Actually, since recently Windows reports a problem with some files on my DXG (while DXG is mounted).
So I suggest I sort this problem out first and then test the reflow again.

chrox:
@eLiNK2gl I tested the file you provided on both KPV emulator and a real Kindle device but I cannot repeat the bug you mentioned. On the emulator I tested every single page of the file and the reflow mode just worked fine. And on my kindle device I tested the first 50 pages and 10 random pages of that document and I didn't come across the problem you addressed above. I suspect that some files are corrupted on your kindle file system given the Windows reports you mentioned. Did you encounter similar problems when you use readers other than PDFReflow like PDFReader?

houqp:
Um, if it's reboot in DXG, then very likely, it's caused by memory leaks. DXG has only half memory of K3, so it could easily use up all the memory, when it happens, framework will try to reboot itself.

@eLiNK2gl did you test it with `Shift`+`P`+`K`? (i.e. stop the framework before launch kpv)

dpavlin:
I have also seen this on DXG, and it's not really a reboot -- Amazon framework restarts but reader is still started (you can check that by pressing space to refresh screen). I also suspect that it's memory related since we have about 20Mb on DXG with Amazon's framework running.

chrox:
@eLiNK2gl @houqp @dpavlin I SSHed to my Kindle 3 and checked free and used memory of the device with `free` command when KPV is running. Below is what I found.

```
Before launching KPV:
             total       used       free     shared    buffers     cached
Mem:        256536     228856      27680          0      45700      90864
-/+ buffers/cache:      92292     164244
Swap:            0          0          0

After launching KPV:
             total       used       free     shared    buffers     cached
Mem:        256536     233336      23200          0      47012      89820
-/+ buffers/cache:      96504     160032
Swap:            0          0          0

After opening zemke2003PhD.pdf in Reflow Mode:
             total       used       free     shared    buffers     cached
Mem:        256536     246044      10492          0      33148      85720
-/+ buffers/cache:     127176     129360
Swap:            0          0          0
```

Although the test was done on Kindle 3 we can assume that the situation will be similar on Kindle DXG. DXG only has 128 MB of RAM memory which will be run out immediately or after several page views in Reflow Mode. And It's not a result of memory leak at all because generally only 5 MB of memory is leaked out after 100 page views. The only plausible solution right now is making a swap file on DXG's internal mmcblock or external flash storage so that when the system runs out of RAM it can still allocate memory from the swap file with much slower Read/Write speed certainly.

tigran123:
A friend of mine (actually, I am proud to say my pupil, as it was me who introduced him to Linux about 12 years ago (we both were UnixWare hackers working for SCO at the time) and doing so was probably my biggest contribution to this planet's technology :) who is in charge of Linux VM (Hugh Dickins) raised the same issue ("should we swap on flash devices?") and if I remember correctly his conclusion was negative. It is for a good reason that Linux running on flash devices like Kindle comes with 0 swap space --- the architects of such devices are not idiots and realize the value of having a non-zero swap device, of course. I cannot remember the exact reasons, but my guess is that this would generate such a massive amount of writes to the flash device which will bring its lifetime to the end very quickly.

Btw, curiously `cat /proc/sys/vm/swappiness` shows 60 on Kindle whereas I would expect to see 0. Try setting it to 0 like this:

```
# sysctl vm.swappiness=0
```

and see if it makes a difference for running KPV.

Also, remember that we have `Shift P K` command to start KPV without Amazon's framework, i.e. kill it first. But I think this command is so dangerous that I never dare use it :) Oh, I see that @houqp already mentioned it...

chrox:
@tigran123 I agree with you. Maybe it's a bad idea to make a swap file on Kindle. This inspires me that we could optimize the reflow algorithm of  K2pdfopt for Kindle devices with very limited memory capacity. And how about to create a koptlib as a third party library of KPV so that we can remove the k2pdfopt.c in the main code tree.

tigran123:
@chrox how about a more radical approach --- to have a separate executable for each reader instead of just having a separate Lua object? I.e. to do what I did for PDF/DjVu (separated into a separate "br" executable in my "br" branch) for other formats. So, the filechooser would exec (overwriting its own address space, i.e. NOT fork+exec) the appropriate reader and the reader would exec again the filechooser on exit? Then we would have no issues with memory leaks as the entire address space of the process would be overwritten. And exec-ing is quite fast on Linux, so before we try this and see how it feels we shouldn't discard the idea as "too slow", because it might not be...

Also, this would mean that each reader is linked only with the libraries it needs. E.g. djvureader would probably be the smallest (apart from picviewer) as it needs neither mupdf(+thirdparty) nor crengine.

And we would be protected against potential memory leaks in crengine or any other complex module like koptreader.

What do you and others think?

hwhw:
I think this would not change much overall, though forking sub-processes
for the sake of stability might be in order. For memory usage, it wouldn't
change a thing (except for recovering leaking memory). The executable is
memory-mapped anyway.

Memory usage counts, especially from "free" tool, can be veeeery
misleading. In the numbers above, lots of memory is actually kind of "free"
- buffers and caches can always be dropped.

However, on the DX, there's not much space available when the framework is
running. So stopping the framework is the solution there.

tigran123:
Yes, I know that executables are memory mapped (btw, speaking of memory maps, that chap I mentioned actually _wrote_ the original mmap(2) system call in several Unix versions: OpenServer and UnixWare at least :)

I definitely agree that the output from free(1) is not the right tool to analyze the memory usage.

On the memory usage and stability --- recovering leaked memory AND also being more conservative wrt mapped libraries will make some difference. But I am NOT saying that this difference is enough to make it worthwhile re-designing the whole KPV to do this. Probably you are right, it will be far less significant than stopping the framework which is a "zero-cost solution", i.e. requires no new lines of code written :)

So, is stopping the framework a safe thing to do? I.e. one would not need a reboot to restart it properly?

hwhw:
I've never seen a framework restart fail, except the case where I mess up
the root filesystem while it's down :-)

tigran123:
Also, what are the reasons for using statically linked libraries for [almost] everything? We could save some space in the text (and data as well) segment if we switch to shared libraries. E.g. if the user never reads djvu files why keep all the text/data of libdjvulibre? And vice versa.

Btw, I am about to commit a more proper memory usage indicator for KPV which should help us monitor memory usage without the need to ssh to Kindle.

hwhw:
No good reasons anymore - it was once a measure to keep overall package
size down, back when it was a simple frontend to mupdf. The linker can
optimize according to actually used functions, so it kept it down a bit.

As I've often thought about converting the C frontends to the backend
libraries to use LuaJIT's FFI (veeery interesting stuff and it really
boosts performance!), it would be most welcome to keep the backends in
separate libraries. Preferably even mostly unmodified/unpatched.

A memory usage indicator would indeed be great. I think we could plug into
the memory allocation functions for most of the backends already, and doing
it with Lua(JIT) should also work (it might even have an interface already
to query used memory).

eLiNK2gl:
@chrox Sorry for coming back only now.
I'm rather pressed for time these days.
Essentially, you circled in on the problem.
While starting the KPV with shift + p + d, it takes me a minute at the longest to make the framework restart.
@houqp while starting the KPV with shift + p + k, I have not managed to get a framework restart so far (but I had not much time).

PR #497, opened 2012-10-27
save koptconfig in doc settings and disable precache in koptconfig

chrox:


dpavlin:
@chrox I played around with reflow on DXG today and it works very good on two column articles... I think we could recommend it as valid alternative to two column mode. Thanks again for your efforts in getting this integrated. Next version of kpv will be a blast!

chrox:
@dpavlin I'm really glad to hear that. And all the configurable options in k2pdfopt that I can think of have been added to koptconfig. So we only need more tests and bugfixes if any before the release of 2012.10. Btw, what's your plan of this new release?

dpavlin:
Since it's mostly adding tag and final testing we can do that whenever we are ready. We could even add tag tomorrow and than wait just announce new version when we feel it's solid.

chrox:
@dpavlin Thanks.

PR #498, opened 2012-10-27
Implements "comics mode"

kai771:
This is bacically my hack no.3, described here: http://www.mobileread.com/forums/showpost.php?p=2215362&postcount=618, but a bit improved, and implemented like a toggle. It's very useful for reading comics, so thus I called it comics mode. 

When I read pdfs, I usually manually set bbox, and then use shift-s. Without this PR, it usually shows a pdf page on 2 screens: on 1st screen the top of the page is lined up with top of the screen. Then, when I press pg_fw, it displays second part of the page (which is usually 3 or 4 lines) at the top of the screen, with the rest of the screen gray. With this PR, when you press pg_fw it will display the 2nd screen with bottom of the page lined up with the bottom of the screen. The whole screen will be filled with the content of the page (most of it was on the 1st screen too). Of course, one can do this with panning, but I find reading it like this much more natural.

Shortcut key for comics mode is C.

### DIFF
--- a/djvureader.lua
+++ b/djvureader.lua
@@ -5,6 +5,7 @@ DJVUReader = UniReader:new{}
 function DJVUReader:setDefaults()
 	self.show_overlap_enable = true
 	self.show_links_enable = false
+	self.comics_mode_enable = false
 end
 
 -- check DjVu magic string to validate
--- a/koptreader.lua
+++ b/koptreader.lua
@@ -325,6 +325,7 @@ end
 function KOPTReader:adjustCommands()
 	self.commands:del(KEY_A, nil,"A")
 	self.commands:del(KEY_A, MOD_SHIFT, "A")
+	self.commands:del(KEY_C, nil,"C")
 	self.commands:del(KEY_D, nil,"D")
 	self.commands:del(KEY_D, MOD_SHIFT, "D")
 	self.commands:del(KEY_S, nil,"S")
--- a/koptreader.lua
+++ b/koptreader.lua
@@ -295,6 +295,7 @@ end
 function KOPTReader:setDefaults()
     self.show_overlap_enable = true
     self.show_links_enable = false
+    self.comics_mode_enable = false
 end
 
 -- backup global variables from UniReader
--- a/picviewer.lua
+++ b/picviewer.lua
@@ -5,6 +5,7 @@ PICViewer = UniReader:new{}
 function PICViewer:setDefaults()
 	self.show_overlap_enable = false
 	self.show_links_enable = false
+  self.comics_mode_enable = false
 end
 
 function PICViewer:open(filename)
--- a/unireader.lua
+++ b/unireader.lua
@@ -1419,10 +1419,12 @@ function UniReader:show(no)
 	fb.bb:blitFrom(bb, self.dest_x, self.dest_y, offset_x, offset_y, width, height)
 
 	Debug("self.show_overlap", self.show_overlap)
-	if self.show_overlap < 0 and self.show_overlap_enable and not self.comics_mode_enable then
-		fb.bb:dimRect(0,0, width, self.dest_y - self.show_overlap)
-	elseif self.show_overlap > 0 and self.show_overlap_enable and not self.comics_mode_enable then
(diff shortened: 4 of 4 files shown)


tigran123:
@kai771 have you tested it? Although I have not tested it, I see that you have modified UniReader:setDefaults() without modifying all the other :setDefaults() which is a bug --- i.e. without even looking into the details of your patch, just treating it as a "black box" I would say that changing UniReader:setDefaults() implies that you have to change the other readers' :setDefaults() appropriately as well.

kai771:
@tigran123 I did test it with a few cbz and pdf files, and it appeared to be fine. I didn't know about :setDefaults for other readers, sorry. The patch is meant for pdf/cbz mostly. I'll take another look. I'd appreciate if you find the time to test it yourself too. Any suggestions greatly appriciated.

tigran123:
@kai771 Yes, as I suspected it has a bug. Namely, if you press `C` while reading one DjVu file and then close the file and open another DjVu file (with no prior history) then it will also be in the comics mode. So, you need to do this:

```
 function DJVUReader:setDefaults()
        self.show_overlap_enable = true
        self.show_links_enable = false
+       self.comics_mode_enable = false
 end
```

And likewise to the functions `PICViewer:setDefaults()` and `KOPTReader:setDefaults()`. Also, for KOPTReader this command makes no sense, so you have to delete it explicitly, like this:

```
 function KOPTReader:adjustCommands()
        self.commands:del(KEY_A, nil,"A")
+       self.commands:del(KEY_C, nil,"C")
        self.commands:del(KEY_A, MOD_SHIFT, "A")
        self.commands:del(KEY_D, nil,"D")
        self.commands:del(KEY_D, MOD_SHIFT, "D")
```

But for DJVUReader and PICViewer your new comics mode may come in handy, so you shouldn't delete the `C` command from these two.

Also, this code:

```
      if unireader.comics_mode_enable then
            InfoMessage:inform("Comics mode ON", nil, 1, MSG_AUX)
       else
            InfoMessage:inform("Comics mode OFF", nil, 1, MSG_AUX)
       end
```

may be more elegantly written on a single line:

```
InfoMessage:inform("Comics mode "..(unireader.comics_mode_enable and "ON" or "OFF"), nil, 1, MSG_AUX)
```

But this last item is just a suggestion, feel free to ignore it if you prefer the longer version.

kai771:
Fixed :setDefaults in picviewer.lua, koptreader.lua and djvureader.lua. pdfreader.lua and crereader.lua didn't have :setDefaults. I'll delete C command from KOPT reader next. As for elegance, I'll leave it as it is for now, because that's the way it was done with other commands. I'll leave it to you to clean it up when you clean up every other occurence (for example for overlap, links, etc. :) )

tigran123:
Also, we should try to do as little work as possible in the :show() method, so instead of doing four conditional checks on each call we can do just two, like this:

```
@@ -1419,10 +1419,12 @@ function UniReader:show(no)
        fb.bb:blitFrom(bb, self.dest_x, self.dest_y, offset_x, offset_y, width, height)

        Debug("self.show_overlap", self.show_overlap)
-       if self.show_overlap < 0 and self.show_overlap_enable and not self.comics_mode_enable then
-               fb.bb:dimRect(0,0, width, self.dest_y - self.show_overlap)
-       elseif self.show_overlap > 0 and self.show_overlap_enable and not self.comics_mode_enable then
-               fb.bb:dimRect(0,self.dest_y + height - self.show_overlap, width, self.show_overlap)
+       if self.show_overlap_enable and not self.comics_mode_enable then
+               if self.show_overlap < 0 then
+                       fb.bb:dimRect(0,0, width, self.dest_y - self.show_overlap)
+               elseif self.show_overlap > 0 then
+                       fb.bb:dimRect(0,self.dest_y + height - self.show_overlap, width, self.show_overlap)
+               end
        end
        self.show_overlap = 0
```

before your change it was also not-optimal, i.e. doing two conditional checks instead of one, but you added two more, making the bad thing worse, so please fix it as suggested above.

tigran123:
Thank you. I see no reason not to merge it now...

PR #500, opened 2012-10-28
Fixes a bug with tall, 1 page pdf files.

kai771:
How to reproduce the bug:
Open tall.pdf in PDFViewer, press shift-s. Press pg_fw three times. Press pg_bk. It won't work.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1599,7 +1599,6 @@ end
 
 function UniReader:nextView()
 	local pageno = self.pageno
-	local numpages = self.doc:getPages()
 
 	Debug("nextView last_globalzoom_mode=", self.last_globalzoom_mode, " globalzoom_mode=", self.globalzoom_mode)
 
@@ -1607,7 +1606,7 @@ function UniReader:nextView()
 	or self.last_globalzoom_mode == self.ZOOM_FIT_TO_CONTENT_WIDTH_PAN then
 		if self.offset_y <= self.min_offset_y then
 			-- hit content bottom, turn to next page
-			if pageno < numpages then
+			if pageno < self.doc:getPages() then
 				self.globalzoom_mode = self.ZOOM_FIT_TO_CONTENT_WIDTH
 			end
 			pageno = pageno + 1
--- a/unireader.lua
+++ b/unireader.lua
@@ -1599,6 +1599,7 @@ end
 
 function UniReader:nextView()
 	local pageno = self.pageno
+	local numpages = self.doc:getPages()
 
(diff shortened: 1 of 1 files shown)


tigran123:
Thank you, very interesting. I will test this. It definitely needs to be tested very thoroughly to make sure it doesn't break the multi-column mode(s).

tigran123:
Ok, I've tested it and it does not break anything. However, let's address two things before we merge it:
1. I don't understand why the code `pageno = pageno + 1` is not _inside_ the check `if pageno < numpages`. I am not saying that it should be, but I don't understand why it should not. I.e. what is the point of returning a page number greater than the total number of pages in a document? And similiarly in `prevView()` for the code `pageno = pageno-1` and condition `if pageno > 1`. Please explain. I would like to understand this before merging it to master. Btw, I see that highlighting doesn't work in two-column mode properly, neither with nor without your fix. Namely, when highlighting a word on the last line of the current view and moving down it doesn't scroll to the next view correctly --- instead it jumps to the next page number.
2. You only need to know the number of pages in `nextView()` under certain conditions and you only use it once. Therefore, calculating it unconditionally on _every_ call to `nextView()` is unwise, please change it to:

```
 function UniReader:nextView()
        local pageno = self.pageno
-       local numpages = self.doc:getPages()

        Debug("nextView last_globalzoom_mode=", self.last_globalzoom_mode, " globalzoom_mode=", self.globalzoom_mode)

@@ -1607,7 +1606,7 @@ function UniReader:nextView()
        or self.last_globalzoom_mode == self.ZOOM_FIT_TO_CONTENT_WIDTH_PAN then
                if self.offset_y <= self.min_offset_y then
                        -- hit content bottom, turn to next page
-                       if pageno < numpages then
+                       if pageno < self.doc:getPages() then
                                self.globalzoom_mode = self.ZOOM_FIT_TO_CONTENT_WIDTH
```

tigran123:
Concerning my question 1. I see that putting those page increments/decrements inside the condition is NOT the right thing to do because it causes an awful lot of unnecessary refreshes then. But I still don't understand why.

kai771:
@tigran123 
No.1: if I understood it correctly, point 1 is handled here:

``` lua
function UniReader:goto(no, is_ignore_jump)
    local numpages = self.doc:getPages()
    if no < 1 or no > numpages then
        return
    end
```

No.2: Good point. Fixed.

tigran123:
Ah, yes, I didn't see the call to goto() in the command handler :)
Ok, that makes sense, although it is a bit non-intuitive, i.e. we return the wrong page number so that goto() does nothing. If we returned the correct (current) page number then goto() would cause a lot of unnecessary refreshes....
But that is the way it is done, so let's not change it. Merging the patch now...

PR #507, opened 2012-10-29
increase margin to 50px and correctly calulate zoom

dpavlin:
This is important to correctly position page with bbox on screen and leave margin. If page doesn't have bbox we just ignore margins because we can't shift page using offset in current code anyway and it would just incorrectly shift page

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -63,7 +63,7 @@ UniReader = {
 	pan_by_page = false, -- using shift_[xy] or width/height
 	pan_x = 0, -- top-left offset of page when pan activated
 	pan_y = 0,
-	pan_margin = 5, -- horizontal margin for two-column zoom (in pixels)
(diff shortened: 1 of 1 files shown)


dpavlin:
`F` zoom mode was designed to be used with two-column documents with bbox, and without margins it would get stuck to screen border which makes it really hard to read. Over the time, it got decreased to 5px which was hiding the fact that we calculated zoom using margin pixels and page points which is very wrong.

dpavlin:
Don't merge this just yet. It would probably make sense to calculate margin size based on screen size to have smaller margin on K3 and bigger on DXG, but I need to test it on device to figure out optimal default size.

houqp:
Yo, your commit reminds me that we can easily add margin support for all modes by playing with bbox! This is the feature that people have been asking for long ;p

dpavlin:
This changes are part of #515

PR #510, opened 2012-10-31
Simple Memory Usage Indicator

tigran123:
The function `util.c:memusage()` returns three numbers like this:

```
rss, data, totalvm = util.memusage()
```

The sizes are in kilobytes and obtained from `/proc/self/status` file. Tested both on the emulator and Kindle.

I shortened the date/time format to just "%H:%M" --- if someone really needs to see the full date/time then he can enter `os.date()` in the calculator :)

### DIFF
--- a/djvureader.lua
+++ b/djvureader.lua
@@ -114,14 +114,16 @@ function DJVUReader:_drawReadingInfo()
 	local load_percent = self.pageno/numpages
 	local face = Font:getFace("rifont", 20)
 	local page_width, page_height, page_dpi, page_gamma, page_type = self.doc:getPageInfo(self.pageno)
+	local rss, data, totalvm = util.memusage()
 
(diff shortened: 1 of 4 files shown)


tigran123:
Oh wait, don't merge yet --- it can be simplified still!

tigran123:
Ok, I think it is polished enough now.

dpavlin:
Probably a stupid question, but do we really want to implement /proc/self/status parser in C instead of lua?

tigran123:
Hmmm, I completely forgot that there is such language called Lua :))))

tigran123:
Ah, I remember why I did it in C --- I didn't know how to call getpid(2) system call in Lua because originally I wanted to open the file /proc/PID/status, but then I remembered about the "/proc/self/status" name.

Ok, I'll redo it in Lua now. Closing this PR now as the Lua version will be in a different PR.

Issue #453, opened 2012-10-16
Request : column zooming function for manga


melib:
Could you enhance "zoom to first column function ('F' and 'shift+F' keys)" in reader mode to be able to start at "a top right corner" or "a top left corner"?

So that I can choose "start at top right corner" when reading two-page manga and "start at top left corner" when reading two-column journal.

Two-page manga, for example, http://www.mangaforkids.com/images/howto-spread.jpg

Thank you.;)

and sorry for my English.

dpavlin:
Do you have any manga document with more than one page? From my limited knowledge about Japanese I would also assume that you start reading on last page and move toward first one (from my western point of view, that is ;-)

Current code have some corner cases where margins won't work well (pdf files from Linux Journal are example for me) so I was thinking about fixing that, so I will also have in mind this feature request when I start working on it.

tigran123:
@dpavlin Yes, that would be syntactically correct for Hebrew (and other semitic languages like Syriac, Aramaic, Arabic, etc) but I wouldn't advise to mix this in with the right-to-left column switching mode. The reason is because many books have right-to-left columns but when they are provided in PDF or DjVu formats they actually start from the physical page=1 and not from page=lastpage. This is for the "user's convenience". So, I suggest keeping the right-to-left column switching and right-to-left book binding separate.

(I don't know about Japanese "manga" documents, but a Google search reveals that yes, those are read from right to left and start at the end of the document, not the beginning, but this, imho, is of minor relevance compared to the above argument)

tigran123:
I would also like to add that PDF specification provides an object which tells which direction (R2L or normal) it is typeset in. My typesetting system actually makes use of it (for right-to-left semitic languages), but most viewers ignore it anyway...

melib:
@dpavlin This is a sample. http://www.mediafire.com/download.php?vwcw3d3gy59mhfl 
Each page is merged from two pages and Texts is too small for reading even reads in landscape mode.

Page ordering is not problem for reading manga in kindle because It reads normally from the first page to the last page.
But it reads from right to left in each page so if  the zoom function can start at the top right corner, It would be awesome.   

Thank you very much.

kai771:
Now I see this was discussed here, so I shouldn't have opened #515. I'm no expert (far from it), but I'd like to add a few comments about japanese "start reading on last page and move toward first one" and clarify a possible misunderstanding.

[...]

PR #515, opened 2012-10-31
implements rtl_mode

kai771:
Implements rtl_mode (rtl=right-to-left). Although I meant it mainly for reading manga, together with comics_mode, I guess it might be useful for other things too. Shortcut key is R. It’s only enabled in pdfreader and djvureader. It works in zoom modes fit-to-page-height, fit-to-content-height and 2 column mode.

This is the initial version, and possibly full of bugs. Please test thoroughly.

It also fixes a bug that was pressent in 2 column mode (at least I think it was a bug). In key_fw_left handling, there was a line:

```
if unireader.offset_x > 0 and unireader.pageno > 1 then
```

On the files I tested it with, unireader.offset_x was never 0 or less. So I changed this line to

```
if unireader.offset_x > unireader.pan_margin and unireader.pageno > 1 then
```

It appears to work for me, but I’m not really sure if it was the right thing to do, so @dpavlin, if you can please find the time to go through the code, I’d be most grateful.


tigran123:
You can't use `R` key because it is already used in djvureader for setting the rendering mode.

Also, would you mind explaining what does it actually do? The phrase "right to left" can mean a few different things in typography...

kai771:
Sorry about `R`, I didn't check djvureader (I don't have many djvu files). I'll change it later. 

In Japanese, you read manga right-to-left. The panels on the right are read first, then the ones on the left. So, when using fit-to-page-height or fit-to-content-height together with comics_mode, it will line up the page to top-right.

When using 2 column mode, if rtl_mode is enabled, it will first show the right column's top first. Then you press fw_down, then fw_left. It will move to left column's top. Further fw_down, then fw_left will move to next page. It's basicly a reversal of order of initial 2 column mode, useful for reading things that are read right to left.

It was requested here:
http://www.mobileread.com/forums/showpost.php?p=2266332&postcount=756

And now I see it was discussed in #453 too.

kai771:
I changed shortcut R to Alt-R. Btw, in crereader.lua and picviewer.lua I already found deletions of Alt-R, and I couldn't find adding Alt-R anywhere. Am I missing something?

kai771:
And now I found #475... so, shortcut changed again... this time, from Alt-R to U. Why U? It was free (hopefully... I'm not too sure anymore), it's only one key, and can stand for... er... unusual :).

kai771:
Hm, it seems I forgot to test with bbox defined. It has some bugs with it.

I manually added changes from #507 to my local rtl_mode branch, since they seemed needed to sort it all out. I don't know if it will cause problems for merging later on. If necessary, I will rewrite the patch line by line and submit again. Let's see if I can sort it out first, and worry about that later.

kai771:
rtl_mode should work properly now with BBox. I still don't fully understand the concept of margin as used here, so it might not work as intended.  I'll take another look tomorrow. It might work better with pan_margin set to 5, so please check that too. This PR incorporates dpavlin's changes from #507. I removed "+ margin" from the following line:

```
self.offset_y = -1 * y0 * self.globalzoom * 2 + margin
```

because in the comments for pan_margin it said that it's for horizontal margin, so I assumed this is a typo. Please correct me if I'm wrong.

Bug fixes include proper handling of first/last page and pressing fw_left/fw_right, and rounding down pwidth and pheight to 2 decimals, because page:getUsedBBox() returns values with 2 decimal places, and without it check whether to use margins on some documents might fail.

There might be some bugs left, but it seems much better now. Please test and comment.

tigran123:
@kai771 I have not tested this yet, but as it touches one of the two or three most fundamental and valuable features of KPV, namely the ability to read in two-column mode, I think it should be tested extremely thoroughly. Not just your new functionality but (more importantly, imho, because the number of two-column Hebrew, Syriac and Arabic books is not that great (though I have a few thousands in my collection) compared to English/Russian/etc ones) to ensure that the existing functionality is not broken.

kai771:
I totally agree. Please test as much as you can. I implemented new functionality independently from old one, so it shouldn't break anything, but I did some bug fixes to old functionality, so it needs to be tested thoroughly.

tigran123:
Ok, I will do that as soon as I finish a few other bits I am working on. Don't worry about the delay --- your new rtl mode is a valuable contribution and will definitely be "attended to" :)

kai771:
Not worried at all, just eager to have some feedback so I can fix it if it's broken. I tried to test it as much as I can, but I might be missing something, so tests from others are most welcome. I don't have any right-to-left column texts, (I mainly needed this for manga), but I tried to properly implement it in case someone else needs it.

tigran123:
Here is a right-to-left Bible in Syriac:

http://www.klib.8tar.com/Books/Syriac/Syriac-OT-Lee-1823.djvu

And here the text is not centered on a page (i.e. odd and even pages differ) so one must define a proper bbox to read in two-column mode. This makes a good example for testing.

kai771:
Thank you, I'll test it later today when I get home.

tigran123:
Actually, you may prefer this one (modern Syriac version):

http://www.klib.8tar.com/Books/Syriac/Syriac-Modern-Bible.djvu

because its verse numbers are printed more clearly in the margin, whereas in Lee's 1823 edition they are barely visible.

kai771:
While you're at it, can you point me to some 2 column djvu files in english too? You don't really need to use actuall right-to-left text to test rtl mode, right? ;)

dpavlin:
@kai771 we need vertical margin in `self.offset_y = -1 * y0 * self.globalzoom * 2 + margin` because without it page is stuck to top of screen which makes first line hard to read.

kai771:
No problem, I'll put it back. Can you explain the use of margins a bit better? What's the benefit of using margins to just leaving some space when making bounding box with shift-x? As I said, I'm not really clear on margins (I'll take another look when I get home). In my understanding, the margin should be the amount of pixels from the edges of the screen. That area should be skipped when doing redraw. But, the way it is now, redraw always bitblits with witdh 600, if I remember correctly (I don't have access to the code right now). It seems that margin just moves the point from which the bitblit starts (which in my opinion is the same as just leaving some space when setting bbox). I probably misunderstood something, so please explain.

Btw, I think we should have different values for each margin. Just my opinion, though.

dpavlin:
Idea is to use margins when pdf already has bbox information (as it's usual in scientific papers). However, in that case bbox is (usually) exact size of printed text so it doesn't leave any margins to the edge of the screen which makes it hard to read (mostly because screen boarder on Kindle drops shadow on screen). Margins don't make much sense with `shift+x`, expect if you manage to do it exactly to edge of font (which is hard because we have fixed step size). When margins are not needed, you can use `shift+f` to disable them.

In current code, margins modify origin points in source bitmap (from mupdf/djvu) which is reason why they don't work well when bbox is same size as page (we can't move more to left/up since we don't have bitmap there). We are always bitbliting screen width to cover "old" screen content.

Does this make (some) sense? ;-)

I don't have preference about separate horizontal/vertical margins. I didn't need them, but we should probably first add user interface to change them, I think.

tigran123:
Here are some two column PDF files:

https://github.com/downloads/tigran123/russian-synodal-bible/Russian-Synodal-Bible-USletter-twocolumn.pdf

http://urantia-book.org/downloads/English/Urantia-A4-twocolumn.zip

kai771:
@dpavlin Yes, it does, thank you. I thought it was something like that, but I needed confirmation. The trouble is, a lot of pdf files have defined bboxes, but incorrectly. So I'd have to manually set it. Then, margins are of no use. I had some issues with margins set to 50, so if you don't mind, I'll put them back to 5 for now. We can sort out margin problems in the separate PR. I'll push my patch shortly (after I do some more testing)

@tigran123 Thank you. I just got home, so it's just in time :).

dpavlin:
Setting them to 5 make sense, I increased it to 50 just to have them more visible and that's how I found out problem with them (they are already user-configurable in .lua file anyway)

tigran123:
Hmmm, I don't understand what you are talking about wrt margins but surely they are always set to 0 (not 5px and certainly not 50px) i.e. it is entirely up to the user to decide the exact dimensions of the bbox.

Ah, perhaps you are talking about the situation when bbox is not yet defined by the user? Hmmm, I think in such situations the books are never quite readable and the user MUST define the exact bbox and do so twice, i.e. for odd and even pages separately. Only after this procedure the book is readable...

tigran123:
@kai771 I promised to help with testing but I only just now came back home and my eyes are very tired. I need to give them some rest by reading :) I might have time to do this testing tomorrow but I think it would be wise of me not to promise anything (but I will try).

kai771:
There, I restored `self.offset_y = -1 * y0 * self.globalzoom * 2 + margin` line and set pan_margin to 5. I took a quick look through tigran's files, and it all appears to work correctly. (It's kinda hard to check since I don't read any of the languages). I'll make and upload customupdate.zip, and invite people on mobileread.com to test it. Hopefully all is good :).

@tigran123 Np, it's not like it's urgent. :)

kai771:
For the people reading this, if you'd like to test it, you can download the binaries here:
https://github.com/kai771/kindlepdfviewer/downloads

tigran123:
@kai771 the "Urantia*" file is in English :)

kai771:
@tigran123 Yes, and as far as I can tell, it works just fine.

@dpavlin WRT my previous comment about needing separate horizontal/vertical margins - scratch that. I misunderstood what they're for. There's no need to have them separate. Also, I thought you wanted pan_margin to default to 50, and that it wasn't just for testing purposes. The problems I had with them was that the space outside of incorrect bbox was smaller than 50 pixels, so it resulted in gray borders on just right columns on some documents. If they're only 5, that's not an issue.

tigran123:
@kai771 I decided to take a look at it now. First comment: do NOT delete the "U" command from picviewer. The main point of writing picviewer was to browse the scans of ancient Hebrew manuscripts --- most of these are in two columns and all in right to left mode. So, the rtl mode can be very useful in picviewer. (I am talking about colour scans for which making a DjVu file has no advantage compared to keeping the raw jpeg images).

kai771:
@tigran123 Ah, no problem. I'll put it right back :). With regards to your picviewer - you might want to consider adding pg_fw and pg_bk to display next/previous picture in the directory, like most picture viewers on a PC do. Can be useful for watching albums (or reading decompressed manga :))  Btw, I had no idea that two columns mode works in picviewer. Never even thought to check :).

tigran123:
@kai771 Yes, I thought about that. I think that the right page Next/Prev keys should be left alone, i.e. behave as in normal books (because you could treat an image as a single-page book and would expect all the functionality to be usable). However, the LEFT Next/Prev keys could be mapped to going through the images in the current directory as you describe, yes.

kai771:
Awwww, I prefer to use it with right hand. What usability would be there for one page book not to respond to page turn buttons? I think both left and right should work.

tigran123:
In `Shift-S` mode pressing page Next/Prev keys does not turn the page but shifts the view down. So you can imagine a huge tall scan where the user defined a bbox (to surround the text) and then entered `Shift-S` mode to go through the scan from top to bottom pressing page `Next`. I suppose he could achieve the same by pressing fiveway Down key and defining big enough vertical shift step so this functionality is not worth preserving. Maybe you are right --- both left and right page next/prev could be mapped to going through the files in the current directory.

tigran123:
So far only one strange thing noticed --- when in 2-column mode going to the next page it seems we have too many screen refreshes. But on the emulator it is hard to tell because everything is so fast. I'll test it on the real Kindle now where it should be more noticeable (from debug messages it is not obvious)

kai771:
I see your point, and it is a valid one for that kind of situation. However, I think that few people would use it for that (e.g. reading long scrolls, more than ten screens high), and vast majority would _expect_ that next/prev would give them the next/prev picture, since most pic viewers work that way.

tigran123:
Ok, it seems I cannot break the new rtl functionality and I cannot detect any breakage of the old existing functionality.
Those extra page flickers in the emulator were probably caused by page_indicator() calls in debug mode as I cannot see any problem when running on Kindle. Maybe we'll test it a bit more and if all is good then merge tomorrow, ok?

kai771:
Of course. It's really reasuring that _you_ couldn't break it (so far), since you're quite good at finding flaws. I hope it stays that way :).

PR #516, opened 2012-11-01
fix makefile for building libdjvu in emu mode

houqp:


tigran123:
Hmmm, the directory is actually created in the "cleanthirdparty" target. Why would one want to "make thirdparty" without having run at least once "make cleanthirdparty"? But, I suppose, there is no harm in creating it in the $(DJVULIBS) target as well.

PR #517, opened 2012-11-01
merge latest changes from master branch

houqp:
That's a long trip and kind of painful...

Anyway, I can now get back to the fun part ;P


dpavlin:
Blame @tigran123 and @chrox they where very busy committing code and making reader better :-)

houqp:
LOL, then you should also be included for all these link support ;p

PR #518, opened 2012-11-01
Add simple memory usage indicator.

tigran123:
The function memUsage() parses the file "/proc/self/status" and extracts the sizes (in KB) of RSS, Data, Stack, Lib and TotalVM. These are displayed when Menu button is pressed inside readers.

(Also, the info displayed by crereader is tidied up, e.g. there is no need for "Pos: " prefix before showing percentage as it is obvious and just wastes space that can otherwise be used for the section title. Also, show the page numbers and total number of pages even though these are "virtual".)

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -384,22 +384,31 @@ end
 ----------------------------------------------------
 -- used in CREReader:showMenu()
 function CREReader:_drawReadingInfo()
-	local ypos = G_height - 50
+	local width = G_width
 	local load_percent = self.percent/100
+	local rss, data, stack, lib, totalvm = memUsage()
+	local face = Font:getFace("rifont", 20)
 
-	fb.bb:paintRect(0, ypos, G_width, 50, 0)
+	-- display page number, date and memory stats at the top
+	fb.bb:paintRect(0, 0, width, 40+6*2, 0)
+	renderUtf8Text(fb.bb, 10, 15+6, face, "p."..self.pageno.."/"..self.doc:getPages(), true)
+	local txt = os.date("%a %d %b %Y %T").." ["..BatteryLevel().."]"
+	local w = sizeUtf8Text(0, width, face, txt, true).x
+	renderUtf8Text(fb.bb, width - w - 10, 15+6, face, txt, true)
+	renderUtf8Text(fb.bb, 10, 15+6+22, face,
+		"RSS:"..rss.." DAT:"..data.." STK:"..stack.." LIB:"..lib.." TOT:"..totalvm.."k", true)
+
+	-- display reading progress at the bottom
+	local ypos = G_height - 50
+	fb.bb:paintRect(0, ypos, width, 50, 0)
 
 	ypos = ypos + 15
-	local face = Font:getFace("rifont", 20)
 
 	local cur_section = self:getTocTitleOfCurrentPage()
 	if cur_section ~= "" then
-		cur_section = "  Section: "..cur_section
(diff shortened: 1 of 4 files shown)


tigran123:
Hmmm, using this indicator I think I already found a bug --- if you keep your finger on Menu button and in another terminal monitor the file descriptor table you will see that it keeps growing.

It may be a bug in Lua (or LuaJIT) whereby the descriptor is not closed if you break out of io.lines(filename) loop. By the manual it is supposed to be closed.... I'll verify this now, but we definitely have a file descriptor leak somewhere....

Update: it is not a bug in Lua or LuaJIT and neither it is a bug in my memUsage() function --- the descriptor is closed correctly when you break out of the io.lines(filename) loop (tested with both normal /usr/bin/lua and luajit). The bug is in KPV application somewhere, i.e. when we press Menu we open some file which we forget to close. I don't know what it is. Don't merge this PR yet as I want to get to the bottom of what is causing this file descriptor leak --- very intriguing...

tigran123:
This file descriptor leak turned out to be a very interesting mystery, now solved :)

As I originally suspected (before any testing) it is due to the descriptor not being closed when breaking out of io.lines(filename) loop. However, then I tested in a small standalone test program and everything was fine, i.e. the descriptor was closed correctly. However, for "large" programs (presumably in the sense of having many other open file descriptors) the behaviour of Lua (or LuaJIT) garbage collector is different --- i.e. for such the descriptor is NOT closed. Its closure is delayed for "later" and this "later" never arrives so the file descriptor table grows infinitely (well, at least until 1024 entries as I didn't have patience to keep my finger for more than 5 minutes on the Menu key :)

This very strange Lua quirk is worth remembering as it is not documented anywhere, as far as I know.

The solution is to NOT use io.lines() loop because breaking out of it is unpredictable. I'll rewrite memUsage() function to use the ordinary file:lines() with the manual file:close() way of doing things. Hopefully this will work, but I'll test it first, of course.

tigran123:
Yes, it does work, so I added the commit above.

Now, using this indicator I have found another bug --- this time in crengine. So, don't commit it just yet while I am making sure that it is not the new memUsage() function that is causing. The bug is this: KPV crashes when you open a chm file and press Menu. Hmmm, maybe chm files have no concept of page numbers at all? I'll check now.

Update: No, the chm file is crashing (sig11 coredump) because of the call to `self:getTocTitleOfCurrentPage()` regardless of my changes in this PR. So this is just yet another crengine bug. And if I comment out a call to this function then Menu is displayed fine but it will crash in a matter of a few more page views anyway... (crengine's support for chm format or any format other than FB2 is not usable).

Btw, the file in question is the Lua 5.1 reference manual (in chm format). It would crash also if one just presses `T` to get the TOC.

houqp:
I am also reading a chm book right now (I hate this format, but this is the only format it comes with ;p), and everything works just fine. Probably we should open a new issue for it?

BTW: is this ready for merge?

tigran123:
Yes, as far as I am concerned it is ready. I'll open a new issue for the chm crash.

tigran123:
I have just downloaded the very latest Lua 5.1.1 (as opposed to "5.1") chm Reference manual from here:

http://board.ptokax.ch/index.php?topic=6543.0

and it does NOT crash anymore. So, maybe my 5.1 chm file was corrupt in some way...
Should we still open an issue and investigate?

houqp:
Yep, kpv should be able to handle corrupted chm anyway ;p

PR #524, opened 2012-11-02
Link against libraries in libs and libs-emu

tigran123:
Instead of linking against the location inside the original build tree we should link against the copy in a single place. This makes the build code cleaner and more compact even now and will do even more so when we build more libraries as shared (which is what I am going to attempt next).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,9 +75,11 @@ ifdef EMULATE_READER
 	endif
 	CFLAGS+= $(HOST_ARCH)
 	CXXFLAGS+= $(HOST_ARCH)
+	LIBDIR=libs-emu
 else
 	CFLAGS+= $(ARM_ARCH) $(ARM_BACKWARD_COMPAT_CFLAGS)
 	CXXFLAGS+= $(ARM_ARCH) $(ARM_BACKWARD_COMPAT_CFLAGS) $(ARM_BACKWARD_COMPAT_CXXFLAGS)
+	LIBDIR=libs
 endif
 
 # standard includes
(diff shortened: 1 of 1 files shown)


dpavlin:
Why did you felt need to have separate directory for emulated libraries as opposed to cross compiled? We can't have emulated and cross-compiled binaries in same tree anyway.

tigran123:
@dpavlin Consider the following sequence of steps:
1. Build kpdfview in EMU mode and rename it to kpdfview.emulator. Edit reader.lua to point to it and do some development.
2. When happy with the result unset the EMU\* environment variables and build the whole thing for ARM. Remember to edit reader.lua to point to kpdfview again. Ship the result to Downloads section.
3. Continue development in EMU mode. Now, if the libs and libs-emu were the same directory "libs" then I would have to rebuild "make thirdparty" again to populate it with i686 libraries, otherwise kpdfview.emulator binary wouldn't run because of djvulibre. But as they are separate I can run kpdfview.emulator with LD_LIBRARY_PATH pointing to libs-emu. (Of course the host system also has a djvulibre library installed but it is NOT a good idea to link against it because it may be a different version than intended, i.e. the one whose headers were used while compiling kpdfview.emulator binary).

houqp:
yeah that makes life much more easier. you don't even need to edit `reader.lua`, just write a script that runs: `./kpdfview.emulator reader.lua` ;p

dpavlin:
Yap, this make sense. I'm currently using two checkouts, one for emulation and one for cross-compile.

PR #527, opened 2012-11-03
more flexible configuration dialog

chrox:
Each option can be configured to show or not and the height of the config dialog is updated accordingly. By default the page margin and line spacing is not drawn to save dialog space. Because these two options just make minorest differences to the view of reflowed pages and the default medium values should work in most case. Even so users can still change the visibility of these two options by editing the `KOPTOptions` table.

### DIFF
--- a/k2pdfopt.c
+++ b/k2pdfopt.c
@@ -498,7 +498,8 @@ void k2pdfopt_set_params(int bb_width, int bb_height, \
 		int wrapping, int straighten, \
 		int justification, int detect_indent,\
 		int columns, double contrast, \
-		int rotation, double quality) {
+		int rotation, double quality, \
+		double defect_size) {
 	dst_userwidth  = bb_width; // dst_width is adjusted in adjust_params_init
 	dst_userheight = bb_height;
 	zoom_value = font_size;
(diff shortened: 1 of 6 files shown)


chrox:
Since PR #530 contains all these three commits this PR should be closed.

tigran123:
Ok. Btw, clicking on "Close & Comment" does the trick :)

chrox:
@tigran123 Thanks. :)

k2pdfopt.c
```diff
@@ -542,7 +546,7 @@ void k2pdfopt_mupdf_reflow(fz_document *doc, fz_page *page, fz_context *ctx) {
 
 	double dpp,zoom;
 	zoom = zoom_value;
-	double dpi = 250*zoom;
+	double dpi = 250*zoom*src_dpi/300;
 	do {
```
houqp, line 550: Any possibility to send these patches upstream?

chrox, line 550: No, right now I only touched the interface between KPV and K2pdfopt. Most bugs I fixed in k2pdfopt.c were in this part. So if I fixed the memory leak bug that originates in the upstream code I would like to send it to upstream.

Issue #509, opened 2012-10-30
problems with koptreader on reflow


dracodoc:
I tried the refow feature on version v2012.09-389-g1f11513, this is very promising, one of pdf got pretty good result. However there are some problems.

Please look at these screenshots I made:
https://picasaweb.google.com/dracodoc/PdfReflow?authuser=0&authkey=Gv1sRgCMXkmL7-gIqPAQ&feat=directlink

pic 1 and 2, a single column pdf reflowed with small and medium font. 
1. the edge of page is cut off so every line is not complete. 
2. Note there are also large white area around the content, this is not right, why would small font in reflow mode have larger white area than medium font? Compare to pic 3 which is the original page, it seemed that some wrong crop box is applied before reflow.

pic 4 is the original page of another pdf (it's not screenshot by kpv, but screenshot from pc), and pic 5 is the reflowed page. 
1. There is also some cut off but it's different from above. Only the first line of paragraph is cut off for 2 letters.
2. In every wrap of lines of original text, several letters were cut off and the display is not complete.

pic 6,7, 8 don't have the cut off of first line of paragraph, but have the cut off in every wrap of lines.

chrox:
@dracodoc Could you please send me a copy of the original document so that I could try to figure out what's wrong with the reflow mode?

dracodoc:
I uploaded the 3 pdf to a zip here

https://www.box.com/s/bm3m9kwel13fvxlfg3mt

chrox:
@dracodoc All tests are done on my Kindle 3 with KPV2012.10 release and KPV emulator on Ubuntu with a resolution of 600x800. 
1. I can repeat that the edge of page 3 of the bacteria article you provided is cut off as in pic1 and pic2. The reason is that the document was pre-cropped which confused K2pdfopt. I then downloaded the original article from ScienceDirect.com and feeded it to KPV. And everything was fine.
2. I can repeat the glitch on the "The art of seeing" at page 80 as in pic6 and pic7. I suspect it's the same reason as the first one I mentioned above. The book was pre-cropped. 
3. I cannot repeat what you have shown in pic5 with the book you provided at page 16 neither on my Kindle nor on the emulator. Given it has the same problem as the first two books. I'm afraid that you sent me the original book but you tested with a cropped one.

I suggest that you test again with original documents. It seems that k2pdfopt will do the crop things itself but probably in a stupid way even the crop work has been done already.

dracodoc:
[...]

PR #531, opened 2012-11-03
update K2pdfopt to the latest version 1.60

chrox:
New features:
- Add new option to specify trimming or no trimming of excess white space. This is the feature we were expecting on issue #509. @dracodoc
- Add new option to set render quality
- Add new option to set defect size in scanned pages

Bugfixes (from upstream):
- Fixed bmpregion_find_vertical_breaks() bug where it wasn't always correctly interpreting the last section of a region.
- Eliminated divide by zero issue in bmpregion_is_clear() when gt_in gets too small.
- Adjusted critierion for too thick / too thin hyphen in bmpregion_hyphen_detect() to be more correct and to allow for slightly thinner hyphens.

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -489,9 +489,10 @@ static int reflowPage(lua_State *L) {
 	double contrast = luaL_checknumber(L, 15);
 	int rotation = luaL_checkint(L, 16);
 	double quality = luaL_checknumber(L, 17);
+	double defect_size = luaL_checknumber(L, 18);
 
 	k2pdfopt_set_params(width, height, font_size, page_margin, line_spacing, word_spacing, \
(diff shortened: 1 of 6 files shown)


tigran123:
@chrox I didn't have time to test it at all, but since it is strictly koptreader-specific, as long as you tested it yourself I am happy to merge it. Ok?

chrox:
@tigran123 Thanks. I think it's better this patch being tested by other developer before merging it in case that I have missed something.

tigran123:
@chrox did you close it because you are re-doing something in a different way? Just curious...
Ah, I see the new PR, never mind :)

chrox:
@tigran123 Yes. There it is.

PR #533, opened 2012-11-03
LuaJIT update (+shared library), MuPDF update

hwhw:
This updates LuaJIT and MuPDF to their latest versions/revisions. Includes some upstream bug fixes.

Major changes, though: MuPDF now manages the thirdparty libraries as git subrepositories, so we don't download the ZIP anymore.

The patch to libjpeg did not apply anymore. I tested without patching and it worked - I think. @houqp will need to comment on this, I think.

The handling of the header generation utilities in MuPDF has changed - MuPDF has a target for building them with the host toolchain and not building them with a cross tool chain. So we can drop the additional targets in the Makefile.

I also switched LuaJIT to compile as shared library. Doing the same with MuPDF will bring a bunch of problems, namely the third-party libraries. We'll probably have to do them all seperately.


hwhw:
This will probably need some extensive testing.
Dis-entangling the thirdparty libraries is a matter of its own. Maybe we need a "thirdparty" directory ourselves, include the muPDF thirdparty repositories there and have them all build shared libraries. This will be a big job.

hwhw:
Forgot to mention: I also introduced a "-Wl,-rpath" to the linking of the kpdfview executable. It specifies a relative path, though, but I though it might make sense so we don't need to play with LD_LIBRARY_PATH, neither on host (emu mode) nor Kindle.

houqp:
Finally they switch to submodules :) 

OK, so far I cannot produce any segfault without the old dirty libjpeg patch too, so I think it is not needed anymore ;)

tigran123:
Thank you, I will try to test as much as possible now.

tigran123:
At the moment it would not even compile, here is what I did:
1. clone a new kindlepdfviewer tree
2. create a luajit branch and merge upstream/hwhw-master into it
3. make fetchthirdparty (everything is fine)
4. try to compile it:

```
$ make thirdparty kpdfview customupdate > comp.log 2>&1 &
$ tailf comp.log 
make[1]: Entering directory `/home/tigran/Kindle/kindlepdfviewer/mupdf'
    MKDIR build/debug
    CC build/debug/cmapdump.o
    LINK build/debug/cmapdump
/usr/bin/ld: cannot find -ljbig2dec
/usr/bin/ld: cannot find -lopenjpeg
collect2: error: ld returned 1 exit status
make[1]: *** [build/debug/cmapdump] Error 1
make[1]: Leaving directory `/home/tigran/Kindle/kindlepdfviewer/mupdf'
make: *** [mupdf/build/release/libfitz.a] Error 2
^C
[1]+  Exit 2                  make thirdparty kpdfview customupdate > comp.log 2>&1
```

Btw, why is it working in "build/debug" when we specified build="release"? Maybe we should specify build="release" for the "generate" target also?

hwhw:
Yep, glitch in the fetchthirdparty target. I'll add a patch in a couple of minutes.

hwhw:
Hm, but I'm unsure if this really fixes the bug for you. I guess it's still present? Why is it trying to link against jbig2dec and openjpeg for the build utilities? Something is messed up there in the build chain, I guess.

tigran123:
Much better now --- it is compiling. Obviously I had to re-do everything from scratch because "make fetchthirdparty" is non-idempotent operation.

hwhw:
Usually, running make fetchthirdparty twice should not create problems. However, this time it does, it's about the same case as when the Artifex guys change their thirdparty.zip to newer contents. Maybe we need a "superclean" target that will wipe the thirdparty stuff completely? But I fear this would encourage to use it too often...

tigran123:
Running on the emulator I see no problems. Let's see how it works on a Kindle...

tigran123:
On the Kindle DXG I noticed that the RSS of kpdfview process is about 1.5-2MB higher than before (doing the same thing, e.g. opening the same file on the same page and having done nothing else _before_ that). In particular, one file which was opening flawlessly before causes the framework to restart now. Interestingly, this restart didn't affect the running KPV instance, i.e. it continued running as if nothing happened.

Btw, I hope that the default value of anti-aliasing bits is still 8. We rely on the default (instead of setting it with `fz_set_aa_level(ctx, 8)`) so I am always nervous about getting the best possible quality rendering... (and the reason I didn't insist on setting it explicitly is because something else may change that may make this setting no longer optimal)

hwhw:
The memory increase is unfortunate, I agree. It would be interesting where the memory is consumed, though probably not easy to find out.

The anti-aliasing default in mupdf/draw/draw_edge.c is still at 8bits.

tigran123:
Ok, then I see no reason to not merge it now. Shall we do that?

PR #534, opened 2012-11-04
rewrite emu mode input handling

houqp:
I was trying to add touch simulation to emu mode today, but find out that it is not a easy task to map one single SDL event to multiple fake touch events. So I rewrote the part of the code and use fifofile to pipe all the fake events.

As a side effect, I notice that this implementation also makes automated testing possible, because we can simple write scripts that generate inputs to KPV automatically :)

### DIFF
--- a/frontend/ui/inputevent.lua
+++ b/frontend/ui/inputevent.lua
@@ -240,7 +240,8 @@ Input = {
 function Input:init()
 	if util.isEmulated()==1 then
 		-- dummy call that will initialize SDL input handling
-		input.open("")
+		os.execute("rm -f emu_event && mkfifo emu_event")
+		input.open("emu_event")
 		-- SDL key codes
 		self.event_map = self.sdl_event_map
 	else
@@ -292,7 +293,7 @@ function Input:waitEvent(timeout_us, timeout_s)
 		end
 		DEBUG("got error waiting for events:", ev)
 		if ev ~= "Waiting for input failed: 4\n" then
-			-- we abort if the error is not EINTR
+			-- we only abort if the error is not EINTR
 			break
 		end
 	end
--- a/frontend/ui/inputevent.lua
+++ b/frontend/ui/inputevent.lua
@@ -239,7 +239,6 @@ Input = {
 
 function Input:init()
 	if util.isEmulated()==1 then
-		-- dummy call that will initialize SDL input handling
 		os.remove("emu_event")
 		os.execute("mkfifo emu_event")
 		input.open("emu_event")
--- a/frontend/ui/inputevent.lua
+++ b/frontend/ui/inputevent.lua
@@ -240,7 +240,8 @@ Input = {
 function Input:init()
 	if util.isEmulated()==1 then
 		-- dummy call that will initialize SDL input handling
-		os.execute("rm -f emu_event && mkfifo emu_event")
+		os.remove("emu_event")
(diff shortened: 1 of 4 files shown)


frontend/ui/inputevent.lua
```diff
@@ -240,7 +240,8 @@ Input = {
 function Input:init()
 	if util.isEmulated()==1 then
 		-- dummy call that will initialize SDL input handling
-		input.open("")
+		os.execute("rm -f emu_event && mkfifo emu_event")
```
hwhw: I'm not very happy with this one, it isn't likely to hurt things, but it's not all that save either. I would prefer a path that is fixed somehow, but I have no immediate idea. So let's do this for now.

houqp: yeah, it's kind of dirty ;p You mean an absolute path?

tigran123: Out of curiosity, why do you think that absolute pathnames are safer than relative ones? Relative pathname inode lookup is faster, so I don't see why we have things like "./clipboard" instead of just "clipboard". As long as the current working directory does not change from /mnt/us/kindlepdfviewer I see no danger in using relative pathname.

houqp: I am just feeling uncomfortable with calling `rm` and `mkfifo` every time on start. Let's see what @hwhw think :)

hwhw: Hm, I need to think longer, that's for sure, but what just popped up in my
mind is this: does VFAT really know about named FIFOs? Maybe we should
create it in /tmp in the first place... Plus, it should probably not depend
on the "rm" to be successful to go and try to create the FIFO (i.e. the
"&&" is probably wrong).

I don't know, spawning shell commands always feels a bit wrong, I think.

tigran123: Why VFAT? This code is executed only in the emulator mode, not on Kindle.

hwhw: Oh gosh, I was really mislead here, so ignore this and thanks for the
reminder.

houqp: OK, replaced rm with os.remove. But still cannot think of a way to get rid of mkfifo :(

hwhw: Don't bother too much. I just dropped a note about what popped through my
head, we can just commit and be done with it for now.

I was talking about LuaJIT and it's FFI module before and I think the
migration of the input event handling to such a scheme would make a good
example case. When I find time for such a thing, I can get back to this.

Maybe we should define a path prefix, defaulting to "./" - so we have some
means to keep track on the creation of temporary files.

houqp: ah I see, you mean something like following?

``` lua
local ffi = require("ffi")
ffi.cdef[[
typedef unsigned short mode_t;
int mkfifo(const char *pathname, mode_t mode);
]]
ffi.C.mkfifo("emu_event", 0x1b0)
```

hwhw: I was even thinking on a larger scale - implementing everything of the
input layer (and maybe in the more distant future, everything of our C glue
code) via LuaJIT+FFI.

For this issue, I would consider it being unnecessary. It's hard to
understand if only used for this single thing.

hwhw:
Feel free to merge when you're ready.

houqp:
OK, will merge it after I test more tonight :)

Issue #261, opened 2012-09-09
Need some configuration methods for different configurations in future


dracodoc:
At the earlier stage the possible user preferred configurations are not much and simple, so it's acceptable for user to directly edit lua files to change some settings, like global refresh.  They may not be preferred by every user, so it cannot be applied to builds directly.

However, there have been some more preferred changes on certain configurations now, it's not easy to apply them manually for every new build. I generated a patch file with mercurial for all the changes I wanted, but the patch process still need manual verification.

Maybe you can consider some configuration methods in future to incorporate these changes? It's not a high priority issue but better to have it. 

I'll list the changes I applied here for your reference:
1. some changes are minor changes, just parameter value change, like global refresh, pdf margin color, popup dialog box time delay, these can be put in a config file easily. Actually I think there was some font config file in certain NuPogodi patch, but it was not used later.
2. some changes involves different key bindings. I asked houqp to implemented a bookmark jump feature, which would use the shift+ <> keys, so I deleted the zoom grade command for this key in one place, moved the default font zooming key to alt + <> in another place before adding this feature.
This change is more complicated, maybe have to wait the user customizing key binding is available. 

I don't know how to upload a file here, so I just uploaded the patch file I made there:
https://www.box.com/s/1g3rabvm24juia2e9hea

tigran123:
I had a look at your unireader.patch; I agree that 2000ms for those bookmark messages is far too long. However, your suggested value of 50ms is too little. In my current pull request #263 I have reduced the value to one and a half second, which is a bit better than two seconds, but long enough to be noticed by the user (50ms wouldn't be noticed).

dracodoc:
The actual value is not important, in my case the 50ms setting will not make the whole dialog disappear in 50ms, the dx refresh rate is simply not that fast, so it is probably same as 300 ms, but I didn't bother to change it. I can image a K3 user could have a different result, and there probably would be other users that doesn't mind to have a little longer delay time.

The patch file is meant to summarize several points that can be parametrized. I'm not suggesting to incorporate them into official build directly.

dracodoc:
It seemed that there used to be some configuration file, according to the wiki
https://github.com/hwhw/kindlepdfviewer/wiki/Userguide

There used to be .reader.kpdfview.lua, then I believe it changed to settings.reader.lua, then it was eliminated all together.

tigran123:
It is not eliminated, it still exists and is in fact the very first thing that the reader loads on startup.

[...]

PR #537, opened 2012-11-04
separates configurable options from the code

kai771:
I preffer different set of defaults than those shipped with KPV. Those defaults are contained in unireader.lua, where 99% of develpment takes place, so I'd have to manually edit it every time. This moves those options to unireader_h.lua, which doesn't have to be changed.


tigran123:
I have no objection to the symbolic defines for the defaults that you are proposing, but I object to separating the object definition (UniReader = {....}) and its methods (UniReader:functions()) into different source files. It is very convenient to have them in the same file and this is done for all other objects in KPV source code. So, please leave UniReader object alone in the unireader.lua but move the defaults into a separate defaults.lua file. It is not really specific to unireader so it shouldn't be called unireader_h.lua.

Btw, I agree that unireader.lua is growing too big --- perhaps we should consider moving utility (i.e. not object methods) functions out of it into separate source files.

hwhw:
I tend to agree to @tigran123. unireader.lua is bloated, but with features, unfortunately. Things to abstract out would probably be the "alternative navigation" modes.

In the new_ui branch, I seperated functionality into "sub-modules" to a reader object. I'm not sure yet if that's the last word there, either. But it makes an interesting approach. It won't be easy to trim that jungle of unireader.lua down to a nice park...

kai771:
@tigran123 The initial commit was just a quick, lazy cut-and-paste. I agree with all your suggestions and will change it shortly.

dracodoc:
I suggest similar things before on #261. And the config file better compatible with most version upgrades so user can just keep his config file and keep all his settings upon upgrade.

kai771:
Adds background_color to defaults, as requested in #261 (dracodoc called it "margin color", but I think "background color" is more suitable).

houqp:
looks nice now, ready to merge?

kai771:
As far as I'm concerned, it is, but you better ask someone more knowledgeable. :)

houqp:
OK, I will leave it to other developers :)

tigran123:
Looks fine to me.

PR #544, opened 2012-11-05
update K2pdfopt to the latest version 1.60

chrox:
New features:
- Add new option to specify trimming or no trimming of excess white space. This is the feature we were expecting on issue #509. @dracodoc
- Add new option to set render quality
- Add new option to set defect size in scanned pages

Bugfixes (from upstream):
- Fixed bmpregion_find_vertical_breaks() bug where it wasn't always correctly interpreting the last section of a region.
- Eliminated divide by zero issue in bmpregion_is_clear() when gt_in gets too small.
- Adjusted critierion for too thick / too thin hyphen in bmpregion_hyphen_detect() to be more correct and to allow for slightly thinner hyphens.

Cleanup:
- Use koptcontext to keep dozons of parameters for k2pdfopt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ POPENNSLIB := $(POPENNSDIR)/libpopen_noshell.a
 all: kpdfview extr
 
 VERSION?=$(shell git describe HEAD)
-kpdfview: kpdfview.o einkfb.o pdf.o k2pdfopt.o blitbuffer.o drawcontext.o input.o $(POPENNSLIB) util.o ft.o lfs.o mupdfimg.o $(MUPDFLIBS) $(THIRDPARTYLIBS) $(LUALIB) djvu.o $(DJVULIBS) cre.o $(CRENGINELIBS) pic.o pic_jpeg.o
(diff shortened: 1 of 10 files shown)


tigran123:
@chrox I admit today was the first time I really used koptreader seriously (i.e. I had no choice but to use it) and I was very happy with the result. Thank you very much!

One thing though is definitely wrong --- when you call up Aa configuration dialog you should NOT try to retypeset things in realtime, but should collect the changed settings and only retypeset when the user presses fiveway Select or Aa button to confirm the new settings. Otherwise it takes UNBEARABLY long time and discourages the user from ever pressing Aa again...

houqp:
> One thing though is definitely wrong --- when you call up Aa configuration dialog you should NOT try to retypeset things in realtime, but should collect the changed settings and only retypeset when the user presses fiveway Select or Aa button to confirm the new settings. Otherwise it takes UNBEARABLY long time and discourages the user from ever pressing Aa again...

Yeah, I remember some user has reported the same issue before :)

tigran123:
I am testing this PR now...

The bottom of the text is still cut off, so it is not useable with the test file I provided here:

https://github.com/downloads/tigran123/libdjvu/Ideal-pdf-reflow-test.pdf

However, as this fact has nothing to do with this particular PR, I am merging it now.

PR #545, opened 2012-11-05
Battery log

tigran123:
Implement "batlog.txt" battery log.


hwhw:
Call me picky, but I'd love to have this disabled by default - i.e. an option in the configuration. Feel free to merge it, though, I can introduce that later.

tigran123:
Ok, we can disable it by default, I have no objections to that.

PR #553, opened 2012-11-07
bugfix: redefine reader association policy

chrox:
Users can still use default reader for specific type. The "Remember this file" option will add exception to the predefined reader preference. And exception for a document can be made with the "Alt+C" command after opening the doc. This patch fixed the bug reported on issue #480 by @dracodoc.


tigran123:
Thank you! :)

tigran123:
I have just tested it and it breaks PDF support entirely, i.e. after opening a DjVu file (with standard reader) and remembering the association "for this file type (T)" and then closing the file and trying to open a PDF file (with no prior history) it does not open it but refreshes the screen and remains in filechooser.

Switching back to master branch and doing the same thing works fine.

chrox:
Patch 2126166 fixed the bug mentioned above.

tigran123:
Thanks, it works now.

PR #555, opened 2012-11-09
Implements page mode as a toggle

kai771:
Implements a toggle between page and viewport mode in fit-to-content-width and 2-column-mode.

I read that many people asked for page buttons in 2-column-mode to move viewport, like in a fit-to-content-width. I also read that dpavlin wasn't happy with that idea, but I hope he won't mind this: a toggle. Initially, page-turn buttons will move viewport now. Pressing P will make them change page. The setting is saved per book.

While I'was implementing this, I came across a few bugs with 2 column mode, and fixed them on the way. Sorry that I didn't separate them from the new feature, but they are kinda connected to each other so... A lot of changes :(. Needs to be thoroughly tested.


tigran123:
What exactly are the bugs in 2-columns mode? It is hard to test your fixes if we don't know what the bugs were.

There are two 2-column modes: `F` and `Shift-F`. Does page mode apply to both? Did you take care of switching (in both directions) between columns also? If so, that would be very useful. I will test it. But I also would like to test the bugfixes, so I need to know what the bugs were in order to do this.

tigran123:
Hmmm, switching between columns works in one direction (forward) but not always backward. I will re-test it and then will provide a test file so you can see this bug for yourself. This is in page mode in normal (left-to-right) books.

kai771:
@tigran123 one of the bugs was incomplete handling of files that already had bbox defined (like Urantia). The way to test: open Urantia, press F, then press up - it will pan up outside the bbox, to the top of the page. Likewise, the bottom of bbox was not respected at all - page turn backward would line up to the bottom of the page, insted to the bottom of the bbox. I also cleaned up formulas, as I came to better understand them.

PR applies to both 2-column modes - both F and Shift-F, and switching between columns in both directions is supposed to work. Of course, there might be bugs - so please test and provide a test file. I suspect the line `if unireader.offset_x - 0.01 > unireader.pan_x then` is the culprit for the bug you encountered, and that the cause is the rounding error. The original line was `if unireader.offset_x > 0 then`, and I encountered the same error as you with it, before I implemented rtl_mode. I then changed it to `if unireader.offset_x > unireader.pan_margin then` and it worked correctly, but it didn't make sense, so I tried to correct it now. I'll take another look with the problematic file you provide.

unireader.lua
```diff
@@ -64,6 +64,8 @@ UniReader = {
 	pan_by_page = DPAN_BY_PAGE, -- using shift_[xy] or width/height
 	pan_x = DPAN_X, -- top-left offset of page when pan activated
 	pan_y = DPAN_Y,
+	pan_x1 = 0, -- bottom-right offset of page when pan activated
```
dpavlin, line 67: Can we call this pan_x_bottom_right to have better idea in rest of the code how it's different from pan_x ?
It's longer, but I originally suggested pan_x_margins, so it's confusing, at least for me.

unireader.lua
```diff
@@ -2704,6 +2831,15 @@ function UniReader:addAllCommands()
 			self:redrawCurrentPage()
 		end)
 
+	self.commands:add(KEY_P, nil, "P",
+		"toggle page-buttons mode: viewport/page",
+		function(unireader)
+			unireader.page_mode_enable = not unireader.page_mode_enable
+			InfoMessage:inform("Page-buttons move "..(unireader.page_mode_enable and "page" or "viewport"), nil, 1, MSG_AUX)
```
dpavlin, line 2838: We should have 1000ms timeout, not 1ms (in emulator I just see blink, on real device slowness of e-ink helps, but...)

kai771:
I think that pan_x _margin wouldn't be appropriate, since it has nothing to do with margin as used in the code - margin being the amount of pixels around the bbox. In unireader:setZoom(), we had `local x0, y0, x1, y1 = page:getUsedBBox()`. If pan_x corresponds to x0, pan_x1 corresponds to x1. That was the original idea. However, I set pan_x1 to the offset of the left edge of second column in ``self.pan_x1 = -1 * x1 * self.globalzoom - margin + G_width`. However again, I didn't use pan_x1 anywhere. Pan_y1 was the needed part, I just left pan_x1 for possible future use. I could change them to pan_x_bottom_right and pan_y_bottom_right, but we will lose the connections with x1 and y1 returned by getUsedBBox. What do you guys think?

dpavlin:
@kai771 ignore my comment, you are right.

kai771:
@dpavlin Btw, you don't mind implementing this? I read that you didn't like page buttons moving viewport.

dpavlin:
I must say that moving viewport with page buttons does work nicely, and it's probably feature which we want to enable by default. My original idea about not touching page buttons was to provide some way to quickly just several pages forwards or backwards but that is probably better done by switching to one of full-page layouts anyway or using bookmarks.

If we don't turn this option by defaults (since it's more logical for new users IMHO), I would at least suggest to make it global for whole reader and not per-document.

Now when I think about it, we might even make it only option, so we don't waste another keyboard shortcut... (but this can be separate PR, I like it so much ;-) 

On third thought, it would be better not to merge code which we will remove later. So, I don't have any objections to make page buttons move viewport always, do you agree?

kai771:
Oh, and I should remove pan_x and pan_y from defaults.lua - they are not parameters that should be set by the user. I'd like to do defaults.lua clean-up in separate PR, after this one is merged, though. There might be some more unneeded parameters there.

@dpavlin it is like this by default, because I wanted it to be consistent with fit-to-page-width behaviour. Pressing P in both fit-to-page-width and 2-column will make page buttons turn pages, for quick browsing without changing zoom mode. That was the idea, anyway.

dpavlin:
So the only question is should we remote toggle and just use viewport page always (I vote for yes -- sorry about writing toggle code).

kai771:
I'd still keep the toggle because there might still be people that prefer the old way, and I don't think we should take it away from them :). Just my opinion, though.

dpavlin:
Could you fix inform timeout so we can merge it?

kai771:
Sorry, I don't understand you. What do I need to fix?

dpavlin:
inform timeout is 1ms, it should probably be 1000ms

kai771:
Do you mean this line?
`InfoMessage:inform("Page-buttons move "..(unireader.page_mode_enable and "page" or "viewport"), nil, 1, MSG_AUX)`
I think tigran recently changed all inform timeouts to 1 (if it's that 1 in `1, MSG_AUX)`)

dpavlin:
It just blinks for me in emulator. You don't see that behaviour?

kai771:
Yes, but all other toggles just blink in the emulator. It does stay for a while on a kindle :).

tigran123:
Was the bug I mentioned (backward column switching broken) fixed? I didn't see any commit since then so the merged version still must have the bug.

tigran123:
Btw, all the toggle inform timeouts were always set to `nil` --- I did NOT change anything. Yes, for the emulator it is too quick, but for the Kindle just right. And in fact it was @dpavlin 's idea to do that (he was the first to set it to `nil` for the gamma update thing).

And no, that 1 has nothing to do with the timeout, rather it is the `nil`.

tigran123:
Ok, two bugs so far:
1. the one I already mentioned about column switching not working.
2. Most importantly, the existing (very critical) functionality of Shift-F is broken. Column switching (in Shift-F mode) now requires one extra key press when the viewport hits the top of the page.

(deleted the one about default value being false because I see that that is intended, ok).

But I haven't tested it properly yet, so there may be more bugs still. I think we should have given @kai771 more time to resolve the bugs before merging it, but it is not too late to fix them in a new PR, of course :)

tigran123:
Also, I agree with @kai771 that this should be left as a toggle because sometimes one wants to turn pages quickly and sometimes to read continuously from top to bottom, so the `P` command is very handy.

tigran123:
actually, one can overcome the "extra key needed when hitting the top" by making sure the bbox doesn't have too much vertical white space at the top. Hmmm, strangely, before the change it wasn't too sensitive to that. Oh well, this is trivial and I can live with it, so don't bother trying to fix it. It is good enough as it is :)

kai771:
I'd still like to fix it, but please provide the files and the description how to reproduce it - I didn't encounter either of above mentioned bugs with my tests, so I wasn't able to try and fix it.

tigran123:
Actually, one of the remaining bugs is critical, i.e. does require fixing (not the vertical space one mentioned above which is merely cosmetic).

Here is the file:

https://github.com/downloads/tigran123/libdjvu/page-mode-enable-test.pdf

To reproduce the critical bug:
1. goto page 3 and set bbox (to include just the two columns, NOT the extra text in the margins)
2. goto page 4 and set bbox (again to include just the two columns)
3. go back to page 2
4. enter Shift-F mode
5. page down a few times (11 or 12 times) to pass a few pages forward
6. now page up to get back to where you started.

Instead of reaching the left column on page 3 you will see only a part of it and also the bit from the margin that reads 

```
PUP: first
numeric-only
cross-reference
appears here. All
OK with you?
```

But according to my choice of bbox I didn't expect to see this. Instead, I expected to see the actual text of the left column.
I don't think this bug was present before your change.

tigran123:
I have reproduced the same bug with Urantia-A4 file as well, this one:

http://urantia-book.org/downloads/English-UK/Urantia-ENG-UK-A4-twocolumn.zip

i.e. after you page down enough times (in Shift-F mode with an appropriate Shift-X bbox) to turn a few pages and then start pressing page Up to get back to the beginning then (after one or two full pages which are shown fine) the left column starts shifting to the right, i.e. the viewport is shifted to the left and you only see a chunk of the left column, not all of it.

So, the twocolumn mode is definitely quite broken at the moment.

kai771:
@tigran123 As far as I can tell, with #558 Urantia works fine too, but please do test it yourself. (If something else comes to mind, please test that too :) )

tigran123:
Yes, I will :)

dpavlin:
Sorry for merging this one with bug, I was carried away with excitement because I figured out that page keys are much more useful now :-)

As far as timeout is concerned, I removed it for gamma since real criteria is page displayed on screen (as opposed to numbers visible in info box). It was not my intention to remove all timeouts, but slowness of e-ink on real device does indeed help. Should we add one "is emulator" check to introduce delay just in emulator? It's hard to read info box when in disappears so quickly.

tigran123:
Yes, it may be a good idea --- to introduce a check inside :inform() method to have a non-nil delay for the emulator.

kai771:
I'd like to suggest a simpler solution - to implement timeout like a settable value in defaults.lua. I've been meaning to do just that - and then I found that there are a lot of different values for timeout in the source - nil, 1000, 1500 and 2000. Do we need all this different values or would only one suffice?
Update: On second thought, both would probably be a good idea. We could add both DINFO_TIMEOUT and DEMU_INFO_TIMEOUT to defaults.lua, along with a check inside :infrom(). The question that reamins is do we need different values, or would only one suffice? Or maybe two - DINFO_TIMEOUT_SLOW and DINFO_TIMEOUT_FAST? What do you guys think?

tigran123:
I think it is important to differentiate (on non-emulator) between nil and non-nil values. But on the emulator, yes, we could change nil to the same value as the non-nil ones.

As for the the different non-nil values (1000, 1500, 2000) I think they should all be set to a SINGLE non-nil value (I suggest 1000). But not by directly changing each instance to a different hardcoded value but by providing a default value for that argument.

kai771:
Ok, I agree with you. Would you like to do it, or should I? ;) (If you choose me, it'll have to wait a couple of hours :))

kai771:
Ok, I did it... #560 :)

PR #556, opened 2012-11-09
feature: add bbox support in koptreader

chrox:
Reflow can be configured to use bbox which could remove side comments and defects in the margin and get much better reflowed page for some documents. Currently the bbox can only be set in PDFReader or DJVUReader reader. But eventually there will be bbox setting routines in KOPTReader.

Thirdparty should be cleaned and refetched and remade to get this feature in operation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,7 @@ HOSTCXX:=g++
 HOSTAR:=ar
 
 # Base CFLAGS, without arch. We'll need it for luajit, because its Makefiles do some tricky stuff to differentiate HOST/TARGET
-BASE_CFLAGS:=-O2 -ffast-math -pipe
-KOPT_CFLAGS:=-O3 -ffast-math -fno-finite-math-only -pipe
+BASE_CFLAGS:=-O2 -ffast-math -pipe -fomit-frame-pointer
 # Use this for debugging:
 #BASE_CFLAGS:=-O0 -g
 # Misc GCC tricks to ensure backward compatibility with the K2, even when using a fairly recent TC (Linaro/MG).
@@ -269,7 +268,7 @@ $(POPENNSLIB):
 	$(MAKE) -C $(POPENNSDIR) CC="$(CC)" AR="$(AR)"
 
 $(K2PDFOPTLIB):
-	$(MAKE) -C $(K2PDFOPTLIBDIR) BUILDMODE=shared CC="$(CC)" CFLAGS="$(KOPT_CFLAGS)" AR="$(AR)" all
+	$(MAKE) -C $(K2PDFOPTLIBDIR) BUILDMODE=shared CC="$(CC)" CFLAGS="$(CFLAGS) -O3" AR="$(AR)" all
 	test -d $(LIBDIR) || mkdir $(LIBDIR)
 	cp -a $(K2PDFOPTLIBDIR)/libk2pdfopt.so* $(LIBDIR)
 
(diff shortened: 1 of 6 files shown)


tigran123:
1. -O3 implies -fomit-frame-pointer, so the latter is not needed.
2. What does reflow OFF setting do? Turn the reflow reader into a non-reflow? Then how would it differ from the standard one?

chrox:
@tigran123 I removed -fomit-frame-pointer in both BASE_CFLAGS and KOPT_CFLAGS as -fomit-frame-pointer is enabled at level -O, -O2, -O3. Turn reflow off is more an option for debugging in reflow mode. And it will automatically remove white margin of the current pages just as manual bbox does in standard reader.

tigran123:
No, -fomit-frame-pointer is not enabled at levels -O and -O2, only at -O3.

Hmmm, strange, the table here says that you are right:

http://www.linuxjournal.com/files/linuxjournal.com/linuxjournal/articles/072/7269/7269t1.jpg

But I remember that we always had to include -fomit-frame-pointer when compiling Linux kernel with -O2. Perhaps the answer lies in the different gcc versions, i.e. the more recent ones must have changed the behaviour.

Anyway, we must definitely TEST (i.e. manually disassemble a chunk of code and see) to make sure that -fomit-frame-pointer is not needed at -O2 level before applying your patch.

chrox:
http://gcc.gnu.org/onlinedocs/gcc-4.4.2/gcc/Optimize-Options.html#Optimize-Options

It says that -fomit-frame-pointer is `Enabled at levels -O, -O2, -O3, -Os.`

tigran123:
Yes, and `/usr/src/linux/Makefile` says:

```
HOSTCFLAGS   = -Wall -Wmissing-prototypes -Wstrict-prototypes -O2 -fomit-frame-pointer
```

I certainly trust Linux kernel developers' knowledge of gcc more than the developers of gcc itself, so this should still be verified by objdump --disassemble.

chrox:
The kernel guys may not do this thing properly. You can test it in this way:

```
echo 'int main() {return 0;}' > test.c && arm-none-linux-gnueabi-gcc -v -Q -O test.c -o test && rm test.c test
```

```
echo 'int main() {return 0;}' > test.c && arm-none-linux-gnueabi-gcc -v -Q -O2 test.c -o test && rm test.c test
```

```
echo 'int main() {return 0;}' > test.c && arm-none-linux-gnueabi-gcc -v -Q -O3 test.c -o test && rm test.c test
```

```
echo 'int main() {return 0;}' > test.c && arm-none-linux-gnueabi-gcc -v -Q -Os test.c -o test && rm test.c test
```

In the `options enabled` section `-fomit-frame-pointer` is present in all these tests.

chrox:
I recommend you guys to try this new feature in the koptreader. If your pdf document has bbox information stored priviously just open it in PDFReflow and switch `Trim Page` option to manual. Then the page margin will be trimmed according to the bbox before the reflow process which could generate rather good reflowed page if the bbox is properly set. And with some optimization the average time to reflow one page has decreased to around 3 seconds at medium rendering quality.

tigran123:
Hmmm, maybe we should use -O3 instead of -O2 -fomit-frame-pointer in BASE_CFLAGS also? But this can be done in a separate PR as it is not related to koptreader. Ok, I see no reason not to merge this patch.

But perhaps I should test it, just to be 100% sure :)

tigran123:
Yes, seems to work, merging now.

PR #557, opened 2012-11-10
bugfix: readerchooser dialog should not be visible when opening a file

chrox:
when select a reader to open a document the readerchooser dialog is
overlayed with opening file notification window. The overlayed three
windows(file chooser, reader chooser and notification window) will make
users feel unconfortable easpecially when opening in koptreader which
usually freezes the screen for several seconds.

With this patch the reader chooser dialog is closed before the opening
file noticifation is shown up.

### DIFF
--- a/readerchooser.lua
+++ b/readerchooser.lua
@@ -1,6 +1,7 @@
 require "font"
 require "keys"
 require "settings"
+require "screen"
 require "pdfreader"
 require "djvureader"
 require "koptreader"
@@ -94,6 +95,7 @@ function ReaderChooser:getReaderByName(filename)
 			end
 		end
 		-- need to choose reader
+		Screen:saveCurrentBB()
 		local name = self:choose(readers)
 		if name then
 			if self.remember_preference then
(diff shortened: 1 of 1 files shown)


tigran123:
I think that save/restore BB in this content is just a waste of cpu cycles (and battery) and "over-engineering". I don't see any reason why the user should feel uncomfortable about seeing the "opening document..." window overlayed on top of the "Complete action using" window. It makes perfect logical sense that after choosing the reader the application proceeds to open the document using that reader and displays appropriate notification.

I would vote against merging this patch, but let @dpavlin and @houqp make the final decision.

chrox:
I beg to differ. The readerchooser dialog and the notification window each has their own function and lies at their own logic scope. Once the reader object is passed to filechooser the mission of readerchooser is done and it should be closed immediately. And then a notification window is popped-up to inform users that file is under processing please be patient. The latter process has nothing to do with readerchooser. So what readerchooser is there for? I don't think saving battery can justify this bug.

tigran123:
Ok, you convinced me :)

chrox:
Thanks. :)

PR #558, opened 2012-11-10
Fixes bugs with viewport mode reported by tigran123 in #555

kai771:
This should fix the bug reported by tigran123 in #555. As I thought, the cause was rounding errors. Don't you just love floats? Please test and report.

I'd also like to fix the bug with vertical space too, because the cause is probably the same (rounding error, just in different place). If you can, please provide the file and details.

@dpavlin I said earlier that I don't use pan_x1. It turned out that I do. Formulas can be written without using it, but I think that using it makes them simpler. So, pan_x is set to the left edge of the left column, and pan_x1 is set to the left edge of the right column.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1357,7 +1357,7 @@ function UniReader:setzoom(page, preCache)
 		self.offset_y = -1 * y0 * self.globalzoom + margin
 		self.pan_x = self.offset_x
 		self.pan_y = self.offset_y
-		self.pan_x1 = -1 * x1 * self.globalzoom - margin + G_width
+		self.pan_x1 = -1 * x1 * self.globalzoom - margin + G_width  -- sets pan_x1 to left edge of the right column
 		self.pan_y1 = -1 * y1 * self.globalzoom - margin + G_height
 		self.pan_by_page = self.globalzoom_mode -- store for later and enable pan_by_page
 		self.globalzoom_mode = self.ZOOM_BY_VALUE -- enable pan mode
@@ -1738,7 +1738,8 @@ function UniReader:twoColNextView()
 	-- can't go down anymore
 	else
 		if self.rtl_mode_enable then -- rtl_mode enabled
-		  if self.offset_x <= self.pan_x then
+			-- can go left?
+		  if self.offset_x + 0.01 < self.pan_x then
 				self.offset_x = self.offset_x + x
 				self.offset_y = self.pan_y
 				self.show_overlap = 0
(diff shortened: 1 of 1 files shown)


tigran123:
Very quick --- thank you! I promise to test it tomorrow.

tigran123:
Some wise man said "don't leave for tomorrow that which can be done today", so I have tested it now and confirm that it works fine. Thank you. As far as I am concerned it can be merged, but I'll leave it a bit to give @dpavlin a chance to review the code. If no objections are raised by him (or anyone else) I'll merge it tomorrow.

kai771:
I'd still like to fix the vertical bug too, but I can't fix what I can't reproduce, so please, if it's not too much trouble... :)

tigran123:
Ok, take the same file (Princet...) and go to page 3 (marked as page 2 in the top left corner). Define bbox with Shift-X and then press Shift-F. Now press Page Down once. It "looks" as if you are already at the bottom of the column so you expect that the next Page Down would move you to the top of the next column, but it does "nothing". Nothing in quotes because I suppose it shifts the viewport one or more pixels down because really there is some vertical white space which was "not shown" yet. I understand now that this is not really a bug and is in fact due to the bugfix you mentioned originally, i.e. that the vertical bbox boundaries were not respected and now they are. But it feels strange --- how come previously this was never a problem, i.e. "not respecting vertical bbox boundaries" never caused not displaying any real data? Was it just by pure lack (and the common fact that all books have some white bottom and top margins)?

Try it and you will see what I mean. It is quite annoying to have to press an extra key when it is clear (to human, but not to the computer :) that there really is nothing more to be displayed there...

kai771:
Yes, it seems it is not a bug. It was not a problem before, because when you reached what appeared to be the bottom of the page, you wouldn't press fw_down anymore - you would press fw_right. You can do that now too :), and it will move you to the top of the right column, just as before. It's quite rare (in my opinion - it took me 11 tries) to set the bbox in such a way that would cause it.

If you still pressed fw_down instead of fw_right in versions prior to #555, it would move you down to the bottom of the _page_. Now it will move you to the bottom of the _bbox_. So, if anything, you can now set the bottom of the bbox a bit more up and avoid this, which wasn't the case before.

(I didn't test what I said above on versions prior to #555, so it's not 100% sure, but I'd say I'm 99% convinced it is :))

tigran123:
Ah, yes, very good point about pressing fiveway right button to force switching the column. Ok, now everything is perfect, thank you for this work and for the explanation! :)

PR #559, opened 2012-11-11
add bbox modification inside of koptreader

chrox:
The modification code is copied directly from unireader as I haven't find a more elegant way yet.


tigran123:
Yes, seems to work. Merging into master...

PR #560, opened 2012-11-11
Info_timeout in emulator and defaults.lua clean-up

kai771:
Cleans up defaults.lua (removes some values that should not be set by the user) and introduces DINFO_TIMEOUT_FAST and DINFO_TIMEOUT_SLOW. All :inform() calls in unireader.lua that had nil as the timeout have DINFO_TIMEOUT_FAST now, and all other values (1000, 1500, 2000) have DINFO_TIMEOUT_SLOW. When run in emulator, :inform() will always use value set as DINFO_TIMEOUT_SLOW.

### DIFF
--- a/defaults.lua
+++ b/defaults.lua
@@ -23,11 +23,8 @@ DSHIFT_Y = 50
 -- step to change zoom manually, default = 16%
 DSTEP_MANUAL_ZOOM = 16
 DPAN_BY_PAGE = false -- using shift_[xy] or width/height
-DPAN_X = 0 -- top-left offset of page when pan activated
-DPAN_Y = 0
 DPAN_MARGIN = 5 -- horizontal margin for two-column zoom (in pixels)
 DPAN_OVERLAP_VERTICAL = 30
-DSHOW_OVERLAP = 0
 
 -- tile cache configuration:
 DCACHE_MAX_MEMSIZE = 1024*1024*5 -- 5MB tile cache
@@ -42,8 +39,9 @@ DBATTERY_LOGGING = false
 -- background colour: 8 = gray, 0 = white, 15 = black
 DBACKGROUND_COLOR = 8
 
--- page mode
-DPAGE_MODE_ENABLE = false
+-- timeout for info messages in ms
+DINFO_TIMEOUT_FAST=nil
+DINFO_TIMEOUT_SLOW=1000
 
 -- toggle defaults
 DUNIREADER_SHOW_OVERLAP_ENABLE = true
--- a/dialog.lua
+++ b/dialog.lua
@@ -136,6 +136,9 @@ function InfoMessage:inform(text, msec, refresh_mode, message_importance, altern
 	end
 	if not popup then return end -- to avoid drawing popup window
 	self.ImageFile = self.Images[message_importance] -- select proper image for window
+	if util.isEmulated()==1 then
+		msec = DINFO_TIMEOUT_SLOW
+	end
 	if not msec or msec == 0 then
 		InfoMessage:show(text, refresh_mode)
 	else
--- a/unireader.lua
+++ b/unireader.lua
@@ -62,13 +62,13 @@ UniReader = {
 	-- step to change zoom manually, default = 16%
 	step_manual_zoom = DSTEP_MANUAL_ZOOM,
 	pan_by_page = DPAN_BY_PAGE, -- using shift_[xy] or width/height
-	pan_x = DPAN_X, -- top-left offset of page when pan activated
-	pan_y = DPAN_Y,
+	pan_x = 0, -- top-left offset of page when pan activated
+	pan_y = 0,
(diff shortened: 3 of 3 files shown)


tigran123:
Btw, this change made the Shift-X function slightly inconvenient in the emulator i.e. one has to either wait 1 second or press extra keys. Maybe for the Shift-X command handler we really should pass `nil` to those two calls to :inform() instead of DINFO_TIMEOUT_FAST?

Update: I see that just passing `nil` wouldn't be good enough. Ok, we may have to either call :show() method or (better) change the logic in dialog.lua to allow "really no timeout" setting.

tigran123:
Also, you missed quite a few places that still have the hardcoded timeout value:

```
$ grep InfoMessage:inform *lua | grep 1500
crereader.lua:      InfoMessage:inform("No Table of Contents ", 1500, 1, MSG_WARN)
crereader.lua:              InfoMessage:inform("Page already marked ", 1500, 1, MSG_WARN)
crereader.lua:              InfoMessage:inform("Page marked ", 1500, 1, MSG_WARN)
filechooser.lua:        InfoMessage:inform("Folder can't be deleted! ", 1500, 1, MSG_ERROR,
$ grep InfoMessage:inform *lua | grep 1000
filechooser.lua:                InfoMessage:inform("File copied to clipboard ", 1000, 1, MSG_WARN,
inputbox.lua:                   InfoMessage:inform("No input! ", 1000, 1, MSG_WARN, "There is nothing to calculate")
inputbox.lua:                       InfoMessage:inform("Invalid input! ", 1000, 1, MSG_WARN)
koptconfig.lua: InfoMessage:inform("New page bbox ", 1000, 1, MSG_WARN, "New page bounding box")
pdfreader.lua:          InfoMessage:inform("No attachments found ", 1000, 1, MSG_WARN)
pdfreader.lua:      InfoMessage:inform("Failed to save attachments ", 1000, 1, MSG_WARN)
screen.lua: InfoMessage:inform(msg..string.format("%.2fs ", diff), 2000, 1, MSG_WARN, msg..math.ceil(diff*1000).." milliseconds")
$ grep InfoMessage:inform *lua | grep 2000
crereader.lua:      InfoMessage:inform("No jump history found ", 2000, 1, MSG_WARN)
crereader.lua:      InfoMessage:inform("No bookmark found ", 2000, 1, MSG_WARN)
crereader.lua:          InfoMessage:inform("Default document font set ", 2000, 1, MSG_WARN,
crereader.lua:              InfoMessage:inform("Already first jump ", 2000, 1, MSG_WARN)
crereader.lua:              InfoMessage:inform("Already last jump ", 2000, 1, MSG_WARN)
crereader.lua:      InfoMessage:inform( found.." hits '"..search.."' pos "..pos, 2000, 1, MSG_WARN)
crereader.lua:      InfoMessage:inform( "'"..search.."' not found in document ", 2000, 1, MSG_WARN)
filechooser.lua:    InfoMessage:inform("Unsupported function! ", 2000, 1, MSG_WARN,
filehistory.lua:                InfoMessage:inform("No search hits ", 2000, 1, MSG_WARN,
filehistory.lua:                InfoMessage:inform("File does not exist ", 2000, 1, MSG_ERROR)
filehistory.lua:        InfoMessage:inform("No reading history ", 2000, 1, MSG_WARN, "The reading history is empty!")
filesearcher.lua:                   InfoMessage:inform("No search hits ", 2000, 1, MSG_WARN,
reader.lua:             InfoMessage:inform(err:sub(1,30), 2000, 1, MSG_ERROR)
reader.lua:             InfoMessage:inform("Error opening document! ", 2000, 1, MSG_ERROR)
screen.lua: InfoMessage:inform(msg..string.format("%.2fs ", diff), 2000, 1, MSG_WARN, msg..math.ceil(diff*1000).." milliseconds")
```

tigran123:
Oh, we also completely broke interactive stuff like "Press `Y` to confirm", so we'll need to redo this properly.

kai771:
Yeah, I just did unireader.lua. Didn't even occur to me that other files needs checking. :(
Don't understand the second part - what interactive stuff?

tigran123:
In filechooser.lua there is code like this:

```
InfoMessage:inform("Press 'Y' to confirm ",nil, 0, MSG_CONFIRM)
if ReturnKey() == KEY_Y then self:deleteFileAtPosition(pos) end
```

we can't just blindly change that nil to non-nil --- try it and see what happens :)

tigran123:
The problem is that if we set the timeout to non-nil then the message will disappear, but we want it to stay on the screen until the user presses something.

tigran123:
Ok, we can differentiate between `nil` and `0`, i.e. those places which really cannot have a non-nil timeout should pass 0. So, we need two values: DINFO_TIMEOUT_FAST and DINFO_TIMEOUT_ZERO and only the first one should be delayed on the emulator, the other one should be allowed to go through with no delay.

kai771:
Should be fixed with #567.

PR #561, opened 2012-11-11
enable overlap in comics_mode

kai771:
When I first implemented comics mode, I meant it for reading comics. Overlap is useless for that, so I didn't bother to support it, and turned it off automatically. Now, on second thought, I thought it might be good for reading books too, and that people would expect overlap there. Having implemented that, I thought it should be made default, to be consistent with the behaviour of 2 column mode. So I made it default. If you don't agree with it, let me now and I'll change it.


tigran123:
When I am reading a book continuously I expect the last line of the current view to be at the top of the next view, so I am against making the comics mode default for any readers.

But in the second (i.e. unireader.lua) part of your patch I have two comments:
1. I totally agree that that "if not self.comics_mode_enable" part should be removed. Each toggle should be independent of others and one should let the user decide how or whether to combine them.
2. What do your changes to next/prevView() do? It would be nicer if you had explained them in the commit message.

kai771:
@tigran123 2. Nothing really, just calculate offset of overlap, that's all.

About making it default - it is how it's done in 2 columns mode - when you reach the end of the screen, and there isn't enough of page left to fill the next one, the next line doesn't start on top of the screen. I think it's more usefull to have top of the screen filled with already read material (or if there are pictures, or formulas) than having the bottom of the screen empty, but it's the matter of opinion. It's easily configurable now. I just thought that the new way is more useful and more consistent with 2 column mode, thus better to be default. Old way is just a press of the button away (or edit of defaults.lua :)).

tigran123:
I would have no objections to making it default as long as you agree to rename "comics mode" to something more descriptive, such as "align viewport to top/bottom" or something like that. Then, when reading any book the users would know how to configure this for their liking, whereas now they would most likely ignore this "comics mode" (i.e. reasonably assume it is for comicses only). Ok?

tigran123:
Btw, you can keep all the variables etc as comics_mode --- that's fine. I am talking only about the help (and info when toggling) message corresponding to the `C` command.

kai771:
I have no objections to renaming it - I called it comics mode lacking better term. :) But "align viewport to top/bottom" doesn't seem quite right... How about "always fill screen with page content"? Hm, I don't really like that either... Let's see if we can come up with something better before I change it.

kai771:
How about "constrain to screen"?

tigran123:
And what was not quite right about "align the viewport to top/bottom"? In comics mode the bottom of the viewport is aligned to the bottom. In non-comics mode the top of the viewport is aligned to the top.

kai771:
I guess you're right. I was overcomplicating a simple thing (and trying to have a "C" somewhere). "align the viewport to top/bottom" it is :).

tigran123:
Ok, very good, but:
1. This PR cannot be merged automatically, you'll have to redo it (it conflicts with other changes just merged).
2. The message "align the viewport to top/bottom ON/OFF " should be shortened to "Align the viewport to TOP" and "Align the viewport to BOTTOM" (i.e. depending on the value of the flag). Also, note the capital "A" here.

kai771:
Sorry, I'll fix it right away.

tigran123:
Would you mind including this trivial set of changes as well, please: SHA1: 7f491c0e4ab24628bac3b58cde99c81551a01078

They would conflict with your new PR and this would save me time re-doing it after merging the new version of your PR. The changes are simple: just get rid of redrawCurrentPage() when not needed (including the `C` command, hence the future conflict, this is nothing to do with the present PR conflict) and change the Alt-Z toggle to use "and ... or ..." style.

kai771:
I'll see what I can do :).

tigran123:
Ok, thank you!

tigran123:
This PR can be closed for now.

tigran123:
No rush --- I am switching computer off until tomorrow: time to actually make use of all these things we are creating happily :)

PR #566, opened 2012-11-11
Bugfix: link with the static libjpeg.a from mupdf/thirdparty

tigran123:
We should explicitly link the picviewer with the libjpeg.a from mupdf/thirdparty and not with the shared libjpeg.so that may happen to be installed on the system.


tigran123:
Btw, ldd kpdfview.emulator would still show libjpeg.so (but NOT readelf -d kpdfview.emulator) because libdjvulibre.so is linked with the system libjpeg.so rather than with our static one. But this is the case only on the emulator, not on the real Kindle.

PR #567, opened 2012-11-11
info_timeout2 - fixes issues reported by tigran123 in #560

kai771:
This should be all.


dialog.lua
```diff
@@ -136,7 +136,7 @@ function InfoMessage:inform(text, msec, refresh_mode, message_importance, altern
 	end
 	if not popup then return end -- to avoid drawing popup window
 	self.ImageFile = self.Images[message_importance] -- select proper image for window
-	if util.isEmulated()==1 then
+	if util.isEmulated()==1 and msec == 0 then
```
tigran123, line 139: Shouldn't this be `and msec == DINFO_TIMEOUT_FAST`?

kai771:
No, because I wanted to leave a way to have different values for slow and fast on emulator too.

tigran123:
Perhaps we should increase DINFO_TIMEOUT_SLOW to 1500. I thought that 1000 is enough but having tried it on Kindle I see that it is too fast for some of the long messages to be read entirely.

kai771:
I was thinking... maybe it's not a good idea to change the value for DINFO_TIMEOUT_FAST from 0 to the value of DINFO_TIMEOUT_SLOW for the emulator, because it really slows it down. Maybe it's better to set it to something like 500?

tigran123:
No, it is just all those which should not be slowed down (like "Opening document...", "Searching...", etc) should be changed back to `nil`. Only the useful ones (like toggles) should be set to DINFO_TIMEOUT_FAST because they need slowing down. Common sense consistency is a lot more important than mere "code consistency".

kai771:
I disagree. They're all meant to be seen. Otherwise, there wouldn't be a call to :inform() at all. The fact is that setting them to nil on kindle doesn't last 0 ms, but more like 300-500 due to the slowness of the screen. Imagine that the kindle's screen is as fast as your pc - you wouldn't want them to just blip on the Kindle, so you wouldn't set them to nil then. So, to stay true to how the real Kindle feels (as much as possible), I think we should set them to the aproximate amount of time they last on Kindle.

tigran123:
Why wouldn't I want them (i.e. the unimportant ones, NOT the toggle indicators!) to "just blip" on the Kindle? I would REALLY want them to do just that. In fact, deleting them altogether is another good option. The only problem with that is that koptreader can take several seconds to open a file and that is why we may wish to keep "Opening document..." one. But the "Searching..." or "Examining folder..." etc can definitely go --- there is no way to have such a huge directory structure on Kindle that would take more than a couple of seconds to run "du" on it.

Also, it is not only koptreader but crengine can take ages opening an ebook. Maybe there should be reader-specific messages --- "Reflowing a book..." coming from koptreader and "Parsing an ebook..." coming from crengine. But by default no messages at all.

kai771:
If you want them removed, that's entirely different matter, and then, imno, calls to :inform() should be removed. Even if some things take several seconds, if the kinde's screen was as fast as the pc, the info message would just blip if set to nil, regardles of the length of time the operation takes to complete, thus failing to inform the user. As for this issue, I think we should strive to simulate the kindle feel.

If you agree, please change the line (in dialog.lua)
`msec = DINFO_TIMEOUT_SLOW`
to
`msec = 500` (or maybe 300). I won't have acces to my pc until late afternoon.

Update: I sent another PR - #568. I set msec to 300 - I think this is comparable to the real Kindle.

PR #568, opened 2012-11-12
speeds up info messages of toggles etc. in emulator

kai771:
In an attempt to slow down Info Messages on emulator, so they can be read, we made them too slow. This will make them "just right" - comparable to how they appear on Kindle.


tigran123:
No problem at all :)

PR #570, opened 2012-11-13
add multi-threaded precache in koptreaders

chrox:
Because the original Lua program is single threaded so that page rendering process has to run in user inputloop. Although there is a pretty good precache system to save user's time spending on waiting for the rendering when going to the next page, user input is indeed blocked when running the precache thing. The situation is even worse in koptreader as reflowing one page would usually take several seconds, in this period of time users cannot move to the next page view even it's already in the cache. This patch will let precache run in the background as a separate thread so that the koptreader is still responsive when precaching the next page. Now it only just works. Welcome to find out bugs in it.


chrox:
The submodule libk2pdfopt has been updated. You have to do `make cleanthirdpaty && make fetchthirdpary && make thirdparty` to test this patch.

hwhw:
Very good idea. However, I would have done it this way in the first place
if it was this easy. I don't think it is. You're sharing memory between
threads and it will likely break in hard to trace, weird ways in corner
cases. E.g. the Lua state is _not_ thread safe, the stack, you name it. You
will need to add veeeery much code for locking and put real hard thought
into how you interact with the Lua state.

I don't recommend to merge this as-is. Technically, threading is a
difficult beast, especially when we're talking about shared memory and
state.

I would love to get proven wrong here, but my advice is to take these
concerns very seriously.

tigran123:
I agree with @hwhw. I also thought about moving some operations (the one I was considering was retrieving TOC as it really slows down many operations, at least the first instance of them) to a separate thread, but could not think of how to implement it safely for the reasons which @hwhw describes above.

@chrox I hope that your changes to the libk2pdfopt submodule do not imply that this PR _must_ be merged, i.e. KPV will continue to work (after fetchthirdparty) even without merging this PR.

chrox:
I add an option to disable multi-threaded precache. And I think users have the freedom to risk for better performance. Actually I have tested several thousands pages with threaded precache in the emulator and I didn't have even one segmentation fault or unexpected exit of KPV. 

I use this command to test:
`watch -n 0.5 xdotool search --classname kpdfview key Next`

It will keep pressing Page_Down on the emulator at the speed of 120 beats per minute. I even tried 600 BPM, there is no problem either.

chrox:
Probably you should test this patch first and then decide whether or not to merge it.

chrox:
@hwhw @tigran123 The only variable shared by the TWO threads is a `koptcontext` called `self.precache_kc`. It stores parameters that will initiate K2pdfopt and the source bitmap that will be processed by K2pdfopt and the pointer `data` that refers to reflowed bitmap. Problems if there are any will be from the bmp pointer that is written in K2pdfopt thread and read in the main thread. But the `precache` guard in `koptcontext` will make sure that `data` will never be read before K2pdfopt has written to it.

chrox:
With a separate thread it's possible to inform users that rendering is undergoing in the background so that he/she will not keep pressing Page_Down.

hwhw:
Huang Xin: Oh, thank you for the explanation about what is shared. Yes, in
that case it makes perfect sense, I guess. Nevertheless, I would love to
have it disabled-by-default for at least one release. I'm still very
sceptical. The koptcontext seems to be not locked against any changes and
you're working on a lot of assumptions there which might be true today, but
are not the least guaranteed by the underlying design.

chrox:
Yes. I'm pretty glad to do that.

tigran123:
Ok, I will test it. But I assume you have tested at least basic functionality on Kindle as well, not just on the emulator.

chrox:
Yes. Of course. Haven't find a problem yet.

tigran123:
Btw, thanks for teaching me about xdotool --- I didn't know this tool existed. Very handy! :)

chrox:
You are welcome. Actually I know this tool tonight too to test the threads feature. :)

PR #571, opened 2012-11-13
Don't slow down :inform() when modifying bbox (and other very minor fixes)

tigran123:
1. We shouldn't delay the calls to :inform() when modifying bbox (Shift-X) on the emulator. Otherwise, the popup window will disappear after a very short delay.
2. Document the 'msec' parameter of :inform() method in the comments above it in dialog.lua.
3. Get rid of some of the "polite" alternative voice messages in cases where the default one will do perfectly.
4. Get rid of the useless "Scanning folder..." message as this operation is fast enough to not deserve any such message.
5. Explain the reason why a directory cannot be deleted instead of just stating this as a fact. (and get rid of those exclamation marks, no need for shouting, the world is a noisy enough place as it is :)
6. Don't try to slow down the "Opening document..." message on the emulator because this would break the "open last file" method of KPV invocation.

### DIFF
--- a/dialog.lua
+++ b/dialog.lua
@@ -105,10 +105,10 @@ Adding new parameters -- 'message_importance' & 'alternative_voice_message' -- i
 Brief description of the function parameters
 -- text : is the text message for visual and (if 'alternative_voice_message' isn't defined) voice notification
 -- msec : parameter to define visual notification method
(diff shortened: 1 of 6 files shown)


kai771:
You seem to have forgotten koptconfig.lua.

tigran123:
Yes, you are right. Thank you, I will do that now.

tigran123:
Merging my own PR before it becomes stale... :)

PR #574, opened 2012-11-13
bugfixes for fit-to-content-width zoom mode for documents with pages of ...

kai771:
...different size

Fixes 2 bugs in fit-to-content-width mode with documents that have pages of different size (e.g. manga with spreads)
1. if height of the page is less than hight of the screen, the page should be centered.
2. when pressing pg_bk, zoom mode should be recalculated.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1332,9 +1332,7 @@ function UniReader:setzoom(page, preCache)
 			-- We must handle previous page turn as a special cases,
 			-- because we want to arrive at the bottom of previous page.
 			-- Since this a real page turn, we need to recalculate stuff.
---			if (x1 - x0) < pwidth then
-				self.globalzoom = width / (x1 - x0)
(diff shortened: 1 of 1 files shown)


tigran123:
In the function UniReader:setzoom() there are local copies of G_width and G_height stored in the local variables width and height respectively. Why not use those?

kai771:
G_height first came to mind. Fixed.

tigran123:
Thank you, I will test this PR tomorrow.

tigran123:
@kai771 When you change the code next time, please explain it in the commit message rather than (or in addition to) in the PR. The explanation in the PR will be forgotten but the commit message is part of the repository and will remain forever, regardless of internet access to github.com. Offline data is always more valuable than online.

Btw, today is KPV's birthday (according to commit log). It is 1 year old baby today :)

Issue #575, opened 2012-11-13
one page away jump to link in pdf should work with Back key


dracodoc:
UPDATE: It's the problem of the pdf, not KPV. But for the special case of jump to link in next/previous page, Back key should bring user to previous page, currently it only works on links to several pages away.

I found some random underlines were showed in one pdf. 

The pdf is here
https://www.box.com/s/sohk5dih04gsmg2tyw7j

screenshot
https://www.box.com/s/f7j9b5zzz8fob0h5fuc5
https://www.box.com/s/ds65siqdx14oryo184vj

I guess these lines are actually underlined links because there are some links in this page, but the line position is off to the link position, probably because I cropped the pdf in pc.

So the issue is the link position need to recognize crop box to get the updated position

tigran123:
This is not a bug, it is a feature. Those locations with underlines actually have local links. E.g. the one under the word "selected" in your screenshot has a jump to page 27 (which is in References). If you press `L` then you will see the shortcuts corresponding to each link and pressing that shortcut key will bring you to the target destination of that local link.

Now, if you don't want to see any such underlines then you can just press `Shift-L` to disable showing them. The setting is saved for the current book only.

Btw, if you open that PDF on a PC (e.g. in evince or Adobe Reader) and hover the mouse over those words you will notice the mouse pointer change shape to indicate there is a link there.

Update: I just noticed the end of your report (was it there before? :) and see that you already know the reason. Ok, concerning the link position I see that KPV calculates it correctly (well, at least as accurate as evince and Adobe Reader), so there is no need (or in fact even a chance) to change anything. If you hover the mouse over the link and move it horizontally carefully, you will get an idea of the actual link location and see that KPV matches it perfectly.

Ok, there is a possibility that you are right and it is a bug SHARED by all three: KPV, evince (i.e. poppler) and Adobe Acrobat Reader. To decide this one would need to look closely at the way link coordinates interact with crop boxes and check PDF specification. But I seriously doubt it....

dracodoc:
I realized these lines are for the links in pdf. But they are not in the position they should be. The underline should appear under the link text, i.e. the one under "selected" is actually the link of [21-26] two lines above it. All these lines are from the links but in a shifted position.

One interesting thing is I cannot click these links in acrobat, so it seemed the pdf have something wrong so the links are not recognized by acrobat but actually recognized by KPV!

[...]

PR #577, opened 2012-11-13
Add any link jump to the history.

tigran123:
If you follow a local link in the PDF file which points to a place only one page away from the current page and then press Back it will bring you to some random location (depending on the previous jump history). But normally you would expect to be returned to the link's location, regardless of its distance from the target page. This PR fixes this problem, also raised as an issue #575.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1579,14 +1579,16 @@ function UniReader:addBookmark(pageno)
 end
 
 -- change current page and cache next page after rendering
-function UniReader:goto(no, is_ignore_jump)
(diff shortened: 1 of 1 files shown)


tigran123:
Merging my own PR, before it grows cold ;)

houqp:
Sorry that I was too busy last week to come and merge the RPs ;P

tigran123:
Yesterday I was explaining the github.com development model to my wife and realized that PR can stand not only for "pull request" but also for "peer review" :)

PR #578, opened 2012-11-13
Optimize UniReader:redrawCurrentPage()

tigran123:
1. There is no need to call the :goto() method from `UniReader:redrawCurrentPage()` which does the extra work of adding to jump history and (most importantly, as the jump history is ignored in this case anyway) pre-caching, which is not relevant for the simple task of redrawing the current page.
2. The page goto `G` function can now be optimized to avoid pre-caching if the page has not actually changed.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1633,7 +1633,7 @@ function UniReader:goto(no, is_ignore_jump)
 end
 
 function UniReader:redrawCurrentPage()
-	self:goto(self.pageno)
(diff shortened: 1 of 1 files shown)


tigran123:
Any objections to this change?

chrox:
No. I can remove the redrawWithoutCache in koptreader after this patch been merged.

tigran123:
Good, go ahead now :)

PR #580, opened 2012-11-14
Adds zip ext to PDFReader and makes extensions accesible through defaults.lua

kai771:
A lot of manga comes as .zip, and not .cbz (for all intents and purposes, it
is the same thing, just the extension is different). This allows the user to
choose the reader to open her zip files - PDFReader or CREReader, in the same
way she can choose to use PDFReader or PDFReflow for pdf files. I also moved
the list of supported extensions to defaults.lua.

### DIFF
--- a/readerchooser.lua
+++ b/readerchooser.lua
@@ -7,6 +7,7 @@ require "djvureader"
 require "koptreader"
 require "picviewer"
 require "crereader"
+require "defaults"
 
 registry = {
 	-- registry format:
--- a/readerchooser.lua
+++ b/readerchooser.lua
@@ -18,7 +18,6 @@ registry = {
 	DJVUReflow = {KOPTReader, DDJVUREFLOW_EXT, 2},
 	CREReader  = {CREReader, DCREREADER_EXT, 2},
 	PICViewer = {PICViewer, DPICVIEWER_EXT, 1},
-	-- seems to accept pdb-files for PalmDoc only
 }
 
 ReaderChooser = {
--- a/readerchooser.lua
+++ b/readerchooser.lua
@@ -11,12 +11,12 @@ require "crereader"
 registry = {
 	-- registry format:
 	-- reader_name = {reader_object, supported_formats, priority}
-	PDFReader  = {PDFReader, ";pdf;xps;cbz;", 1},
-	DJVUReader = {DJVUReader, ";djvu;", 1},
-	PDFReflow = {KOPTReader, ";pdf;", 2},
-	DJVUReflow = {KOPTReader, ";djvu;", 2},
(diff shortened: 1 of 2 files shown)


tigran123:
I don't see how you have overcome the conflict with crengine. Please explain how did you manage to make them co-exist as I don't see any detection code in your patch.

The .zip extension is already taken by crengine in order to support .fb2.zip files (btw, the only format that crengine supports properly --- all others like epub/mobi is on "toy level" and don't really work).

tigran123:
ah its done by readerchooser ok...

kai771:
It just works :)

tigran123:
yes i got it now :)

tigran123:
And if you feed an .fb2.zip or some corrupt .zip to PDFReader --- what would happen then? A crash of KPV?

kai771:
Hm, good point. Forgot to test that. Be right back :)

kai771:
It freezes with "Opening document..." in emulator. However, you can press home and continue like nothing happened. Not ideal, but I don't think it's that bad either.

tigran123:
To get access to those D\* defaults in defaults.lua you are relying on the fact that readerchooser.lua requires koptreader.lua which in turn requires unireader.lua which requires defaults.lua. If any link in this chain breaks then those D\* would be inaccessible, so it seems a good idea to require "defaults" explicitly in readerchooser.lua as it needs something from defaults.lua directly.

kai771:
Yes, you're quite right. I forgot to add requires defaults in readerchooser.lua.

readerchooser.lua
```diff
@@ -7,16 +7,17 @@ require "djvureader"
...
 	-- registry format:
 	-- reader_name = {reader_object, supported_formats, priority}
-	PDFReader  = {PDFReader, ";pdf;xps;cbz;", 1},
-	DJVUReader = {DJVUReader, ";djvu;", 1},
-	PDFReflow = {KOPTReader, ";pdf;", 2},
-	DJVUReflow = {KOPTReader, ";djvu;", 2},
-	CREReader  = {CREReader, ";epub;txt;rtf;htm;html;mobi;prc;azw;fb2;chm;pdb;doc;tcr;zip;", 1},
-	PICViewer = {PICViewer, ";jpg;jpeg;", 1},
+	PDFReader  = {PDFReader, DPDFREADER_EXT, 1},
+	DJVUReader = {DJVUReader, DDJVUREADER_EXT, 1},
```
tigran123, line 16: This comment now belongs to the end of line 76 in defaults.lua and not here. I mean the comment about pdb and PalmDoc.

tigran123:
If you move that comment to the end of line 76 in defaults.lua I see no reason not to merge this PR.

kai771:
Sure, just a sec.

PR #581, opened 2012-11-14
Bugfix: we still need those calls to self:redrawPage() in toggles, unfortunately...

tigran123:
I have removed the calls to self:redrawPage() recently after the calls to InfoMessage:inform() with DINFO_TIMEOUT_FAST because on the emulator the screen was refreshed automatically. But not so on the real Kindle hardware --- we have to refresh the page explicitly, otherwise the popup window stays until one forces manual refresh via Space or turns the page (or causes a refresh by some other way, e.g. pressing Menu twice). This patch restores those calls to self:redrawCurrentPage(). Also, it adds space to the end of some messages for symmetry and switches Alt-Z toggle to DINFO_TIMEOUT_FAST mode as well.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -2760,9 +2760,8 @@ function UniReader:addAllCommands()
 		"toggle battery level logging",
 		function(unireader)
 			G_battery_logging = not G_battery_logging
-			InfoMessage:inform("Battery logging "..(G_battery_logging and "ON " or "OFF "), DINFO_TIMEOUT_FAST, 1, MSG_AUX)
+			InfoMessage:inform("Battery logging "..(G_battery_logging and "ON" or "OFF"), DINFO_TIMEOUT_FAST, 1, MSG_AUX)
 			G_reader_settings:saveSetting("G_battery_logging", G_battery_logging)
-			self:redrawCurrentPage()
 		end)
 	self.commands:add(KEY_T,nil,"T",
 		"show table of content (TOC)",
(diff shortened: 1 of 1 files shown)


tigran123:
Oops, a few of the DINFO_TIMEOUT_FAST instances was changed to DINFO_TIMEOUT_SLOW by mistake. Will fix in a minute...

tigran123:
Actually, maybe we should change all toggles to DINFO_TIMEOUT_SLOW instead? Then we won't need all those calls to self:redrawCurrentPage()? Opinions are welcome.

kai771:
I think that would make toggle info messages stay for too long. It's would be annoying to me.

tigran123:
Try pressing Alt-Z on Kindle now. It uses DINFO_TIMEOUT_SLOW and it doesn't seem to be too long. Also, you can terminate the delay by pressing any key, e.g. Space.

kai771:
I'm aware of the fact that you can press any key to shorten the message, but it doesn't eliminate the fact that it's annoying. Why are you so against calling redraw once again? Trying to save a few sec of batteries?

tigran123:
Ok, I'll try to figure out _why_ that redraw is needed on the Kindle, but not on the emulator. This is the proper way of doing things instead of curing the symptom.

You see, if the infrastructure somehow saves/restores the part of the screen dirtied by the popup window then why should we redraw the whole page? Some instances (e.g. toggling link underlines or overlaps) do require such a redraw, but toggling the battery logging (and a few others) certainly can't change anything on the current page, so redrawing it is the wong thing to do.

kai771:
I'm always for making the emulator resemble real Kindle as much as possible. Btw, there might be a simpler solution that would satisfy us both - why not make `DINFO_TIMEOUT_FAST = 1`? It should still make it fast on Kindle, and you shouldn't need to call redraw. (for emu, just change the part that says if msec == 0 to msec == 1). I didn't test this, but I think it should work.

tigran123:
I am closing this PR  so as to investigate it properly in a separate branch, not my local master.
And yes, I will test your idea with `DINFO_TIMEOUT_FAST=1` as well, thank you.

chrox:
I use cpulimit to slow down the emulator to make it resemble real kindle device.

Maybe you should try this:
`cpulimit -p `pidof kpdfview` -l 10`

It will allocate 10 percent of cpu cycles for kpdfview. Now the emulator will run on a 200MHz CPU which feels like a real Kindle.

kai771:
@chrox Good idea, thank you.

tigran123:
@kai771 Yes, your idea of setting `DINFO_TIMEOUT_FAST=1` works fine, thank you. And now we can also get rid of setting pagedirty flag in filechooser for those commands.

PR #582, opened 2012-11-14
Changes keybindings for zoom modes so zoom to content is used without Shift

kai771:
This will probably not be accepted, but I thought I should try it anyway. I think that, due to the small size of Kindle screen, a user is MUCH more likely to use zoom to content modes compared to zoom to page modes. Therefore, I think zoom to content modes should be the ones she gets without pressing Shift. I know that some might be used to it, but I think it won't take them long to adjust to better. :) What do you guys think?

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -2696,42 +2696,42 @@ function UniReader:addAllCommands()
 		end)
 	-- to leave or to erase 8 hotkeys switching zoom-mode directly?
 
-	self.commands:add(KEY_A,nil,"A",
+	self.commands:add(KEY_A,MOD_SHIFT,"A",
 		"zoom to fit page",
 		function(unireader)
 			unireader:setglobalzoom_mode(unireader.ZOOM_FIT_TO_PAGE)
 		end)
-	self.commands:add(KEY_A,MOD_SHIFT,"A",
+	self.commands:add(KEY_A,nil,"A",
 		"zoom to fit content",
 		function(unireader)
 			unireader:setglobalzoom_mode(unireader.ZOOM_FIT_TO_CONTENT)
 		end)
-	self.commands:add(KEY_S,nil,"S",
+	self.commands:add(KEY_S,MOD_SHIFT,"S",
 		"zoom to fit page width",
 		function(unireader)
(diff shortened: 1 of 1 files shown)


tigran123:
What about dozens of messages I have written on various forums explaining how to use "Shift-S mode" and "Shift-F mode", even saying that all other zoom modes are useless and can be safely thrown away. People rely on those explanations as a kind of "user's guide"...

You are right, this PR probably will not be accepted, for compatibility reasons.

But I totally agree that zoom to content is far more useful than zoom to page. In fact, only Shift-S and Shift-F zoom modes have value. The rest are junk.

kai771:
Well, they're rarely used, but I wouldn't call them "junk". As for explanations, you could update major ones, and we really should write a proper manual and put it on our wiki. We could also write with big letters in our release notes that we changed it (maybe even make a pop-up on first start of new version). It will be a bit confusing at first, but I don't think it will take a long time to sort it out. And although you refered to it as "Shift-S" mode it was never officially called that - it was (and is) called "Fit zoom to content width".

tigran123:
If we do this, I would also suggest to take this as an opportunity to clean up and get rid of any modes that are truly and utterly useless. And also get rid of the zoom mode menu `M` because it doesn't ever seem to get things right anyway (because of some built-in "precautions" restricting access to only "real modes" etc).

kai771:
I totally agree. I'd really like to see Z, Alt-Z and Shift-Z gone. I never understood what they really do (does anyone?) and how they are useful now that we have modify page bbox. Does anyone use that?

(In my guide [here](http://www.mobileread.com/forums/showthread.php?p=2217835#post2217835) I called them "relics of the past". Was I wrong?)

tigran123:
Alt-Z is easy to understand --- it enables and disables bbox. So Shift-S would work like S if bbox is disabled. But I agree it is useless. And the other two I never understood (I thought I did after @dpavlin explained them, but in a couple of hours I forgot the explanation (only remembered that they were needed _before_ Shift-X was introduced) and am now  back to where I was before, i.e. in total agreement with you that they should be gone :)

tigran123:
This feature should properly be implemented in a more generic way as user-defined custom key bindings.

kai771:
Even if they're implemented as user-defined custom key bindings, they could still default to those suggested here. Just saying.

PR #583, opened 2012-11-14
bugfix: wait only when the bgthread is running

chrox:
When the bgthread has finished, do the fgthread's reflow.


tigran123:
Seems to work.

PR #584, opened 2012-11-14
Bugfix: toggle popup windows don't disappear on Kindle

tigran123:
Thanks to @kai771 's idea of setting the value of `DINFO_TIMEOUT_FAST=1` (in defaults.lua) we can still have the same behaviour of calls to InfoMessage:inform(..DINFO_TIMEOUT_FAST..) on both the emulator and the Kindle without the need to manually refresh the page after the call.

Also, now we don't need to set pagedirty=true in filechooser when displaying a popup window with DINFO_TIMEOUT_FAST flag.

Also, add space to the end of some text messages passed to :inform() for symmetry.


tigran123:
Hmmm, there is a conflict now, so I'll have to re-do this in a separate branch. Closing this PR.

PR #585, opened 2012-11-14
Bugfix: toggle popup windows don't disappear

tigran123:
Thanks to @kai771 's idea of setting the value of `DINFO_TIMEOUT_FAST=1` (in defaults.lua) we can still have the same behaviour of calls to InfoMessage:inform(..DINFO_TIMEOUT_FAST..) on both the emulator and the Kindle without the need to manually refresh the page after the call. 

Also, now we don't need to set pagedirty=true in filechooser when displaying a popup window with DINFO_TIMEOUT_FAST flag.

Also, add space to the end of the text messages passed to :inform() for symmetry.

### DIFF
--- a/defaults.lua
+++ b/defaults.lua
@@ -40,7 +40,7 @@ DBATTERY_LOGGING = false
 DBACKGROUND_COLOR = 8
 
 -- timeout for info messages in ms
-DINFO_TIMEOUT_FAST=0
+DINFO_TIMEOUT_FAST=1
 DINFO_TIMEOUT_SLOW=1500
 
 -- toggle defaults
@@ -75,4 +75,3 @@ DPDFREFLOW_EXT = ";pdf;"
 DDJVUREFLOW_EXT = ";djvu;"
 DCREREADER_EXT = ";epub;txt;rtf;htm;html;mobi;prc;azw;fb2;chm;pdb;doc;tcr;zip;" 	-- seems to accept pdb-files for PalmDoc only
 DPICVIEWER_EXT = ";jpg;jpeg;"
-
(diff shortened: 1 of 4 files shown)


tigran123:
Forgive me for merging my own PR but otherwise I will have to keep re-doing it because of the "commit+revert" pair which I foolishly put into the master. If someone disagrees with the idea of setting DINFO_TIMEOUT_FAST=1 then the change can be undone in the subsequent commit.

tigran123:
I had to fork a new instance of kindlepdfviewer because my instance got into trouble...

kai771:
Ok, I probably shouldn't be giving advice, being a noob and all, but here it goes... never change your master, no matter how small the changes are. Always do changes in a new branch, and let your master be only a copy of hwhw master. That way anything you change could be easily tested by anyone, and that way you'll never have to do that rebase you were mentioning before. :)

tigran123:
Yes, that is exactly the same advice I was always giving to others, but this time I forgot to follow it myself :)

tigran123:
Btw, the approach of this PR doesn't work in all circumstances, unfortunately. E.g. when pressing Alt-S to save attachments in a PDF (i.e. inform(FAST) followed by inform(SLOW)) it requires an extra key-press when it shouldn't.

So, we still need to understand how InfoMessage:show() fundamentally differs from showInfoMsgWithDelay(). Looking at the code I see that the latter has save/restore BB sequence whereas the former does not. Ok, tomorrow I'll think how to make a solution that works everywhere (both on Kindle and emulator).

tigran123:
Yes, that saving/restoring BB is the key to everything. My next coming PR will fix this problem properly and forever (I hope :)

PR #587, opened 2012-11-15
Cleanup InfoMessage:inform() API and its usage

tigran123:
Ok, this is the Final Solution, hopefully :)
As Linus says, how many times we can get it wrong, really --- we should have run out of wrong ideas by now :)

This patch changes the second argument of InfoMessage:inform() function to be either DINFO_DELAY or DINFO_NODELAY with the following semantics:

DINFO_DELAY - display the message and delay DINFO_DELAY ms before returning. Save and restore the screen content. This is useful when the message is only conveying information and nothing has changed on the screen. For example, this is ideal for informing the user about options toggled on/off. I repeat: delaying after the display of toggle message is THE RIGHT THING TO DO (I didn't understand this until now, hence all those earlier unnecessary commits...)

DINFO_NODELAY - display the message and return immediately without saving/restoring the screen content. This is useful when there is work after the call that will change the screen content anyway and will have to take care of it. Note that if you follow a DINFO_NODELAY call with a DINFO_DELAY one then you have to manually refresh the screen (see the usage in Alt-S command handler in pdfreader.lua)

On the emulator DINFO_NODELAY is delayed by a small amount in order to simulate Kindle's behaviour (and make the message noticeable). But the semantics of DINFO_NODELAY (notably the NOT saving/restoring screen content) is exactly the same on the emulator and on Kindle, so no extra "redraw page" calls are needed specifically for Kindle but not on emulator.

The "ON" and "OFF" were changed to "on" and "off" and some other words lowercased to make it easier on the TTS to pronounce them (otherwise it thinks they are abbreviations). Of course I know that we could have passed a separate tts_text argument but there is nothing wrong with the lowercase "on" and "off" (I like them better) so I didn't do that.

### DIFF
--- a/commands.lua
+++ b/commands.lua
@@ -170,7 +170,7 @@ function Commands:new(obj)
 			if G_charging_mode == false and G_screen_saver_mode == false then
 				if G_battery_logging then logBatteryLevel("SLEEP") end
 				Screen:saveCurrentBB()
-				InfoMessage:inform("Going into screensaver... ", DINFO_TIMEOUT_FAST, 0, MSG_AUX)
+				InfoMessage:inform("Going into screensaver... ", DINFO_NODELAY, 0, MSG_AUX)
 				Screen.kpv_rotation_mode = Screen.cur_rotation_mode
 				fb:setOrientation(Screen.native_rotation_mode)
 				util.sleep(1)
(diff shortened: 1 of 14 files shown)


kai771:
Sigh... I don't want to argue about this, so I'm just gonna state my opinion once again, and let it rest. I agree that messages should be seen, otherwise we shouldn't even display tham. What I do not agree is that toggles should last as long as other messages. They ended up as nil because people complained about them being too long, and someone took advantage of the slowness of kindle's screen (I vaguely remember reading about it, I think dracodoc pointed out that even setting them to 0 doesn't make them last 0. If I remember correctly, it was something about DX being even slower, so it was better to set them to nil than to set them to say 100ms, because even that 100ms would slow it down even further). Of course, that was too fast for the emulator, so that's why we introduced a check and changing that value for the emulator. If at all possible, leave the distiction between slow and fast. 1.5 sec just for toggle is annoying. Make it slow, fast and no delay. Please.

tigran123:
Ok, your opinion noted, but I understand that if people complained about toggle messages being too slow it was probably because there was a subsequent unnecessary page redraw or simply because the whole screen redraw was taking too long and the users mistakingly attributed it to the toggle info message.

tigran123:
closing this PR as it conflicts with merged changes. Will prepare a new one in a minute...

PR #588, opened 2012-11-15
refactoring: move function in "Shift+X" command into the modBBox method

chrox:
So that other modules like koptconfig could reuse the code to have the functionality of modifying the page bbox.


tigran123:
Very good idea, but conflicting with my PR #587. But I don't want to be merging my own PRs anymore --- I am not playing a dictator (unless there is an emergency like yesterday :)

tigran123:
Ok, your other PR merged OK, but this one will definitely conflict with mine. What shall we do? Merge this one and then I would have to redo #587 or merge #587 and you will redo this one?

tigran123:
Ok, I'll merge this one and will redo #587.

tigran123:
Hmmm, have you only tested Shift-X in koptreader? It seems broken in the standard reader. I'll try to find out why.

tigran123:
Oh, I should have reviewed it before merging: the call `self:modBBox(unireader)` should be `unireader:modBBox()`.
And why do you need to pass koptreader as a variable from koptconfig? There is only one KOPTReader object and it can be accessed from anywhere. Oh, you are doing that with `self.koptreader:redrawWithoutPrecache()` as well. Why? Please explain. That `self.koptreader` thing seems unnecessary.

tigran123:
Ok, the Shift-X in the standard reader works fine, but still all that passing of reader objects as variable seems unnecessary.

chrox:
The unnecessary parameter is removed in the patch in #591. I'm sorry that I forgot to remove parameter of the caller when I removed the parameter from the method definition. As for passing reader object in koptconfig, I just feel uncomfortable to use global variable, so I would like to pass reader object explicitly. Maybe that's why I'm not a good Lua programmer.

tigran123:
Ok, it is up to you, but please be advised that it is a bad idea because: a) it is slower and b) it makes the code less readable because it suggests that some alternative reader could go into those calls in koptconfig whereas in reality the only valid choice would be KOPTReader --- nothing else.

There is nothing wrong with using global variables properly.

chrox:
Thanks for your advice.

PR #589, opened 2012-11-15
add function to log page reflow duration

chrox:


tigran123:
Having a ".txt" file as a log is a very good idea, so it can be viewed from within KPV (that is why I did the same for battery logging).

Btw, if I merge this PR now then my own PR will no longer apply as it modifies the call to :inform() which is visible within the context of your changes.

tigran123:
Actually, I just tried it and it merged both your and my changes successfully. So, I'll merge your PR and see what happens --- mine should still be mergeable after it.

PR #590, opened 2012-11-15
Cleanup InfoMessage:inform() API and its usage

tigran123:
This patch changes the second argument of InfoMessage:inform() function to be either DINFO_DELAY or DINFO_NODELAY with the following semantics:

DINFO_DELAY - display the message and delay DINFO_DELAY ms before returning. Save and restore the screen content. This is useful when the message is only conveying information and nothing has changed on the screen. For example, this is ideal for informing the user about options toggled on/off. I repeat: delaying after the display of toggle message is THE RIGHT THING TO DO (I didn't understand this until now, hence all those earlier unnecessary commits...)

DINFO_NODELAY - display the message and return immediately without saving/restoring the screen content.This is useful when there is work after the call that will change the screen content anyway and will have to take care of it. Note that if you follow a DINFO_NODELAY call with a DINFO_DELAY one then you have to manually refresh the screen (see the usage in Alt-S command handler in pdfreader.lua)

On the emulator DINFO_NODELAY is delayed by a small amount in order to simulate Kindle's behaviour (and make the message noticeable). But the semantics of DINFO_NODELAY (notably the NOT saving/restoring screen content) is exactly the same on the emulator and on Kindle, so no extra "redraw page" calls are needed specifically for Kindle but not on emulator.

Also, tidy up the call to modBBox() from Shift-X command handler, it should be unireader:modBBox() and not self:modBBox(unireader) --- there is no need to pass the reader object to the :modBBox() method.

### DIFF
--- a/commands.lua
+++ b/commands.lua
@@ -170,7 +170,7 @@ function Commands:new(obj)
 			if G_charging_mode == false and G_screen_saver_mode == false then
 				if G_battery_logging then logBatteryLevel("SLEEP") end
 				Screen:saveCurrentBB()
-				InfoMessage:inform("Going into screensaver... ", DINFO_TIMEOUT_FAST, 0, MSG_AUX)
+				InfoMessage:inform("Going into screensaver... ", DINFO_NODELAY, 0, MSG_AUX)
 				Screen.kpv_rotation_mode = Screen.cur_rotation_mode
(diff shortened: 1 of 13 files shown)


tigran123:
Ok, merging this one as it: a) works and b) see a)

PR #591, opened 2012-11-15
cleanup: remove unnecessary parameter when calling modBBox

chrox:


tigran123:
Ah, I already included this in my PR #590 and I am not redoing that one again.

I was talking about the usage in koptconfig --- i.e. why do you need to pass self.koptreader there? Why not refer to KOPTReader object? It saves you table dereferences and generally makes the code more readable.

tigran123:
Oops, forgot to close this PR so that it is not accidentally merged causing me to have to redo my PR.

PR #593, opened 2012-11-15
bugfix: bottom of bbox should be respected in fit-to-content-width mode

kai771:
Personally, I always set just left and right, but some people would probably find it broken if I didn't fix this :)


tigran123:
Tested, works fine, thank you.

PR #594, opened 2012-11-15
Overlap on/off improvement

kai771:
Turning off overlap was instantly visible, but turning it back on was visible only on the next view. This makes both turning on/off overlap instantly visible. Since the result of a keypress is immediately visible, I commented out the inform message, since it's not needed anymore.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -2078,6 +2078,7 @@ function UniReader:showToc()
 				InfoMessage:inform("External links unsupported ", DINFO_DELAY, 1, MSG_WARN)
 				toc_curitem = ret_code
 			else
+				self.show_overlap = 0
 				return self:gotoTocEntry(toc_entry)
 			end
 		elseif item_no then -- expand or collapse item
@@ -2130,6 +2131,7 @@ function UniReader:showJumpHist()
 		if item_no and item_no <= #self.jump_history then
 			local jump_item = self.jump_history[item_no]
 			self.jump_history.cur = item_no
+			self.show_overlap = 0
 			self:goto(jump_item.page, true)
 			-- set new head if we reached the top of backward stack
 			if self.jump_history.cur == #self.jump_history then
@@ -2161,6 +2163,7 @@ function UniReader:showBookMarks()
 		}
 		ret_code, item_no = bm_menu:choose(0, fb.bb:getHeight())
 		if ret_code then -- normal item selection
+			self.show_overlap = 0
 			return self:goto(self.bookmarks[ret_code].page)
 		elseif item_no then -- delete item
 			table.remove(menu_items, item_no)
@@ -2568,6 +2571,7 @@ function UniReader:addAllCommands()
 	self.commands:addGroup(MOD_ALT.."H/J", {Keydef:new(KEY_H,MOD_ALT), Keydef:new(KEY_J,MOD_ALT)},
 		"go to prev/next TOC entry",
 		function(unireader,keydef)
+			unireader.show_overlap = 0
 			if keydef.keycode == KEY_H then
 				self:gotoPrevNextTocEntry(-1)
 			else
@@ -2580,6 +2584,7 @@ function UniReader:addAllCommands()
 		Keydef:new(KEY_PGFWD,nil),Keydef:new(KEY_LPGFWD,nil)},
 		"previous/next page",
 		function(unireader,keydef)
+			unireader.show_overlap = 0
 			unireader:goto(
 			(keydef.keycode == KEY_PGBCK or keydef.keycode == KEY_LPGBCK)
 			and unireader:prevView() or unireader:nextView())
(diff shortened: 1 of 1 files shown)


tigran123:
Hmmm, but the result is not always visible, i.e. not in every possible situation, right? So the info message would still be needed in those cases. E.g. when one enables overlap in the modes where there can be no overlap (like `A`), simply because he does this first and then switches to `Shift-S` mode. There would be no way of finding out whether overlap is turned on or off.

Actually, what I really wanted is to show all these things in the Menu status info display. I.e. all the toggles and any other useful info. So that one doesn't need to discover them by actually toggling the things.

kai771:
It seems I missed a lot of things in my initial PR. Needs to be tested. I'm counting on you, @tigran123 :).

kai771:
@tigran123 I agree with you that a menu status would be nice. As for figuring out if overlap is on or off - pressing next/prev page will make that clear. I could enable the info message if you consider it necessary, but let me explain a use case that I find likely - a user doesn't like overlap, but occasionally she may turn it on for a sec to see where she were before page turn. Info message will slow down this operation considerably, and make it impractical (ok, she could press a button to end Info message, but then she would need to press it 4 times repeatedly for a quick check - I don't think it's really convenient)

About other modes - overlap is only enabled in "smart" modes - fit-to-content-width and 2 column mode. It doesn't matter if it's on or off if the user is not in one of those modes, and when she uses a smart mode, it will be obvious.

tigran123:
Yes, that usage pattern is what I also tried to use it for, but without this PR it was impossible (because one would have to press next to make it into effect). So, yes, don't just comment that :inform() call --- delete it entirely.

tigran123:
Thank you, I will test it now.

tigran123:
Btw, instead of putting show_overlap = 0 code into so many places, wouldn't it be better to put it into those few functions that are called at those places, like setGlobalZoom(), gotoPrevNextTocEntry(), setglobalzoom_mode()?

Also, I suspect the placement of show_overlap = 0 is wrong in the Back command handler (i.e. it should be only prior to goto() but not on "Already first jump" path), but I haven't tested it yet --- just reviewing the code.

Update: yes, confirmed the above bug in "Back" handler. Also, a similar bug in "shift-Back" handler. I.e. if you press Back or Shift-Back in the situation when you get "Already first/last jump" then overlap on/off toggle doesn't work. Shifting "show_overlap = 0" code to just prior to :goto() solves the problem.

And another bug: in the Alt-K/L handler the overlap=0 should only be done if there are bookmarks. Otherwise pressing Alt-K or Alt-L when there are no bookmarks destroys the overlap on/off behaviour. That means just before :goto().

I just noticed that it cannot be put INSIDE :goto() because there are calls (from fiveway handler) to :goto() which shouldn't set show_overlap = 0.

And yet another bug: if there is no TOC then pressing Alt-H or Alt-J kills the overlap on/off behaviour. This is because the setting of show_overlap = 0 should be INSIDE gotoPrevNextTocEntry() function and only in the code that will lead to a :goto(), i.e. after ascertaining that there is a valid TOC.

Also, I don't understand why show_overlap = 0 is present in Alt-Back handler but NOT in the Home handler. Please explain. At the moment (despite the comments and the help page) both handlers work exactly the same, i.e. "keep_running" is a dummy global variable that is not actually used anywhere (the one in inputLoop() is local, so it shadows the value in the global).

Also, you set show_overlap = 0 before every call to setglobalzoom_mode() except the one reached via `M` menu. Is it the subconscious desire (present in me also :) to see the `M` mode gone? I think it should be _inside_ the setglobalzoom_mode() function.

kai771:
Thanks a lot for testing. As seen in my original PR, I expected it to be a lot simpler :). I'll try to fix it when I get back home. I'll be counting on you for testing it again then :) (yes, I do test it myself too, but it seems that I often miss a lot :( )

tigran123:
Ok, good.

tigran123:
I think instead of setting show_overlap = 0 in Alt-Back/Home handlers you should set it in inputLoop() where the traces of previous document are erased to not affect the next document, i.e. under the comment "do clean up stuff".

kai771:
@tigran123 There, I implemented the changes you suggested. Please check if I missed something (again). Sorry about the mess - I expected it to be much simpler. When I tried initial PR, it seemed to work, so I submited it. I rarely use bookmarks, TOC etc, so I forgot to test that. When I realised that it doesn't work completely, I tried to do a quick fix, because I didn't have much time to work on it. As can be seen, it was really sloppy. Thanks for testing and all your suggestions. I tested this one and it seems to work fine. However, for some reason, it seems that Alt is not working for me in the emulator on my emulated Ubuntu Linux, so I couldn't test that. If you have any idea why my Alt key isn't working in the emulator (it works just fine in emulated Ubuntu, just not in KPV emulator), please let me know. 
It is quite possible that I missed something again in this commit, so... be advised :)

tigran123:
Ok, thank you, I'll have a look.

All is good.

tigran123:
When you press Alt key (with "reader.lua -d ." running) you should get message like this in the terminal:

```
# key pressed: _Alt_+[64]
```

Do you get them?

kai771:
@tigran123 No, but when I let go of it I get

```
# ignored ev  {
    ["type"] = 1,
    ["code"] = 64,
    ["value"] = 0
```

Any idea what's wrong?

kai771:
I found the solution. The problem was Ubutnu's Unity. One needs to disable HUD shortcut key in _System Settings/Keyboard/Shortcut/Launchers/Key to show the HUD_ - or change it from _Alt L_ to something else.

PR #595, opened 2012-11-16
some minor tweaks before introducing gesture detection module

houqp:
- refine single tap touch input simulation in input.c
- update document.lua according to API change in cre.cpp
- add time subtable to lua inputevent table
- add back waitForInput timeout for emulator

### DIFF
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -34,7 +34,7 @@ function CreDocument:engineInit()
 			if _v ~= "Dingbats.cff" and _v ~= "StandardSymL.cff" then
 				local ok, err = pcall(cre.registerFont, Font.fontdir..'/'.._v)
 				if not ok then
-					Debug(err)
+					DEBUG(err)
 				end
 			end
 		end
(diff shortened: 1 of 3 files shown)


chrox:
@houqp I would like to get a Kindle Touch to test the new UI since Kindle Paperwhite is so hard to get in China. BTW, is the basic functionality of the new UI usable in Kindle Touch?

houqp:
No, it's not usable yet because zoom mode is not implemented. So it is only usable for book supported by CREngine. You can indeed test it first in emulator to see how simple it is now ;p

PW is out of stock in US too, crazy amazon.

chrox:
Thanks. Anyway I will get involved in the new UI development as soon as I get my KT.

houqp:
That will be awesome :) Let me know if you need to buy PW, I can send you one, but it will need sometime before I can get it from Amazon ;p

chrox:
Thanks. I have got one from taobao. Now the KP is already jailbroken and usbnetworked. And I will compile the new UI code for it now. :)

houqp:
merging for now since no objections ;p

houqp:
WOW, remember to keep airplane mode on ;p

Issue #573, opened 2012-11-13
making rotate screen mode a saved configuration specific to book


dracodoc:
I found that when I rotate screen 90 degree for a book, pressing home will actually close book but the display is not updated correctly. Five way button will move a long landscape orientation selection line above the current book page, but not really usable. Another pressing Home will quit KPV.

So it seemed pressing home in rotated mode is not processed correctly. I think KPV should rotate screen back to normal and return to file browser when Home is pressed, since operating file browser in landscaping mode is not convenient and pointless. .  

Besides, every time user open a book suitable for landscape reading, he need to rotate screen manually. So I suggest to make rotated screen as a a saved configuration specific to book
1. When user rotated screen to read book A, directly closing book will return to regular portrait mode file browser.
2. The rotated screen configuration(include screen orientation and zoom mode) is saved in book A history like other history information, so next time when user open book A, the screen will rotate to last reading mode automatically, user can start read immediately without any interference. 
3. The usual bookmark or last reading position may be based on page number only, but landscaping mode with fit width view may need to save more detailed last reading position. This involves more effort but it should be done, otherwise user need to flip through to find the exact last reading position. I believe this kind of detailed reading position is also saved already in fit content width mode.

Whether to rotate screen to read is always specific to each book, with this design, each book only need to be manually rotated for one time if needed, then user will never need to adjust screen orientation manually again.

tigran123:
I think we should fix the bug reported earlier: filechooser does not work in landscape mode. After this bug is fixed we could implement some of the useful suggestions you are making here.

And I don't think that landscape mode is useless for filechooser. Some may prefer it because it has a bigger chance of showing the full filename (some people give very long filenames to their books :)

dracodoc:
Yes, there are two methods to solve these problems:
1. make filechooser work in landscape mode. And it need additional efforts to save screen orientation.
2. Just save screen orientation and always operation filechooser in landscape mode. So one step will solve two problems.

Will anybody really operate filechooser in landscape mode? How can he use the keyboard in landscape mode? Sony 505 have a landscape mode which will rotate all the operations, I found it difficult to operate, and 505 don't have too many keys. I think it'll be impossible to input something with keyboard in landscape mode.

As for long file name, pressing right can show all the details of that file, and the showed first part of long file name should be enough to recognize different files.

tigran123:
Yes, I personally prefer the info shown by Right key when I need to see the full filename.

[...]

PR #596, opened 2012-11-16
Implements Kindle screen rotate on emulator and forces portrait mode on doc close

kai771:
Although I said “don’t look at me” in #573, it turned out to be pretty simple. This is not a full implementation - both J and K will just flip back and forth to landscape/portrait. I don’t think full implementation is necessary for development (who’d want to read upside down?), and I’m lazy by nature, so if anyone wants to do all modes - be my guest.

Second commit contains changes that will force portrait mode on closing documents. I tested it on both emu and Kindle, and it appears to work fine (at least better than without the patch. Try rotating the screen in a version prior to this PR, and then press Home. You’ll see what I mean). But, given my history of missing things, please do test yourself. I uploaded latest build [here](https://github.com/kai771/kindlepdfviewer/downloads), so it should be easy for anyone to test and report bugs.

### DIFF
--- a/einkfb.c
+++ b/einkfb.c
@@ -23,6 +23,11 @@
 
 #include "einkfb.h"
 
+#ifdef EMULATE_READER
+	int emu_w = EMULATE_READER_W;
+	int emu_h = EMULATE_READER_H;
+#endif	
+
 static int openFrameBuffer(lua_State *L) {
 	const char *fb_device = luaL_checkstring(L, 1);
 	FBInfo *fb = (FBInfo*) lua_newuserdata(L, sizeof(FBInfo));
@@ -89,14 +94,14 @@ static int openFrameBuffer(lua_State *L) {
 	if(SDL_Init(SDL_INIT_VIDEO) < 0) {
 		return luaL_error(L, "cannot initialize SDL.");
 	}
-	if(!(fb->screen = SDL_SetVideoMode(EMULATE_READER_W, EMULATE_READER_H, 32, SDL_HWSURFACE))) {
+	if(!(fb->screen = SDL_SetVideoMode(emu_w, emu_h, 32, SDL_HWSURFACE))) {
 		return luaL_error(L, "can't get video surface %dx%d for 32bpp.",
-				EMULATE_READER_W, EMULATE_READER_H);
+				emu_w, emu_h);
 	}
-	fb->vinfo.xres = EMULATE_READER_W;
-	fb->vinfo.yres = EMULATE_READER_H;
-	fb->buf->pitch = (EMULATE_READER_W + 1) / 2;
-	fb->buf->data = calloc(fb->buf->pitch * EMULATE_READER_H, sizeof(char));
+	fb->vinfo.xres = emu_w;
+	fb->vinfo.yres = emu_h;
+	fb->buf->pitch = (emu_w + 1) / 2;
+	fb->buf->data = calloc(fb->buf->pitch * emu_h, sizeof(char));
 	if(fb->buf->data == NULL) {
 		return luaL_error(L, "cannot get framebuffer emu memory");
 	}
@@ -191,7 +196,6 @@ static int einkUpdate(lua_State *L) {
 /* NOTICE!!! You must close and reopen framebuffer after called this method.
  * Otherwise, screen resolution will not be updated! */
 static int einkSetOrientation(lua_State *L) {
-#ifndef EMULATE_READER
 	FBInfo *fb = (FBInfo*) luaL_checkudata(L, 1, "einkfb");
 	int mode = luaL_optint(L, 2, 0);
 
@@ -199,6 +203,7 @@ static int einkSetOrientation(lua_State *L) {
 		return luaL_error(L, "Wrong rotation mode %d given!", mode);
 	}
 
+#ifndef EMULATE_READER
 	/* ioctl has a different definition for rotation mode. */
 	if (mode == 1) 
 		mode = 2;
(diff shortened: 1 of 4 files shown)


dracodoc:
I'm going to test it to show my support... :P

dracodoc:
It seemed to me that keep pressing J will keep rotating just as before. Actually I'd prefer a user configurable landscape angle, then just using one key to switch between landscape and portrait. 

Right now I have to use different keys to switch between landscape and portrait: K for rotate, J for rotate back. If there is only one landscape angle, one key can switch it back and forth which is simpler and easy to remember.

The force portrait mode on closing is working fine. The next thing should be remember the rotating mode in history so next time it can be opened and rotated automatically.

kai771:
@dracodoc Yeah, the only change I did on kindle is forcing portrait on document close. Function of J and K is not changed (on the Kindle. I implemented screen rotation for the emulator (it wasn't supported in the emulator before) and I was lazy to implement it fully, so in the emulator it just switches landscape/portrait, instead of really rotating it, so you can have upside down text. You don't normaly rotate your monitor, so I think it's not needed for emulation :) ). It's true that you need to use 2 keys to switch back and forth portrait/landscape, but they're right next to each other. Anyone can do it even with their eyes closed. They're "rotate" buttons. You want "portrait/landscape toggle". If we implemented it like that, I'm sure someone would complain that she can't rotate it (or that she has to choose different landscape angle before she gets what she wants, because she often switches hands holding the Kindle, and she prefers it with buttons on the left when holding it in landscape with the left hand, and buttons on the right when holding it with the right hand... let your imagination run wild). Supporting both is just too much effort for a simple little thing, that isn't at all difficult as it is now. Just my noob opinion... feel free to differ.

I have a similar opinion of the saving orientation with the book. Is it really that hard to press just 1 key after opening the book (and I thought I was lazy :D)? Well, I suppose I could see some justification for this... much more than for landscape/portrait toggle, at least.

(Note: sorry if this came out harsh, this is just my personal opinion, and I assure you I had no intention of offending anyone) 
Oh, right, and thank you for testing :).

dracodoc:
No problem. Since the current implementation is already there, it's probably not worth it to change to my preference.

Though I'd say saving orientation is definitely have many advantages, zero side effect and it should be really easy to do. If one book was to be read in landscape mode, user probably will read it later still in landscape mode. If user want to change to portrait mode anytime, he can press a key and change it.

kai771:
I was more of an opinion that she would look at the TOC first, or bookmarks, or flip a few pages to see where she left off before she continues reading... all more easily done in portrait mode. What do you guys think? As for implementation, it should be easy, especially now that I implemented :setRotationMode() :).

dracodoc:
The last screen should at least have some content that user have read it before, so it should be enough to remind user. The bookmarks only provide some page number list. And he can always flip back in landscape mode to help remembering last read.

The key difference is, the landscape view is vastly different from the portrait view even for same page, same contents. So always keep in landscape view for one book is better to maintain the reading memory. 

At last, user can always press one key to return to portrait mode if he felt it is needed. So the question is, which case is more often, to read continuously in landscape mode or switch back to portrait, move around a bit then switch to landscape again?

kai771:
I don't know about others, but I use fit-to-content-width almost all the time (when I don't, I use 2 column mode). Even if I used landscape, it would be fit-to-content-width. The page I see is the same in both portrait and landscape, and it looks the same - the text is just smaller in portrait, but there's more of it. So I wouldn't say it's vastly different. I must say I'd prefer if the page should be in portrait when I open it, because I already hold the kindle the right way. If I wanted to read in landscape, I'd already have to move my arms, so pressing a key on the way is not too much effort. Let's assess the following situation: you have many books, and you are not sure which one is the one you'd like to open. You'll open one book, see that that's not the one you're looking for, close it, open another... this is easier if you don't have to rotate the kindle after every opening to see if it's the book you were looking for.

And final reason - remembering book orientation requires more code. Always opening in portrait doesn't xD.

Again, these are just my opinions. The fact that I don't agree with you doesn't mean that others won't.

dracodoc:
The reason I say landscape mode is vastly different is that the amount of text you can see in one screen is very different. So with portrait mode you will have a memory of the whole page arrangement, but with landscape mode you probably can only remember the specific content and cannot see much page arrangement in that view.

Of course user can rotate the screen by pressing one key, but that will show the first part of current page, so it could be not the last reading position and user need to flip down several times to find the exact location.

The difference of our opinions are not much, we can always press one key to switch in either case, so it's just which mode should be default. Another solution will be make this behavior configurable, but that will introduce more complexity.

Anyway, this case is not a major one. I can live with the current status.

tigran123:
Interesting, I will have a look at some point.

tigran123:
At the moment I am testing only on the emulator and I see only one bug: fiveway up/down don't work on the emulator in landscape mode. But in portrait mode they still work.

kai771:
I think they're replaced by left-right in landscape. Please test, as I can't do it now.

tigran123:
Yes, you are right. It's ok because on the Kindle the keyboard is rotated too :)
(and on the emulator we'll just have to remember about this)

tigran123:
Ok, in that case I can't break it on the emulator. I'll test it on the Kindle after breakfast...

Thank you very much --- very impressive work in such small number of lines of code, that's what I like!

kai771:
I like it that way too :). But it wasn't me - it WAS simple. It just looked complicated. :)

tigran123:
There is one inconsistency: if you switch Amazon framework to landscape mode then the filechooser will start in landscape mode, but after opening/closing a document it will be put into portrait mode. So, we have to choose between the following two options:
1. Force portrait mode in filechooser on startup as well (not just on closing the document).
2. Upon closing the document force not the portrait mode but whatever mode was used at startup. But this option implies the extra work of fixing filechooser to work in landscape mode upon closing the document (as we know it currently does not).

The minimalistic approach is option 1. and I vote for it. There is no real benefit in running filechooser in landscape mode as we all agreed, I think.

kai771:
I agree with you. I'll fix it when I get home (using option 1). Or, you can merge now if you like, and submit the fix yourself. :)

tigran123:
Just calling Screen:setRotationMode(0) before the call to FileChooser:choose() would not work, so I leave it up to you to fix it properly.

kai771:
Ok, I'll take a look when I get back home. Won't be before evening though.

tigran123:
Ok.

kai771:
There, that should do it... it even restores the mode that was active in the framework on close :) (not my doing, it just turned out that way :))

Issue #597, opened 2012-11-16
shift S mode grayed out effect was not cleared later


dracodoc:
I'm testing the new test build of @kai771 : kindlepdfviewer-v2012.10-169-g3f6bea1. I found the implementation on Shift-S with Shift-X is much better than previous one but have bugs.

https://picasaweb.google.com/dracodoc/PdfRendering?authkey=Gv1sRgCPy6qpm2vtG1igE#5811545847241937922

The first pic is the first screen of one page on shift-S mode, there are some area left in this page. So in next pic we saw the left area was shown with the upper area grayed out. This is much better than the previous implementation which will move to next screen with several lines and a big blank area.

However, if I continue to read next page, I found the next page was shown in grayed out effect too. I have to switch to fit page first then switch back to get the normal view -- see next pic.

So here the grayed out effect should be cleared for next page.

kai771:
Ah, I knew I missed something... I'll fix it. Sorry. Fixed in #599.

dracodoc:
Thanks, that is quick!

PR #599, opened 2012-11-16
fixes a bug that dracodoc reported in #597. I missed it :(

kai771:


tigran123:
Hmm, I missed it too...

Issue #546, opened 2012-11-05
Bottom of the text cut off by koptreader


tigran123:
Here is the test file that illustrates the bug:

https://github.com/downloads/tigran123/libdjvu/Ideal-pdf-reflow-test.pdf

chrox:
This bug is likely to be fixed when I implement scroll mode reading for koptreader.

chrox:
The fix is already in master, so this issue should be closed.

PR #601, opened 2012-11-17
bugfix: fixed missing words in last line

chrox:
This patch fixed the "text cut off at page bottom" bug reported by @tigran123 at #546.

### DIFF
--- a/koptreader.lua
+++ b/koptreader.lua
@@ -321,11 +321,11 @@ function KOPTReader:drawOrCache(no, preCache)
 				return nil
 			end
 		else
-			local secs, usecs = util.gettime()
+			--local secs, usecs = util.gettime()
 			page:reflow(kc, self.render_mode)
-			local nsecs, nusecs = util.gettime()
-			local dur = (nsecs - secs) * 1000000 + nusecs - usecs
-			Debug("Reflow duration:", dur)
+			--local nsecs, nusecs = util.gettime()
+			--local dur = (nsecs - secs) * 1000000 + nusecs - usecs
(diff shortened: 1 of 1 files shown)


chrox:
@tigran123  Thanks to Willus, the author of K2pdfopt. He told me that I need only one line to fix the bug. And it truns out that he is right. That line just works like a charm. To test this patch, you need to rebuild libk2pdfopt.

tigran123:
It doesn't actually compile at the moment:

```
In file included from k2pdfoptlib/k2pdfopt.h:44:0,
                 from koptreflow.h:26,
                 from koptreflow.c:30:
willuslib/willus.h:303:8: error: conflicting types for 'dprintf'
/home/usrlocal/arm-2012.03/bin/../arm-none-linux-gnueabi/libc/usr/include/stdio.h:421:12: note: previous declaration of 'dprintf' was here
make[1]: *** [koptreflow.o] Error 1
make[1]: Leaving directory `/home/tigran/Kindle/kindlepdfviewer/libk2pdfopt'
make: *** [libs/libk2pdfopt.so.1] Error 2
```

tigran123:
Hmmm, this suggests that you are using a different toolchain version. We are using 2012.03 from CS.

chrox:
This bug is already fixed in PR #603. I simply removed the dprintf definition in libk2pdfopt as it's never used.

chrox:
I use the same toolchain as you guys do. But probably I've commented out the dprintf in stdio before. I'm sorry for the inconvenience.

tigran123:
No problem, thank you for fixing it so quickly.

PR #607, opened 2012-11-18
bugfix: Screen Saver not working if Help page is displayed

kai771:


tigran123:
I am not sure if this is worth fixing... Help page can be considered a kind of "modal dialog", i.e. it demands that the user interacts with it before it lets him do anything else.

Are you sure it is safe to call those command handlers like that, directly? Have you tested on Kindle? I see nothing wrong but I haven't actually tested it.

Btw, speaking of screen savers --- on Kindle DXG I now start KPV via Shift-P-K (i.e. kill the framework) and I really like the feature (that may look to someone like a bug, but to me it is a feature) whereby no picture is shown and everything appears to "freeze" with the old state of the screen plus the message "Going into screensaver..." on top. This way, even if one forgot to press Alt-. to enable battery logging one can still see the old battery state and then the new one upon returning from "screensaver". Very handy, please don't fix that one.

On DXG there is no reason whatsoever (I agree with @eLINK2gl now) to use Amazon's framework, because the only use of the latter is for the dictionary lookup and mobi files, but it is silly to read mobi files and use dictionaries on DXG when one can do that more conveniently on a Kindle 3. I deregistered my DXG from my main account and registered it on my wife's so as to get 5GB extra space in Amazon's Personal Documents area for physics/mathematics books, leaving Kindle 3 for History, Philosophy, Religion, Fiction and Dictionaries.

kai771:
I tested on Kindle, and it seems to work. I can't guarantee it will never break, but as far as I can tell, it works, the same as when going to screensaver from filechooser or opened book, for example. 

Btw, imno - if it is a bug, it needs fixing.

PR #610, opened 2012-11-18
correction for #608. I added the line in the wrong place

kai771:
My apologies. I made a mistake in a previous PR, #608. The correct place of the added line is this one. Sorry again.


tigran123:
I was about to raise a bug about it --- I noticed that bbox value is not restored correctly on opening the document which had the bbox set.

Issue #611, opened 2012-11-18
KPV won't build on emulator because of libk2pdfopt


tigran123:
Even though I did a "make clean ; make cleanthirdparty" on the emulator I get this error:

```
gcc -g -shared -fPIC -Wl,-soname,libk2pdfopt.so.1    -o libk2pdfopt.so.1 koptreflow_dyn.o k2pdfoptlib/k2file_dyn.o k2pdfoptlib/userinput_dyn.o k2pdfoptlib/k2parsecmd_dyn.o k2pdfoptlib/k2proc_dyn.o k2pdfoptlib/breakinfo_dyn.o k2pdfoptlib/k2version_dyn.o k2pdfoptlib/k2menu_dyn.o k2pdfoptlib/k2sys_dyn.o k2pdfoptlib/k2publish_dyn.o k2pdfoptlib/k2mark_dyn.o k2pdfoptlib/bmpregion_dyn.o k2pdfoptlib/k2settings_dyn.o k2pdfoptlib/k2bmp_dyn.o k2pdfoptlib/bmpregions_dyn.o k2pdfoptlib/k2usage_dyn.o k2pdfoptlib/k2master_dyn.o k2pdfoptlib/wrapbmp_dyn.o k2pdfoptlib/k2mem_dyn.o k2pdfoptlib/pagelist_dyn.o k2pdfoptlib/devprofile_dyn.o k2pdfoptlib/k2ocr_dyn.o willuslib/mem_dyn.o willuslib/fontrender_dyn.o willuslib/wmupdf_dyn.o willuslib/ocrjocr_dyn.o willuslib/ocr_dyn.o willuslib/bmpmupdf_dyn.o willuslib/win_dyn.o willuslib/wzfile_dyn.o willuslib/wsys_dyn.o willuslib/filelist_dyn.o willuslib/gslpolyfit_dyn.o willuslib/ansi_dyn.o willuslib/bmp_dyn.o willuslib/pdfwrite_dyn.o willuslib/math_dyn.o willuslib/linux_dyn.o willuslib/fontdata_dyn.o willuslib/point2d_dyn.o willuslib/willusversion_dyn.o willuslib/token_dyn.o willuslib/strbuf_dyn.o willuslib/wfile_dyn.o willuslib/wgs_dyn.o willuslib/string_dyn.o willuslib/render_dyn.o willuslib/array_dyn.o willuslib/bmpdjvu_dyn.o -lm  
[...]

PR #614, opened 2012-11-19
bugfix: remove koptreflow.o in make clean

chrox:
The koptreflow.o was not cleaned when running `make clean` in libk2pdfopt. This patch fix it.


chrox:
This patch fixed the bug reported in #611.

tigran123:
I think there were two .o files, not just koptreflow.o that were not removed in the clean target.

chrox:
Yes. I add the line `rm -rf *.o` in `make clean`.

tigran123:
Yes, I tested building in emulator mode (after building for ARM) and everything is fine. Thank you.

PR #616, opened 2012-11-19
bugfix: FileChooser crashes when closing landscape document

kai771:
Yes, we already fixed that forcing portrait mode on KPV start, and document close. However, this is a fix that will prevent the crash if we don't force portrait. I'm not suggesting that we stop forcing portrait - I think we all agreed that FileChooser in landscape mode on K3 is not very usable. But, this might be handy on Touch or KPW (we'll see about that later), so it should be usable should we need it. To test this, you need to comment out `self:setRotationMode(0)` line in UniReader:inputLoop() in unireader.lua, thus removing forcing of portrait on document close.

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -215,7 +215,7 @@ function FileChooser:choose(ypos, height)
 		local cface = Font:getFace("cfont", 22)
 
 		if self.pagedirty then
-			fb.bb:paintRect(0, ypos, fb.bb:getWidth(), height, 0)
+			fb.bb:paintRect(0, ypos, fb.bb:getWidth(), fb.bb:getHeight(), 0)
 			local c
 			for c = 1, self.perpage do
 				local i = (self.page - 1) * self.perpage + c
@@ -254,7 +254,7 @@ function FileChooser:choose(ypos, height)
 		end
 
 		if self.pagedirty then
-			fb:refresh(0, 0, ypos, fb.bb:getWidth(), height)
+			fb:refresh(0, 0, ypos, fb.bb:getWidth(), fb.bb:getHeight())
 			self.pagedirty = false
 		end
 


tigran123:
Hmmm, is this a correct fix? If the `height` variable is no longer needed in FileChooser:choose() method then you should have got rid of it altogether (and modified reader.lua (are there any other callers?) to call it without passing G_height) rather than leaving the interface the same but internally ignoring the variable (except for calculating the number of items per page). I think this is the wrong way to fix it.

Also, I remind you (for the second or third time, I think) that the commit message should explain what the change does and NOT what the user-observable effects of the change (or the absence of change) might be. You should focus on explaining _how_ you fixed something and not on _what_ you fixed. For example, if you did the right thing and explained the commit the proper way then you would have written something like this:

"Ignore (well, almost) the variable `height` passed to the function and forget about fixing the buggy callers of the function that cause the crash by passing garbage/stale value of `height` to it. Who cares why the crash happens if we can just tweak the code of the function internally to avoid it..."

And while typing this description you would have probably stopped and asked yourself "Hang on, am I doing the right thing here? Maybe I should take care of the caller who is passing garbage instead of changing the function". This is why proper commit messages are very important. I have aborted several of my own commits while typing the commit message precisely for this reason --- while trying to explain what exactly I did, I realized that the change I am making is the wrong thing to do, even if it appears to fix the problem.

kai771:
Ok, I'll try to explain:
The only change really needed to avoid the crash is in the 2nd line. The first can be the same, and it won't crash. It's true that height should be eliminated from the call of the procedure, but I was lazy. No, there aren't any callers except reader.lua.

Crash happens because of this:
FileChooser gets the height variable through a call. It then calls the adequate reader. While in reader, the screen rotates, and height (of the screen, not the variable "height") changes. After we exit reader, we go back to the FileChooser. Variable height in FileChooser is the same as when FileChooser is called from reader - say it's 800. However, the screen rotated while in Reader, and now G_height is 600. Calling `fb.bb:PaintRect` with height 800 when the screen is now 600 causes the crash. Instead of `fb.bb:getHeight()` we could've also used G_height, because G_height is the global variable and it gets updated on screen change.

All this is obvious from the code. That's why I didn't bother explaining. I much prefer writing code than writing essays. The change I made works. Imno, something that does work is always better than something that does not work. Even if it's sloppy. If someone does it better, then by all means, change it. But until then, use the one that works in preference to the one that doesn't.

I'm not going to waste my time explaining every little change that I make. I don't have that much free time.You can test for yourself if the change I made works. You don't really need to know how I fixed something (not because it's a secret, but because the program is pretty complex - I'm sure you're not familiar with every little detail of KPV, despite working on it for so long. You don't need to be). If you really want to know, you can look at the code. It's all in the code.

Let me tell you once and for all - I really don't like that arogant attitude of yours. It made NuPogodi stop showing up here. You're not my boss, nor am I your subordinate. I am not a developer. If I had to qualify myself, I'd call it "independent contributor". I have a very selfish reason for all the changes I contributed - I needed them. Personally. I'd much rather someone else did it already, so I could devote my time elsewhere, but since no one did, I had to. Having my code in the master is better than to manually modify it every time a new version comes out. Even if it means dealing with someone like you.

So, I'll continue to "independently contribute" my code whenever I feel like it, in a way that's most convenient for me. I'll continue to explain "observable effects of the change" in the commit message because that's what should be tested. How I did it is in the code, and there's no need to translate code into English. Whether you'll accept it or not is your choice.

tigran123:
Thank you for the explanation. In that case using the local copies of G_width and G_height shouldn't actually work. Hmmm, I'll need to retest #618 then...

tigran123:
As for the other part of your diatribe (not dignifying personal insults with a response, but stopping only on that which is beneficial for all of us) --- knowing how you fixed something is very important for all developers because it makes transparent all the assumptions held in each particular fix. Do you not agree that it is important?

tigran123:
Listen, if you did what I advised you to do, i.e. explained in the commit exactly what you did in this so-called "fix" then you would have noticed that your "fix" does not actually work. 

I have now tested it in the emulator and it does NOT WORK (unless the current directory has few enough books to fit on a single page). I cannot even say that your "fix" fixed the symptom rather than a bug. No, it did not even do that --- it only prevented the crash and hid a possible memory corruption. Such a "fix" is definitely worse than nothing (this is in answer to your question "is A(working) better than B(not working)" or something like that).

@houqp merged your PR without testing, obviously. This is fine --- we don't have time to test every single PR, but we definitely expect the submitter to test at least on the emulator.

So, in the future, please kindly follow the advices of those more experienced than you are AND also please test the changes before submitting a PR. Thank you. If you also find the strength of character to apologize for the personal insults I would appreciate it and would esteem you to be a man of honour and just character.

To be fair I should also say that your PRs are normally very good quality and show that you have tested the change both on the emulator and Kindle. So, this one is just an exception and not a rule.

tigran123:
when I said "your fix hid possible memory corruption" I guessed that it must be the case (intuition aided by experience is sometimes more powerful than observation :), but now I proved that it actually _is_ the case, i.e. with your changes it is very easy to cause it to coredump with "Segmentation fault".

I'll add a proper fix for this to PR #618

kai771:
I always try to change as little as possible in my fixes. Without actually checking, I'd say that you're probably right about there being a problem with my fix and a greater amount of books. I did test it, both on Kindle and on Emulator - I happened to have just a few books in each at the time. If I fixed it the way I actually wanted to, I would've simply used G_height instead of height in all instances of the :choose method. But, given your abstinence to using global variables (I don't understand why, since we're only reading them) I tried to find a solution that would actually satisfy YOU. So I used the fb.bb:getHeight(), thinking that YOU would prefer it that way. It was my mistake for not being consistent with it - in my test it appeared to work after just this 2 changes, so I left it at it. But if I did it my way, I would've replaced height with G_height everywhere in :choose() method, and it would all work fine. (Again, without checking, but I think I'm right on this one)

If my fix at least prevented a crash, and yours didn't even do that - who is the one that didn't test it?

As for insulting you, I did no such thing. Calling you arogant is just stating a fact.

kai771:
ROFL, now I see that you did just that - using global variables!

tigran123:
I always advised using global variables where this is appropriate. If you look in the history of PRs you will see this (recently I advised @chrox about this in koptreader). In fact my words were "there is nothing wrong with using global variables properly". It was in the PR #588

kai771:
Ah, but you asked me to use local in one of my recent PRs.

tigran123:
The keyword is "properly" :)

If you have two or more accesses to a global variable which cannot change in some scope then it is better to save a copy of it in the local variable. This is just Lua strangeness. You wouldn't need to do that in C.

But in FileChooser:choose() those variables CAN change and so we can't just save their local copies.

See, I sense now, that when we understand each other, all the childish hostility disappears. I offer you my hand as a friend and forgive your insults even though you did not ask for forgiveness, because you are my brother.

kai771:
Even in Lua, in that particular case, (even if there were more than 2 accesses) it wouldn't induce any percivable performance loss. You're just too stubborn about optimizations. All this mess trying to please you... sigh.

tigran123:
You are absolutely correct about 2-3 accesses. Yes. But think about the bigger picture. Another developer comes and looks at the code and thinks "it is all right to do so" and copies it in another context where the accesses could be not 2-3 but 20,000 and then there certainly would be performance penalty.

As for my being very careful about optimization --- it is because I am "originally" a Unix and Linux kernel developer. In the kernel we can't afford to be relaxed about performance, i.e. we have to be vigilant about every such little possibility. Ok, some of this may not translate into the world of application programming like KPV, but this can be "forgiven", I hope, as a peculiarity of each individual developer.

tigran123:
There is no need to please me...

Ah, this is actually an instance of a more generic philosophical issue that comes up when I teach my students. I explain to them that they must NOT try to "please God" because they take their idea of God, solidify it in the present shape and then try to "please" that solid image in the future. This is SLAVERY and is very tiring. But if, instead, they tried to look into themselves and find the best ideals corresponding to the idea of God at each and every moment then they would suddenly find that they are FREE, i.e. they found true FREEDOM by pursuing the TRUTH, BEAUTY and GOODNESS at each particular moment in time.

Likewise, you, instead of pleasing me, try to form your own best conception of what is RIGHT (and it is all right that this RIGHT will change in time, as you learn) and do that instead. You will find (as long as you are sincere) that this is a much more satisfying and cooperative way of doing things. Every thing, even programming.

kai771:
When they use them 20,000 times, that's the time when you should correct them, not every thime someone uses it more than 2 times :). And I'm pretty sure kernel is not written in Lua, so you can relax a bit. Isn't it also the unix way to do optimizations only when they're necessary - the rationale being that code readability being more important, and programmer's time being to valuable to be vasted on unneeded optimizations? (I read it in some book, but can't remember which one now. It was by some Unix guy :). I'll look it up if you're interested :)

kai771:
Well, if I don't do it the way you like it, you'll ask me to change it until you do, so... I tried to save some time :). "Pleasing" was in that sense - to avoid having to redo it.

tigran123:
Yes, I absolutely agree with you that code readability is the most important thing. Amen to that! That is why I was so peeky about some of your less detailed commit messages. I.e. if something is not readable then I hope that at least the commit message explains it.

tigran123:
If I ask you to do something that you feel is NOT RIGHT then you must INSIST on your opinion and explain the reasons why you hold it. Surely there are witnesses (and you are one of them if you read all PRs) who can testify that I have changed my opinion upon being convinced of the rightness of the opposite one? I am neither obstinate nor arrogant. But I love all and I wish to give all that I am to all and that makes me vulnerable to all kinds of attacks and accusations (which I readily forgive, for I know the true essence of the relationship between all living beings, including humans --- and that essence is BROTHERHOOD :)

kai771:
It's not that it's not right, it's that it doesn't matter either way. If you know what I mean.

PR #618, opened 2012-11-19
Cleanup the forcing of portrait mode and make FileChooser work in landscape mode

tigran123:
1. Move the setting of G_width/G_height inside the function Screen:setRotationMode()
2. Don't call self:clearCache() twice on closing the document (once is enough).
3. Get rid of UniReader:setRotationMode() method because Screen:setRotationMode() is enough, as there is nothing reader- or document-specific in this function.
4. Enhance FileChooser:choose() to work in both portrait and landscape modes. Namely, I removed the assumption that G_width and G_height retain the values they had upon the entry to this function.

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -202,7 +202,11 @@ function FileChooser:setPath(newPath)
 	end
 end
 
-function FileChooser:choose(ypos, height)
+function FileChooser:choose()
+	local ypos = 0
+	local width, height = G_width, G_height
+	local wlen = width - 2*self.margin_H
+
 	self.perpage = math.floor(height / self.spacing) - 2
 	self.pagedirty = true
 	self.markerdirty = false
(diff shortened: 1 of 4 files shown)


tigran123:
Please don't merge this PR yet as I will try to cleanup the fix done in PR #616 here as well.

tigran123:
Ok, all done and tested on Kindle as well as on the emulator.

tigran123:
Wait, there is still a problem here, namely we are using stale values of G_width/G_height in FileChooser:choose(), i.e. relying on forcing portrait mode on closing the document. I'll add a commit that fixes this _properly_.

tigran123:
Just wanted to confirm that I tested it on both the emulator and Kindle.
(also tested that it doesn't conflict with any other PRs I submitted today).

PR #626, opened 2012-11-21
bugfix: calculate current cache size and free oldest cache if needed

chrox:
Cache ttl should not be decreased only when there is no free slot for
new cache but decreased when claiming new cache each time. Otherwise
caches will be freed altogether which obviously is not the purpose of
the cache system. This patch only free the oldest caches and make just
enough free slots for the new one.

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1062,21 +1062,31 @@ function UniReader:cacheClaim(size)
 		error("too much memory claimed")
 		return false
 	end
-	while self.cache_current_memsize + size > self.cache_max_memsize do
-		-- repeat this until we have enough free memory
+	-- reduce ttl of each cache
+	local calculated_size = 0
+	for k, _ in pairs(self.cache) do
+		self.cache[k].ttl = self.cache[k].ttl - 1
+		calculated_size = calculated_size + self.cache[k].size
+	end
+	-- free oldest cache until we have free memory
+	while calculated_size + size > self.cache_max_memsize do
+		local oldest_cache_ttl = self.cache_max_ttl
+		local oldest_cache_key = nil
 		for k, _ in pairs(self.cache) do
-			if self.cache[k].ttl > 0 then
-				-- reduce ttl
-				self.cache[k].ttl = self.cache[k].ttl - 1
-			else
-				-- cache slot is at end of life, so kick it out
-				self.cache_current_memsize = self.cache_current_memsize - self.cache[k].size
-				self.cache[k].bb:free()
-				self.cache[k] = nil
+			if self.cache[k].ttl <= oldest_cache_ttl then
+				oldest_cache_ttl = self.cache[k].ttl
+				oldest_cache_key = k
 			end
 		end
+		-- cache slot is at end of life, so kick it out
+		if oldest_cache_key then
+			Debug("free cache:", oldest_cache_key, "ttl:", self.cache[oldest_cache_key].ttl)
+			calculated_size = calculated_size - self.cache[oldest_cache_key].size
+			self.cache[oldest_cache_key].bb:free()
+			self.cache[oldest_cache_key] = nil
+		end
 	end
-	self.cache_current_memsize = self.cache_current_memsize + size
+	self.cache_current_memsize = calculated_size + size
 	return true
 end
 


tigran123:
Hmmm, maybe @houqp or @hwhw (or whoever wrote that stuff) should comment before we merge it, just in case if you missed some reason for the current situation (but I don't see any, so that would mean I am missing it too :)

houqp:
Make sense to me :) Not sure how @hwhw thinks about it, since he wrote the cache system.

unireader.lua
```diff
@@ -1062,21 +1062,31 @@ function UniReader:cacheClaim(size)
...
-				self.cache[k].bb:free()
-				self.cache[k] = nil
+			if self.cache[k].ttl <= oldest_cache_ttl then
+				oldest_cache_ttl = self.cache[k].ttl
+				oldest_cache_key = k
 			end
 		end
+		-- cache slot is at end of life, so kick it out
+		if oldest_cache_key then
+			Debug("free cache:", oldest_cache_key, "ttl:", self.cache[oldest_cache_key].ttl)
```
chrox, line 1083: You can simply add a tracer of cache free in the original code here. `Debug("free cache:", k, "ttl:", self.cache[k].ttl)` And you will repeat the bug.

```
# Reader:setZoom globalzoom_mode: -1  globalzoom: 1.2176560121766  globalrotate: 0  offset: 66.2100456621 0  pagesize: 467.57989501953 800  min_offset: 0 0
# free cache: 411_1.2176560121766_0_1 ttl: 0
# free cache: 412_1.2176560121766_0_1 ttl: 0
# free cache: 413_1.2176560121766_0_1 ttl: 0
# free cache: 424_1.2176560121766_0_1 ttl: 0
# free cache: 425_1.2176560121766_0_1 ttl: 0
# free cache: 426_1.2176560121766_0_1 ttl: 0
# free cache: 427_1.2176560121766_0_1 ttl: 0
# free cache: 436_1.2176560121766_0_1 ttl: 0
# free cache: 437_1.2176560121766_0_1 ttl: 0
# free cache: 414_1.2176560121766_0_1 ttl: 0
# free cache: 415_1.2176560121766_0_1 ttl: 0
# free cache: 416_1.2176560121766_0_1 ttl: 0
# free cache: 417_1.2176560121766_0_1 ttl: 0
# free cache: 428_1.2176560121766_0_1 ttl: 0
# free cache: 435_1.2176560121766_0_1 ttl: 0
# free cache: 434_1.2176560121766_0_1 ttl: 0
# free cache: 430_1.2176560121766_0_1 ttl: 0
# free cache: 431_1.2176560121766_0_1 ttl: 0
# free cache: 429_0.91324200913242_0_1 ttl: 0
# free cache: 418_1.2176560121766_0_1 ttl: 0
# free cache: 419_1.2176560121766_0_1 ttl: 0
# free cache: 420_1.2176560121766_0_1 ttl: 0
# free cache: 421_1.2176560121766_0_1 ttl: 0
# free cache: 422_1.2176560121766_0_1 ttl: 0
# free cache: 423_1.2176560121766_0_1 ttl: 0
# free cache: 432_1.2176560121766_0_1 ttl: 0
# free cache: 433_1.2176560121766_0_1 ttl: 0
# free cache: 410_1.2176560121766_0_1 ttl: 0
# rendering page 438
# page_indicator 438
# pre-cached page  438  and restored offsets
# globalzoom_mode: -1  globalzoom: 1.2176560121766  globalrotate: 0  offset: 66.2100456621 0  pagesize: 467.57989501953 800  min_offset: 0 0
```

When rendering page 438 there is no free cache slot left. So the program just free all previously cached page from page 410 to page 437. Actually what we want is just making enough free slot for page 438 and just kicking out page 410 or page 411. The test is running with PDFReader. Similar result is available with koptreader. And the self.cache_current_memsize is not accurate either which will make it hard to free any slot in some case in koptreader.

With this patch you will get the result like this:

```
# Reader:setZoom globalzoom_mode: -1  globalzoom: 1.2176560121766  globalrotate: 0  offset: 66.2100456621 0  pagesize: 467.57989501953 800  min_offset: 0 0
# free cache: 410_1.2176560121766_0_1 ttl: -8
# rendering page 438
# page_indicator 438
# pre-cached page  438  and restored offsets
# globalzoom_mode: -1  globalzoom: 1.2176560121766  globalrotate: 0  offset: 66.2100456621 0  pagesize: 467.57989501953 800  min_offset: 0 0
```

When rendering page 438 only page 410 is freed from the cache. I think it will be more efficient to work in this way. Note that both tests are performed in the same way of paging from 410 to 438 with PDFReader.

tigran123:
Hmmm, this patch made cacheClaim() function 25-30 times slower than the old one. But I haven't measured yet whether the overall page draw performance is affected. I hope it is not, i.e. the contribution of cacheClaim() should be less than 1% of the overall time needed to render a page. I will check now.

Update: my guess was correct --- the contribution of cacheClaim() is negligible, so it doesn't matter that it is now 30 times slower. Specifically, it was 1 microsecond and now became 30 microseconds but the total time of rendering is 8000 microseconds, so extra 30 make no difference. These times are measured on the emulator but the same proportion (not absolute values of course) should be observable on Kindle also.

chrox:
The analysis is encouraging. Thanks. And I should not calculate the current memsize each time the cacheClaim is called. I think it will save a fraction of time.

chrox:
So on an average case, when cacheClaim is called the cache table is traversed twice, one for decreasing ttl of each cache and the other is finding the cache with the least ttl and free it. And both of the traversing should be unnecessary which I think contributes most to the 30 microseconds.

PR #627, opened 2012-11-21
bugfix: write cache before new reflows && fix memory leak

chrox:
Reflowed bitmap should be written to cache before a new reflow is
started because all reflowing processes share the same masterinfo in
libk2pdfopt which contains the data of reflowed bitmap. If a new reflow
is started before a priviously reflowed bitmap is written, the
priviously reflowed bitmap will be corrupted which may cause
segmentation fault when it is written into cache finally.

### DIFF
--- a/koptreader.lua
+++ b/koptreader.lua
@@ -345,7 +345,7 @@ function KOPTReader:logReflowDuration(pageno, dur)
 	local file = io.open("reflowlog.txt", "a+")
 	if file then
 		if file:seek("end") == 0 then -- write the header only once
-			file:write(string.format("FILE\tPAGE\tDUR\n"))
+			file:write("FILE\tPAGE\tDUR\n")
 		end
 		file:write(string.format("%s\t%s\t%s\n", self.filename, pageno, dur))
 		file:close()
@@ -359,14 +359,15 @@ function KOPTReader:logMemoryUsage(pageno)
 	if status_file then
 		for line in status_file:lines() do
 			local s, n
-			s, n = line:gsub("VmData:%s-(%d+) kB", "%1")	
+			s, n = line:gsub("VmData:%s-(%d+) kB", "%1")
 			if n ~= 0 then data = tonumber(s) end
 			if data ~= -1 then break end
 		end
+		status_file:close()
 	end
 	if log_file then
 		if log_file:seek("end") == 0 then -- write the header only once
-			log_file:write(string.format("PAGE\tMEM\n"))
+			log_file:write("PAGE\tMEM\n")
 		end
 		log_file:write(string.format("%s\t%s\n", pageno, data))
 		log_file:close()
--- a/koptreader.lua
+++ b/koptreader.lua
@@ -296,7 +296,7 @@ function KOPTReader:drawOrCache(no, preCache)
 				Debug("precache pagehash", self.precache_pagehash)
 				if self.precache_pagehash == pagehash then
 					Debug("write cache ", self.precache_pagehash)
-					return self:writeToCache(self.precache_kc, page, self.precache_pagehash, preCache)
+					return self:writeToCache(self.precache_kc, page, self.precache_pagehash, true)
 				else
 					self.precache_kc = nil
 					self.precache_pagehash = nil
@@ -312,22 +312,31 @@ function KOPTReader:drawOrCache(no, preCache)
 			Debug("threaded preCache is returned.")
 		end
 	else
-		if use_threads and self.precache_kc ~= nil and self.precache_kc:isPreCache() == 1 then
-			if self.cache[self.cached_pagehash] then
+		if use_threads and self.precache_kc ~= nil then
+			if self.precache_kc:isPreCache() == 1 and self.cache[self.cached_pagehash] then
 				InfoMessage:inform("Rendering in background...", DINFO_DELAY, 1, MSG_WARN)
 				return self.cached_pagehash, self.cached_offset_x, self.cached_offset_y
+			elseif self.precache_kc:isPreCache() == 0 then -- cache is ready to be written
+				Debug("write cache", self.precache_pagehash)
+				self:writeToCache(self.precache_kc, page, self.precache_pagehash, true)
+				Debug("reflow page", pagehash)
+				local ok, page = pcall(self.doc.openPage, self.doc, no) -- reopen current page
+				kc = self:getContext(page, no, true)
+				page:reflow(kc, self.render_mode)
+				return self:writeToCache(kc, page, pagehash, false)
 			else 
 				Debug("ERROR something wrong happens .. why cached page is missing?")
 				return nil
 			end
 		else
 			--local secs, usecs = util.gettime()
+			Debug("reflow page", pagehash)
 			page:reflow(kc, self.render_mode)
 			--local nsecs, nusecs = util.gettime()
 			--local dur = (nsecs - secs) * 1000000 + nusecs - usecs
 			--Debug("Reflow duration:", dur)
 			--self:logReflowDuration(no, dur)
-			return self:writeToCache(kc, page, pagehash, preCache)
+			return self:writeToCache(kc, page, pagehash, false)
 		end
 	end
 end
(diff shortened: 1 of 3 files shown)


tigran123:
So, you managed to put to rest that memory leak bug --- congratulations, well done @chrox ! :)

chrox:
Yes. Below is the stats of memory usage when turning over 200 pages. You can see that memory usage fluctuates at around 36MB. That's pretty good result. Cheers!

Memory Usage Stats

koptreader.lua
```diff
@@ -343,26 +352,47 @@ function KOPTReader:logReflowDuration(pageno, dur)
...
+	local status_file = io.open("/proc/self/status", "r")
+	local log_file = io.open("reflow_mem_log.txt", "a+")
+	local data = -1
+	if status_file then
+		for line in status_file:lines() do
+			local s, n
+			s, n = line:gsub("VmData:%s-(%d+) kB", "%1")	
+			if n ~= 0 then data = tonumber(s) end
+			if data ~= -1 then break end
+		end
```
tigran123, line 365: There should be a call to `status_file:close()` here --- otherwise you will run out of file descriptors (plus memory leak due to Lua's overhead per each descriptor).

koptreader.lua
```diff
@@ -343,26 +352,47 @@ function KOPTReader:logReflowDuration(pageno, dur)
...
+	end
+	if log_file then
+		if log_file:seek("end") == 0 then -- write the header only once
+			log_file:write(string.format("PAGE\tMEM\n"))
```
tigran123: Also, `string.format()` is not needed for the header, but only for the data. I'll get rid of it in battery.lua as well.

chrox:
Yes. I will fix it right now. The data of above figure is from non-multi-threads mode. I found that in multi-thread mode the data field in "/proc/self/status" is not accurate at all. It keeps going up at a surprising speed which cannot be the truth. When I use top command to get the resident memory size of the process, the two modes look the same.

chrox:
So I would like to use VmRSS field in "/proc/self/status" to indicate the resident memory size.

tigran123:
The VmRSS is good for some purposes, but most memory leaks will escape its notice because "lost memory" is not usually referenced. So, VmData would catch memory leaks but VmRSS is good for analyzing active memory usage patterns of a process.

Btw, if VmData keeps growing then we may have a memory leak.

chrox:
No. Because VmData is not accurate in multi-threads mode. I said it keeps going up at a surprising speed as after tens of page turnings the VmData just exceeds 256MB in my Kindle 3. But KPV is still running. It's impossible to be right. How can a process use memory more than the physical memory considering there is no swap memory in Kindle 3. You can test it after turning on multi-threads in koptreader and watch the memory stats.

tigran123:
I was about to say that before jumping to conclusion about memory leak one should check the semantics of VmData in the case of a process with multiple threads. Of course it is not "not accurate". It is just different and it depends on what flags were passed to the clone(2) system call that was made when the thread was created by pthread_create(3) library function.

Btw, what is so special about 256MB VmData value on Kindle? The address space of a 32bit process can grow to much higher value (arch-dependent and also dependent on kernel configuration) regardless of the actual amount of physical memory available. Oh, I have just re-read your message and noticed the question "How can a process use memory more than the physical memory" --- this is normal under Unix. Process can allocate (but it would never be able to fault it in) a lot more memory than the amount of physical memory available. If we were to restrict applications to just the physical memory then things like Mozilla Firefox or Skype or any large applications would probably never even start up :) It is like a bank --- it claims to have a lot more money than it actually does and as long as all customers don't try to withdraw it all simultaneously all is fine.

On the emulator the kpdfview process is single-threaded regardless of the setting in Aa dialog (multi threads on/off). Do you know why? Update: I see the other thread now. Yes, VmData is growing infinitely. This implies a memory leak somewhere. VmData does not lie. (well, this _assumes_ that one has checked the VmData semantics in multi-threaded case and confirmed that it has the same meaning with the particular type of pthread_create() call that we are using. I haven't done this and it must be done before assuming that we have a memory leak.)

tigran123:
Btw, djvulibre is multi-threaded too. But if you watch kpdfview process while browsing a djvu file the value of VmData doesn't grow as much as it does with koptreader, even though the number of threads can go up to 3 (if you press pgdown key fast enough). So it definitely looks like a memory leak in koptreader and NOT a feature of VmData with multi-threaded processes.

chrox:
Yes. I found where is wrong now after a few searches in Google. As you can see in the figure attached below, the VmData increase at an average speed of 8MB/page. And on each page turning one thread is spawned which is allocated 8MB of stack size. Obviously the stack allocated for each child thread is not reclaimed or reclaimed but not updated in VmData.

tigran123:
Plausible analysis but it does not explain why we don't see this 8MB per thread create/exit growth when we page through a djvu file.

Actually, if you "watch -n 0" (as opposed to -n 1) then you will see LOTS of thread create/exit events when viewing a djvu file. So we should, according to your theory, see lots of 8MB increments in VmData, which is not the case. Ergo, your theory is not correct.

chrox:
Yes. Because libdjvu uses `pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED)` when child thread exit. And this probably solve our problem too. I will try it right now.

tigran123:
I noticed that you pass NULL as attributes. Yes, we thought about the same thing simultaneously :)

I was reading pthread_create(3) manual page and noticed that they use pthread_attr_init() to initialize attributes and you are passing NULL. Maybe we should set up attributes too?

tigran123:
Ok, try your idea first (I see that it is a slightly different idea) and see if it solves our issue.

chrox:
Ok. It works now. See the figures attached below.

tigran123:
Great. Don't forget the two minor cleanups I mentioned as well. So, in multi-threaded mode it needs a lot more memory. But this is not physical memory (that would be VmRSS), so it's ok. As you found out (by using top) the physical memory requirements for both modes are about the same, which is the important thing.

chrox:
And I found when opening a djvu document in DJVUReader the VmData is also around 100MB which is similar to multi-thread mode in koptreader.

tigran123:
You forgot to call `pthread_attr_destroy(&attr)` immediately after `pthread_create(...)`. Destroying the `attr` object has no effect on the threads created using that object, so this is a safe thing to do. Otherwise we are leaking any resources that may be allocated by `pthread_attr_init()` function.

chrox:
Yes. I'm sorry I forgot that. I'll fix it right now.

chrox:
Now I'm considering making multi-threads mode the default in koptreaders since I cannot find any obvious bug in it now.

tigran123:
I am assuming that the functions `pthread_attr_init(), pthread_attr_setdetachstate(), pthread_attr_destroy()` are extremely fast, i.e. setting up a few structures in memory (possibly allocating them dynamically). But if this is the wrong assumption then you should create and initialize the `attr` on open() and destroy it in close() of the document, to not slow down the reflow operation.

chrox:
Actually in `djvulibre/libdjvu/GThreads.cpp` threads are created like this.

```
int  
GThread::create(void (*entry)(void*), void *arg)
{
  if (xentry || xarg)
    return -1; 
  xentry = entry;
  xarg = arg;
#ifdef DCETHREADS
  int ret = pthread_create(&hthr, pthread_attr_default, GThread::start, (void*)this);
  if (ret >= 0)
    pthread_detach(hthr);
#else
  pthread_attr_t attr;
  pthread_attr_init(&attr);
  pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
  int ret = pthread_create(&hthr, &attr, start, (void*)this);
  pthread_attr_destroy(&attr);
#endif
  return ret;
}
```

So we should not worry about the performance of these function calls.

tigran123:
Yes, I saw that code and also checked `pthread_attr_init()` implementation (though not a Linux one, but it should be similar) and saw that we should be ok :)

PR #629, opened 2012-11-22
make multi-threads mode default in koptreader

chrox:
Since there is no reported bug in multi-threads mode and multi-threads do improve paging performance a lot.


tigran123:
The only objection to this would be if the initial memory requirements for multi-threaded mode (i.e. actual faulted in VmRSS) are substantially higher than for single-threaded mode. This would cause KPV to fail to open some file and the user wouldn't know which bit of source to edit to switch to single-threaded. But you said you checked it, so I believe you that it is safe.

Now, it would be great if you could do it via a symbolic define in defaults.lua. The koptreader.lua part of your commit is trivial to change for this, but the koptconfig.lua part is not so obvious, i.e. would need some code to initialize `default_item` and `current_item` values. But don't you think it would be nice to have a single configurable value listed in defaults.lua? If you don't mind making this change, please do...

chrox:
Since a PR with a better solution will be send, this PR is closed in case that it would be merged by mistake.

PR #630, opened 2012-11-22
add koptconfig defaults in defaults.lua and some cleanups

chrox:


tigran123:
Oh, this is much nicer implementation, thank you. I'll test it and merge it if all is ok.

Update: seems to work fine.

PR #634, opened 2012-11-24
Touch gesture detection module

houqp:
I am lazy, so I just copied the design from Firefox OS:

https://github.com/mozilla-b2g/gaia/blob/master/shared/js/gesture_detector.js

Basically it's just state machine. State is changed according to the touch input. 

I added a scheduler in inputevent.lua module so sorry that we have two scheduler now (another in ui.lua). Because I need to simulate the setTimeOut function in Javascript during the gesture detection, so this is the only way I can think of so far. Maybe we can clean it up later. Also to help with the detection and scheduling, I also added a time.lua module which abstract the operations on time values.

Currently, only single tap, double tap and hold gesture is supported. Since single tap is enough to handle most of the operations we need during reading, I decide to stop developing for new gestures but switch back to more important part of the reader.

If you want to test it but you don't have kindle Touch, you can try it in emulator. Mouse click is translated into touch events.

Of course this is just one possible solution for touch support, feel free to come up with any better implementation :)

PR #636, opened 2012-11-25
bugfix: waitForInput should return error when select returns zero

chrox:
On the RETURN VALUE section of `man select` it says:

```
       On success, select() and pselect() return the number of file descriptors contained
       in the three returned descriptor sets (that is, the total number of bits that  are
       set  in  readfds,  writefds,  exceptfds)  which may be zero if the timeout expires
       before anything interesting happens.  On error, -1 is returned, and errno  is  set
       appropriately; the sets and timeout become undefined, so do not rely on their con‐
       tents after an error.
```

So if select call returns zero error should be returned in waitForInput.


chrox:
It seems that this PR contains PR #634 as I have merged the 'ges-detect' branch from @houqp's repo.

houqp:
I would suggest apply this patch on top of new_ui_code branch, so I can merge it without merging the gesture detection code, since we haven't decided the implementation yet ;p

chrox:
Since PR #634 wants to merge into new_ui_code I thought it was decided. Anyway I will pickup the kopt-specific commits and the waitforinput bugfix and open a new PR tomorrow. And this PR is closed in case that it's merged by mistake.

PR #637, opened 2012-11-26
bugfix: waitForInput should return error when select returns zero

chrox:
And cleanups for the new submodule libk2pdfopt.


houqp:
@chrox , for the pdfopt part, is it a new commit or a cherry pick? If I cherry pick previous commit from master branch, will I have a conflict? I still hope that all the backend module are kept in sync between master branch and new_ui_code branch :)

I plan to cherry pick all the latest commits from master to new_ui_code in following two days to keep new_ui_code up to date.

chrox:
This is a new commit. I thought it would be hard to cherry pick from the master to include the latest kopt-specific commits as I always changed several files in one commit. But I will try it later.

chrox:
I would like to cherry pick commits from master so this new commit will be discarded.

chrox:
@houqp Yes. Cherry-pick is a good tool. I like it.

PR #638, opened 2012-11-26
zoom mode support in reader menu & menu module bug fix

houqp:
Zoom mode is still incomplete because bbox support for mupdf and libdjvu hasn't implemented yet. I will try to fix that in this week...


chrox:
All zoom modes work on Kindle Paperwhite except the fit content modes as bbox has not been implemented yet.

houqp:
Ok, merging now since @chrox has tested it ;p

sorry for keep merging my own RP...

PR #642, opened 2012-11-27
add a demo version of koptreader in new UI KPV

chrox:
Only page down and page up are working in this demo. More will be added soon.

### DIFF
--- a/frontend/ui/reader/readerview.lua
+++ b/frontend/ui/reader/readerview.lua
@@ -44,6 +44,7 @@ function ReaderView:paintTo(bb, x, y)
 			self.state.zoom,
 			self.state.rotation,
 			self.render_mode)
+		self:recalculate()
 	else
 		self.ui.document:drawCurrentView(
 			bb,
--- a/frontend/document/document.lua
+++ b/frontend/document/document.lua
@@ -220,4 +220,5 @@ end
 
 require "document/pdfdocument"
 require "document/djvudocument"
+require "document/koptdocument"
 require "document/credocument"
--- a/frontend/document/koptdocument.lua
+++ b/frontend/document/koptdocument.lua
@@ -1,0 +1,185 @@
+require "cache"
+require "ui/geometry"
+require "ui/screen"
+require "ui/device"
+
+-- Any document processed by K2pdfopt is called a koptdocument
+KoptDocument = Document:new{
+	_document = false,
+	-- muPDF manages its own additional cache
+	mupdf_cache_size = 5 * 1024 * 1024,
+	djvulibre_cache_size = nil,
+	dc_null = DrawContext.new(),
+	screen_size = Screen:getSize(),
+	screen_dpi = Device:getModel() == "KindlePaperWhite" and 212 or 167,
+	configurable = {
+		font_size = 1.2,
+		page_margin = 0.06,
+		line_spacing = 1.2,
+		word_spacing = 0.15,
+		quality = 1.0,
+		text_wrap = 1,
+		defect_size = 1.0,
+		trim_page = 1,
+		detect_indent = 1,
+		multi_threads = 0,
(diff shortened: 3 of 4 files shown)


chrox:
What's interesting is that even though the Kpw has a faster CPU than K3 the reflowing process is even slower in Kpw. I will look into this more carefully and find out why this could happen.

Update: No. With more controlled tests the result now favors Kpw which speeds up reflowing by roughly 20%. It's much more reasonable now.

houqp:
wow, that's awesome :)

too sad that I will have exams today and tomorrow :( 
so I won't be able to test thoroughly until tomorrow night.

houqp:
I tried open `test/tall.pdf` and was trapped in the do loop in line 579 in `pdf.c`. Thus it keeps printing `reading page:0,0,0,0 zoom:0.00 dpi:0` in terminal. Can you reproduce it in your environment?

Open `test/2col.pdf` has the same effect.

chrox:
No. I have no such problem. I updated libk2pdfopt in this Patch. So have you run `git submodule update` and `make cleanthirdparty thirdparty`?

houqp:
Yes, that's strange...

```
 $ git submodule status                                                                                   
 01a26bed05c8f5ed36e015d16cc54e52dddcccb7 djvulibre (debian/3.5.24-9-24-g01a26be)
 a9516b455e811b9f57f2805d6c08b6744f797083 kpvcrlib/crengine (cr3.0.57-15)
 2e18531e4d48f46eba8c12749723ba23417b8920 libk2pdfopt (remotes/origin/HEAD)
 149e0fb8ec6295325e1e448bbe79b7e9bf2285aa luafilesystem (v1.5.0-15-g149e0fb)
 eb6f890ebd01ee4cf4f31a66d8a946bc8dccd885 luajit-2.0 (v2.0.0-rc1)
 05219d086a5136ee19f643cf062bd5c0d3aef5d3 mupdf (1.1-233-g05219d0)
```

houqp:
OK, it's working now after `make clean`, though I have no idea why `make cleanthirdparty` is not enough...

chrox:
That makes sense. Because koptcontext.o is dependent on libk2pdfopt/koptreflow.h and in recent libk2pdfopt I added a new field called dev_dpi in KOPTContext stucture.

houqp:
Ok, cool, merging now.

frontend/ui/reader/readerview.lua
```diff
@@ -44,6 +44,7 @@ function ReaderView:paintTo(bb, x, y)
 			self.state.zoom,
 			self.state.rotation,
 			self.render_mode)
+		self:recalculate()
```
chrox, line 47: @houqp I would like to see your comment on this. Because this recalculation is a hack for koptreader which cannot determine page size before the page is rendered. Should we have a more elegant way to update the page size?

houqp, line 47: Yes, we need a more elegant way. I will look into it this weekend. Feel free to change the UI framework if you come up with any good idea.

houqp, line 47: how about call `page:reflow(kc, render_mode)` in `getPageDimensions(pageno, zoom, rotation)` and put `kc` in cache, so `KoptDocument:renderPage` can get it from cache?

chrox, line 47: Well. It is better than the recalculation hack. And I will implement it in this way.

houqp, line 47: I removed the recalculate call in my latest commit: https://github.com/houqp/kindlepdfviewer/commit/ae1c489a0fb94317bef1fa5a60b4d320af4f8ea6

So we might have a conflict for your next RP, but we can handle it manually later.

BTW, if you have better idea for the implementation, just ignore my suggestion ;P

PR #643, opened 2012-11-28
Bugfix: crash when switching user privilege level

tigran123:
A simple sequence of steps can cause segmentation violation (coredump), namely:
1. Start KPV in a directory with just two page worths of books visible
   in "beginner" mode but containining more than two page worths of files
   visible in "expert" mode.
2. Change user privilege level from beginner to expert
3. Page down three times (to arrive on page 3)
4. Now change user privilege level back to beginner
5. Observe that the cursor is sitting on an empty line (non-existent book)
6. Try paging down -> segmentation violation. If you page up instead of down then everything will be OK and you won't be able to page down anymore.

### DIFF
--- a/filechooser.lua
+++ b/filechooser.lua
@@ -180,6 +180,8 @@ function FileChooser:setPath(newPath)
 
 		-- Dijkstra will hate me for this...
 		if newPath == oldPath then
+			self.page = 1
+			self.current = 1
 			return
 		end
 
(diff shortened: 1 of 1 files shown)


tigran123:
Oh, I guess I broke my own rule --- instead of explaining how I fixed it I explained how to reproduce the problem :)

Here is the explanation of the fix: when changing the privilege level and calling FileChooser:setPath() function we must ensure that the current page and position within the page are valid even if the underlying content (i.e. the list of files and directories) has changed. So, if no better values are obtainable then the position/page should be reset to the very top of the first page.

tigran123:
Hmmm, don't merge it just yet as this fix has the unpleasant side-effect of `Space` key resetting the position to the top. I'll think about how to cope with this...

tigran123:
Closing this PR as a better fix will come in a separate PR (no need to pile up commits unnecessarily).

PR #650, opened 2012-12-04
add touch support for readerrolling

houqp:


chrox:
OK. Rolling works on Epub files now.

PR #655, opened 2012-12-07
add close button for menu widget

houqp:
So far it is just a "x" letter, we might want to change it to a nice pic later ;)


chrox:
Tested. It works.

PR #657, opened 2012-12-08
bug fixes and new feature for menu widget

houqp:
now a menu can be closed by tapping on any area that is outside of the menu. while the close button on right upper corner is used as "go back to last menu".

### DIFF
--- a/frontend/ui/filechooser.lua
+++ b/frontend/ui/filechooser.lua
@@ -1,6 +1,8 @@
 require "ui/menu"
 
 FileChooser = Menu:new{
+	height = Screen:getHeight(),
+	width = Screen:getWidth(),
 	path = ".",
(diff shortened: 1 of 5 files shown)


chrox:
When opening KPV in emulator with command `reader.lua -d test`, clicking the blank area of the file chooser window will raise a lua config error saying:

```
lua config error: ./frontend/ui/ui.lua:153: no dialog left to show, would loop endlessly
stack traceback:
    [C]: in function 'error'
    ./frontend/ui/ui.lua:153: in function 'run'
    ./reader.lua:126: in main chunk
```

And KPV exits unexpectedly.

houqp:
fix attached :)

chrox:
It works on paperwhite now. :)

PR #658, opened 2012-12-08
add change view mode feature for crereader

houqp:


defaults.lua
```diff
@@ -68,6 +68,9 @@ DPICVIEWER_COMICS_MODE_ENABLE = true
 DPICVIEWER_RTL_MODE_ENABLE = false
 DPICVIEWER_PAGE_MODE_ENABLE = false
 
+-- 0 for scroll mode, 1 for page mode
```
tigran123: This is a very dangerous definition, because the actual values rely on the type `enum LVDocViewMode` defined in `kpvcrlib/crengine/crengine/include/lvdocview.h`. I am not saying we shouldn't do this --- I did exactly the same thing for djvu render mode in djvureader.lua, but I described this danger in the comment, so perhaps we should add a comment explaining why these specific numeric values of 0 and 1 must be chosen (and cannot be changed), referring to the lvdocview.h header file.

tigran123:
Also, shouldn't the toggle actually switching between these two modes be implemented in the same PR? If you have page mode only, then (if it behaves like CoolReader) the pictures won't be rendered properly sometimes (i.e. cut between pages if they don't fit). So, it would be nice to be able to switch between them and not lose scroll mode completely.

Update: I see, you are suggesting that this be a user-editable defaults.lua type of value, rather than a programmatic toggle... Hmmm, ok. But it would be nice to have it controlled from within KPV so one doesn't have to exit and start a shell and run `vi defaults.lua` and then re-start KPV...

houqp:
OH, yes, I will redo the RP. Just being a little bit lazy recently ;p

tigran123:
Thank you :)

I see that this variable is passed to the open() method, so it requires re-opening the file. I suggest (the laziest way) to set it withing crereader.lua (and save in the document-specific or even global (?) settings) while the document is open, but to make it into effect on the next open (and try to condense it into one-line help description of the command like "toggle scroll/page (requires re-open)" or something short enough to fit). This is because putting it into filechooser is inappropriate (as it is reader-specific) and having to close and re-open the document on the fly is too complicated and I bet 95% chance that it will create a memory leak  :)

houqp:
Any suggestion for the "scroll/page viewmode" toggle shortcut? I am so bad at choosing this kind of stuff ;p

tigran123:
The `V` command seems to be unused in crereader.

houqp:
OK, view mode toggle added :)

tigran123:
Ok, I'll test it now.

Wonderful, using the page mode fixes the most serious problem with KPV's "builtin crengine" (as opposed to CoolReader), namely it shows the footnotes at the bottom of the page instead of as links. However, the footnote label is shown centered on a line by itself instead of at the beginning of footnote paragraph. Do you know if this can also be fixed? Then our builtin crengine will be almost as good as standalone CoolReader (only hyphenation will be missing still).

Now, in your PR there is a bug: the setting is not restored properly when the file is opened again. It seems to be saved in the history/file.lua but not loaded, i.e. reset to "scroll" mode on the next open.

Also, I didn't like the two consequtive :inform() (DELAY + NODELAY) calls because the second one stays on the screen indefinitely. How about shortening them into a single one like this:

```
-      InfoMessage:inform("Changed to "..self.view_mode.." mode.", DINFO_DELAY, 1, MSG_AUX)
-      InfoMessage:inform("Please reopen the document.", DINFO_NODELAY, 1, MSG_AUX)
+     InfoMessage:inform("Viewmode: "..self.view_mode.." (needs re-open)", DINFO_DELAY, 1, MSG_AUX)
```

tigran123:
Also, I see that in page mode the "standard CoolReader header" is displayed. This is very good, but can we configure crengine to display the battery (in percent, not those useless bars) and the author/title as well, just like in the standalone CoolReader?

Also, I would definitely suggest to make "page" the default reading mode, to make it look like CoolReader.

houqp:
Yes, I will take some time to dig through crengine's code and hopefully I can fix another cache bug

>  the setting is not restored properly when the file is opened again.

That's very strange, actually it works in my emulator...

tigran123:
Did you simply re-open the document from "Last Documents" or did you exit reader.lua command altogether? In the first instance it does work, but if you exit all the way and restart KPV then it doesn't work.

But what is really weird is that now I have re-tested (after deleting history/\* files) and found that the footnotes are rendered correctly (i.e. with the number in the same paragraph rather than on a separate line, centered). But the above bug is still there. I.e. if you exit KPV and re-open the document then the viewmode is always scroll. But if you close the document and re-open from Last Documents (or filechooser) then the viewmode is correct.

houqp:
Oh, yes. I did not test it that way. Good catch :) Will fix it tonight.

houqp:
I will probably look into the title bar tomorrow

tigran123:
But you left "scroll" as the default. Don't you think that "page" would be a better default? I.e. to have the same look and feel as standalone CoolReader.

tigran123:
Btw, I figured out why I had those "centered line" footnote labels --- it was specific to a book. I.e. in that particular fb2 file the footnotes had `<subtitle>N</subtitle>` before the `<p>` with the footnote text.

tigran123:
I confirm that the current state works fine. If you don't like the idea of having page mode as default please let me know and I'll merge the PR now.

houqp:
ha, my bad, forgot to change default mode

tigran123:
For the header stuff we need to do something like this in cre.cpp:

```
doc->text_view->setPageHeaderInfo(PGHDR_AUTHOR | PGHDR_TITLE | PGHDR_PAGE_NUMBER | PGHDR_PAGE_COUNT | PGHDR_CHAPTER_MARKS | PGHDR_BATTERY | PGHDR_CLOCK);
```

But I haven't tested it on real Kindle yet (and on emulator there is no battery :)

tigran123:
Ok, let's merge this one and deal with the header in a separate PR.

houqp:
Good find ! But the battery is not working in kindle, which is expected. We have to teach crengine how to read kindle's battery ;p

kai771:
Links seem to be broken when using page view mode. I tried it with this file (emu only, didn't try it on kindle):
http://www.epubbooks.com/books/download/hdhv/gulliver-s-travels.epub

When pressing L, link shortcut keys will be displayed way off. Strangely, if you press Back, and then L again, their position will change and will be much better - only off for a height of crereader header. On some pages, pressing L will report "No page links on this page", even though they're clearly visible between square brackets. Changing font size up and back allows the links to be recognised.

I only saw new footnotes on fb2 files. Is it supported for epub too? (Just checking if my epub files are bad)

tigran123:
@houqp Ok, then we should at the very least OR `PGHDR_CHAPTER_MARKS` into the flags to get the chapter marks.

It is very strange that the Author and Title fields don't work either. Maybe we need to do some initialization magic to tell crengine to parse metadata and discover Author/Title values.

houqp:
@kai771 , oops, I completely forgot the link stuff... we have to recalculate link position for page mode then.

@tigran123 , yes, `PGHDR_CHAPTER_MARKS` is very useful :) Indeed, title and author do work for me. What book did you use for testing?

tigran123:
@houqp really? I tried several books (Herodotus, Plutarch, Hitler) and none showed the Author/Title. But the standalone CoolReader does show it correctly.

Ah, wait.... maybe it is the font issue --- i.e. the font which crengine uses for the header has no Russian glyphs maybe? Obviously all the books I tried were in Russian (as crengine is only stable with fb2 files and most fb2 files are in Russian :)

Update: yes, it is the font issue. I tried the English fb2 file and Author/Title are displayed correctly. We can use the crengine API to set the font face to be used for the header to something more sensible. Or, even better, allow it to be set via KPV's generic `F` font setting interface. That would be a very nice PR for someone to implement it --- not complicated at all, but about half a day's work.

Issue #77, opened 2012-03-28
need help from integrating crengine library


houqp:
I am not familiar with cmake and get stuck in the build system :( 

Since we are now using freetype, I think it's better to keep using it as the font engine in CREngine. By default, CREngine enables Fontconfig. To disable it, we need to set the `USE_FONTCONFIG` macro to 0. But CREngine uses a global `crsetup.h` header to define macros and since this file is part of upstream, I cannot change its content. 

CREngine uses `cmake` build system, and it seems that arguments from command line like `CFLAGS` and `CC` are not recognized by the auto generated makefile. 

One possible solution might be adding an outer CMakeLists.txt in kindlepdfviewer's root directory and invoke CREngine's cmake system with custom changes of `CFLAGS` `CC` etc. But introducing cmake to kindlepdfviewer is obviously over kill.

Or we can add command in current makefile to temporally modify `crsetup.h` before building. However, this seems too dirty and it still cannot specify variables like `CC`.

I have run out of ideas. Any guidance?

PS: in case you want to look at the source, a half-done demo for crengine can be found at [my branch](https://github.com/houqp/kindlepdfviewer/tree/epub).

dpavlin:
I spend way too much time trying to implement GUI=KINDLEPDFVIEWER target and failing to compile crengine without gui propery. So I decided to hard-code @houqp hack into make fetchthirdparty to provide some help in bootstrap.

My eventual goal is patch to crengine's CMakeLists.txt or new DEVICE_NAME definition which can be submitted upstream or patched on the fly like we had for mupdf.

Android crengine target, on the other hand, includes full Makefile which me might adopt for our needs if we don't want to mess with pushing changes upstream.

houqp:
I am for patching to upstream.

houqp:
We also need to update the jpeg library used by CREngine to release 8d, to keep up with MuPDF.

houqp:
OK, I just managed to force CREngine use the same jpeglib as mupdf does with dirty hacks. 

Firstly, use the headers from mupdf's thirdparty by modifying the `CMakeList.txt` in `crengine`:

``` diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 94abccb..de41aa1 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,8 +176,9 @@ FIND_PACKAGE(JPEG)
 endif (NOT WIN32 AND NOT CR3_JPEG)
[...]

PR #659, opened 2012-12-12
multiple fixes & cre view mode support & update from master

houqp:
also added registration in ReaderMenu so all other reader widgets can add items in the main menu.


houqp:
I updated crengine's version and the build environment is broken now, so don't test it for now. I am fixing the build error.

houqp:
OK, it working now. I finally find out the fix for the sigfault error I mentioned in #77. Please help test it :)

houqp:
merging now since it has been out there for too long ;p

PR #665, opened 2012-12-14
bugfix: when link underlines are off, bottom line should be drawn when pressing L

kai771:


tigran123:
Can you elaborate please? What exactly did you do to reproduce this bug? When I press L with link underlines off nothing spectacular happens --- the shortcuts are drawn correctly and then pressing Back makes them disappear.

houqp:
Hi @kai771 , I also cannot reproduce this in my books, could you please give us the file you tested on?

kai771:
Any pdf file with links will do. It's not a spectacular bug in the first place. When you press L, it's supposed to draw a box around the link. In the initial version, before we added turning link underlining on/off, it did just that. It didn't draw the bottom line, because links were already underlined - so the underline was the bottom line of the box. With link underlines turned off, the bottom line of the box is not drawn. This fix draws that line. Purely cosmetic bugfix. Nothing major.

houqp:
Confirmed, thanks for the fix :) I will leave the merge to @tigran123

tigran123:
Ok, I understand now. Thank you for the fix. Sorry about the delay, I am having email problems because Gmail insists on having SSL certificates only signed by "recognized authority" and rejects self-signed SSL certificates as of 3 days ago (so I didn't realize that almost 3 days' worth of email is missing :)

PR #666, opened 2012-12-15
Enable chapter marks in crengine header.

tigran123:
Since crengine can't get to the battery level I didn't enable the battery flag.

As for showing the Author/Title correctly (for Russian fb2 books), crengine tries the following fonts in this order:

```
DEFAULT_STATUS_FONT_NAME = "Arial Narrow, Arial, DejaVu Sans"
```

So, the solution is very simple: just copy the appropriate files (arialn*) to kindlepdfviewer/fonts directory and then the Author/Title in the header are displayed correctly.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -66,6 +66,7 @@ static int openDocument(lua_State *L) {
 	doc->text_view->setViewMode(view_mode, -1);
 	doc->text_view->Resize(width, height);
 	doc->text_view->LoadDocument(file_name);
+	doc->text_view->setPageHeaderInfo(PGHDR_AUTHOR|PGHDR_TITLE|PGHDR_PAGE_NUMBER|PGHDR_PAGE_COUNT|PGHDR_CHAPTER_MARKS|PGHDR_CLOCK);
 	doc->dom_doc = doc->text_view->getDocument();
 	doc->text_view->Render();
 


kai771:
Or you can add this line:

```
doc->text_view->setStatusFontFace(lString8("Droid Sans"));
```

and make it use Droid Sans. It displays Russian Author/Title in emu correctly.

tigran123:
Good find. I didn't know about setStatusFontFace() function. I looked at setInfoFont() and this required a strange object of type `font_ref_t` so I quickly gave up.

However, your solution has the disadvantage of hardcoded "Droid Sans" font whereas Arial Narrow font is ideal for this sort of thing, because the space in the header is very limited and book titles are usually too long to fit. But it saves 544K disk space and, besides, we can't re-distribute Arial Narrow anyway.

Anyway, I have no objection to your solution, but I think that mine is equally acceptable.

But if you could enhance the font selection dialog to add a new field "crefont" (or whatever you choose to call it) so that the face could be selected at runtime then I would certainly prefer your solution as more flexible and superior.

kai771:
Nah, I won't bother. Just wanted to share the knowledge. You are free to use it any way you wish.

tigran123:
Ok, no problem. I might implement it myself. If you also find how to enable crengine hyphenation, please feel free to share that knowledge also :)

PR #669, opened 2012-12-18
bugfix: Back when link shortcut keys are shown moves to the top of the page

kai771:
When you are in fit-to-content-fidth mode, and the screen is not at the top of the page, and you press L to see link shortcut keys, and then press back, you’ll be moved to the top of the page. This fixes it.


tigran123:
Thank you. The fix tested, works fine.

PR #670, opened 2012-12-18
bugfix: when in fit-to-content-width, and you start panning, overlap should not be shown

kai771:


tigran123:
Yes, this makes sense, thank you.

PR #672, opened 2012-12-19
Enable hyphenation and set status font

tigran123:
Two changes:
1. Enable hyphenation engine. So, now our built-in crengine (with proper fonts and configuration settings) renders books as good as a standalone CoolReader (but some books take a bit longer to load).
2. Set the font used for status line to "Droid Sans" to render Russian Author/Title correctly. Although I personally prefer using "Arial Narrow" for this purpose, we cannot re-distribute it and therefore setting it to "Droid Sans" is a good compromise. Thanks to @kai771 for showing how to set the status font.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -39,6 +39,7 @@ static int initCache(lua_State *L) {
 	int cache_size = luaL_optint(L, 1, (2 << 20) * 64); // 64Mb on disk cache for DOM
 
 	ldomDocCache::init(lString16("./cr3cache"), cache_size);
+	HyphMan::initDictionaries(lString16("data/hyph/"));
 
 	return 0;
 }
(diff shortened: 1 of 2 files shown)


houqp:
Nice! Don't know it can be done in one line of code :)

houqp:
I think a better way to handle status font is use what ever font user chose for the document content.

tigran123:
@houqp For the document content the best font to use is serif like Adobe Garamond Premier Pro, but for the status font the best one is sans-serif Arial Narrow. Using Garamond for status would be quite unreadable (too small) and likewise reading with Arial Narrow makes it look like a computer terminal and not a real paper book (albeit electronic paper :)

houqp:
I see, agree with you ;)

houqp:
Oh, since we are already using "Droid Sans", then why not switch to "Droid Sans Fallback"?

tigran123:
Or, even better, add an option to our font-selection menu instead of hardcoding anything there. This way one can install "Arial Narrow" and use them instead of "Droid Sans" or anything.

houqp:
Agree.

kai771:
I already did that in Librerator, if you want reference:
https://github.com/kai771/kindlepdfviewer/commit/8a1ea4994c03c57969ac2dd67f816223d87c6551

tigran123:
@kai771 Thank you, I will have a look.

PR #674, opened 2012-12-20
bug fix in readerfont & minor code adjustment & page status in readerrolling

houqp:
# reader font fix

After set font in ReaderFont:onReadSettings, we have to send UpdatePos
event in order to set m_is_rendered(member of LVDocView) to true.

Otherwise position inside document will be reset to 0 on first view
render, this makes the first page of document is always displayed on
document open.

So far, I don't know why this call will alter the value of m_is_rendered.
It just works...
# page status for readerrolling

now when rendering in page view mode, we only keep track of page
number instead of pos inside document. Because using pos inside
document for page view mode is really a bad idea and will lead to
many unsovlabe bugs...

XPointer is also added to credocument so we can do accurate jumping inside document.

view_mode status in removed from credocument because it should be kept track by UI widgets instead.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -294,6 +294,33 @@ static int getFontFaces(lua_State *L) {
 	return 1;
 }
 
+static int setViewMode(lua_State *L) {
+	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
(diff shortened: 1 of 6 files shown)


houqp:
merged since it's out there for too long and later development is dependent on the bug fix

PR #675, opened 2012-12-20
Bug fix for screen rotate in crereader & fix from @kai777

houqp:
this fixes bug reported by @kai771, i.e. CW rotation
is not handled in crereader.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -68,7 +68,6 @@ static int openDocument(lua_State *L) {
 	doc->text_view->Resize(width, height);
 	doc->text_view->LoadDocument(file_name);
 	doc->text_view->setPageHeaderInfo(PGHDR_AUTHOR|PGHDR_TITLE|PGHDR_PAGE_NUMBER|PGHDR_PAGE_COUNT|PGHDR_CHAPTER_MARKS|PGHDR_CLOCK);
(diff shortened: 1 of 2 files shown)


kai771:
@houqp btw, you might find my changes to cre.cpp interesting too. They enable you to change view mode without the need to reopen document :).

houqp:
Thanks @kai771 , that's awesome!

tigran123:
Thank you, I will test all this tomorrow.

tigran123:
I wonder if ->setViewMode() is the function we should be using for toggling the view mode, as opposed to ->toggleViewMode(). What I am worried about is the need to update TOC when switching between page/scroll modes --- this sort of bug would be hard to notice by testing (as normally the "pages" of all sections will remain the same when re-typesetting the text in page and scroll modes), so I would have to check the implementations of ->setViewMode() and ->toggleViewMode() to make sure they are equivalent for this purpose.

Update: toggleViewMode() just calls setViewMode() appropriately, so we are okay.

As I promised, I will test everything today.

tigran123:
Everything looks good and user-selection of crereader header font should be in a separate PR.

houqp:
@tigran123 , yeah, i forgot to change create a new branch when adding @kai771 's patch ;p my bad

PR #680, opened 2012-12-24
demo koptreader for new UI

chrox:
Actually this is the first usable koptreader for new UI KPV. All configurations are working except manual bbox settings. As the new ui code is much cleaner than the old one, it's possible to integrate koptreader into normal pdfreader and djvureader. So let's reconsider the reflow mode.


houqp:
Can we switch back and forth between normal mode and reflow mode now?

chrox:
@houqp Not yet. I would like to send another PR with the reflow mode. So this PR is closed.

PR #681, opened 2012-12-30
add more items to gitignore file

houqp:


chrox:
It seems that the `.kpdfview.lua` files created along with the document files are not used at all. All document settings are stored in the `history` directory.

houqp:
the `.kpdfview.lua` file is used when no configuration file is found in history directory. Mainly for backward compatibility.

PR #682, opened 2012-12-30
Swipe gesture support

houqp:
also this gesture is used in menu widget for page turn


chrox:
It works fine on my Paperwhite except the following debug info after each swipe gesture:

```
# {
    ["handler"] = "onInputError",
    ["args"] = {
        [1] = "Waiting for input failed: 22\
"
    }
}
```

houqp:
Ok, I will test it that on my KP the day after tomorrow

houqp:
OK, fix attached, please let me know if the error still exists

chrox:
Good. The problem is solved.

PR #684, opened 2013-01-03
add reflow mode in pdf/djvu readers

chrox:
Tap on the bottom of the screen will popup the reader configuration dialog. Reflow mode can be toggled on and off in the "Layout" options.


houqp:
I like current implementation a lot :)

Just one more comment I think it is easy for people to mistakenly tapped on the bottom and invoke the config menu while they are expecting a page turn. Because it is easy to distinguish upper and lower half, right and left half but not 1/8 of the screen.

How about divide the upper portion of the screen into left and right. Then use the left part for menu and right part for config interface.

chrox:
The input layout looks like this which leaves one half of the screen area for next page turn, twice larger than the previous page turn. I think it will be OK after users are familiar with this layout. 

input_layout

houqp:
Ok, then let's leave it as is now. We will make it customizable by users at the end anyway :P

PR #687, opened 2013-01-06
bug fix in readerrolling

houqp:
- use ReaderRolling.view_mode instead of self.ui.document.view_mode


chrox:
Tested. Rolling works now with this patch.

PR #689, opened 2013-01-06
add last_percent field in pdf/djvu document history

chrox:
And normalize last_percent for cre document

### DIFF
--- a/frontend/ui/reader/readerrolling.lua
+++ b/frontend/ui/reader/readerrolling.lua
@@ -103,8 +103,9 @@ end
 
 function ReaderRolling:onCloseDocument()
 	local cur_xp = self.ui.document:getXPointer()
-	self.ui.doc_settings:saveSetting("last_percent", 
-		self.ui.document:getPosFromXPointer(cur_xp) / self.doc_height)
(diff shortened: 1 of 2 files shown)


frontend/ui/reader/readerrolling.lua
```diff
@@ -214,7 +214,7 @@ function ReaderRolling:gotoXPointer(xpointer)
 end
 
 function ReaderRolling:gotoPercent(new_percent)
-	self:gotoPos(new_percent * self.doc_height / 10000)
+	self:gotoPos(new_percent * self.doc_height)
```
houqp: Is it really necessary to get rid of the 10000 factor in order to sync with the native profess bar? There are two drawbacks to do so:
1. say we are on page 6 in a book of 400 pages, the last percent becomes 0.015 then. When we pass it to cre:gotoPos, luaL_checkint will turn it into 0, so we end up in the first page. 
2. it breaks backward compatibility of configuration files.

If we really need to get rid of that factor, we could create a new entry only used for sync and leave last_percent as is.

chrox: That makes sense. I will create a new entry called `percent_finished` for KPVBooklet.

PR #692, opened 2013-01-08
Modify those functions about the view_mode equal "page" when the kindle in Horizontal screen state.

ZenRoi:
Modify those functions about the view_mode equal "page" when the kindle in Horizontal screen state.

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -129,7 +129,7 @@ function CREReader:preLoadSettings(filename)
 		end
 	else
 		self.view_mode = DCREREADER_VIEW_MODE
-	end	
+	end
 end
 
 function CREReader:loadSpecialSettings()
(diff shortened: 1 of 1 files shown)


tigran123:
With this change the TOC doesn't work, i.e. following a TOC entry jumps to a completely wrong location. Also, changing the font and the header font causes the location to always jump to the end of the book (maybe this is a "random location", but it just happened to be the end of the book in the couple of tests I have done). I tested with a random/typical fb2 book, nothing special about it.

houqp:
I tested it only on EPUBs and cannot reproduce the bugs you mentioned. Could you provide one of the fb2 book for testing purpose? I don't have any fb2 books here ;p

tigran123:
I forgot to mention that to reproduce the bug you have to change either the base font or the header font. And if you close the document and re-open it again then it will work UNTIL you change the base or header font again.

tigran123:
Yes, sure, here is a test fb2 file:

http://www.bibles.org.uk/test.fb2.zip

houqp:
Thanks for the test file, bug confirmed :)

And this is caused by out of date `self.toc` after document is reformatted. That means every time we re-typeset the document with new font, new line space or what ever, we need to set `sel.toc` to nil so new toc can be retrieved on request.

I will leave the fix to @ZenRoi , since it's a good chance for him to practice on git.

tigran123:
Ah, I see, yes, that makes sense. Thank you for the explanation :)

PR #694, opened 2013-01-08
add render mode to hint page

houqp:
this fix crash on djvu file


houqp:
I will fix the conflict later

houqp:
OK, it's ready for test now

chrox:
Tested. It works on pdf and djvu files.

PR #695, opened 2013-01-09
return to caller when there is no window to show in UIManager

houqp:
So forked child processes can be reaped by parent.


chrox:
Tested. Both reader.lua processes are killed when KPV exits. So I can remove the killing code in KPVBooklet. Thanks.

houqp:
Awesome, sleeping time now, will try to fix the screen saver bug tomorrow :)

PR #696, opened 2013-01-09
set refresh type on eink update in fillMxcfbUpdateData

houqp:


einkfb.c
```diff
@@ -97,16 +97,18 @@ inline void fillUpdateAreaT(update_area_t *myarea, FBInfo *fb, lua_State *L) {
 }
 
 inline void fillMxcfbUpdateData(mxcfb_update_data *myarea, FBInfo *fb, lua_State *L) {
+	myarea->update_mode = luaL_optint(L, 2, 0);
 	myarea->update_region.top = luaL_optint(L, 3, 0);
 	myarea->update_region.left = luaL_optint(L, 4, 0);
 	myarea->update_region.width = luaL_optint(L, 5, fb->vinfo.xres);
 	myarea->update_region.height = luaL_optint(L, 6, fb->vinfo.yres);
 	myarea->waveform_mode = 257;
-	myarea->update_mode = 0;
```
chrox, line 105: If I guessed right it will do full refresh when `update_mode` equals 0. But it didn't before this fix.

chrox:
Anyway I will test this patch right now.

houqp:
Oh, wait! seems that it's doing full refresh when update_mode is set to 1. I will test it now...

houqp:
OK, now it should work...

chrox:
Thanks. It works now.

chrox:
Still don't know how does it work on Kindle Touch.

houqp:
Yeah, hopefully it should work :)

PR #698, opened 2013-01-10
retrive new toc when document is re-formatted

houqp:


chrox:
The "Finished Percentage" increases each time when a EPUB document is opened and closed immediately without page rolling. This can be confirmed by running `watch 'tail -n 20 /tmp/kpvbooklet.log'`.

houqp:
OK, seems that we cannot use pos to save read progress for page mode, this bug also exist in master branch.

chrox:
It's much better now. Thanks.

PR #699, opened 2013-01-10
Bring back screen saver support

houqp:
Seems that 5.3 firmware is different from v3 firmware on lipc-wait-event. In v5 firmware, when cvm is stopped, lipc-wait-event won't be able to catch any events. So we have to leave it running...

Currently, switching between screen savers works fine as long as we don't rotate the screen. I will rewrite screen rotation code to make cvm happy later.


chrox:
It works like a charm. Thanks.

PR #704, opened 2013-01-11
two bug fixes

houqp:
- use last_xpointer to keep reading progress for crereader

adapted from commit https://github.com/houqp/kindlepdfviewer/commit/e7e6a2be9c3258c452a8329db3c2d56638e99480

This fix a bug in crereader, i.e. when in page mode, keep reopening a document gives you different pages each time.
- reset self.toc on document reformat as discussed in #692

### DIFF
--- a/crereader.lua
+++ b/crereader.lua
@@ -160,12 +160,14 @@ end
 
 function CREReader:getLastPageOrPos()
 	local last_percent = self.settings:readSetting("last_percent")
+	local last_xp = self.settings:readSetting("last_xpointer")
+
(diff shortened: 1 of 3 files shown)


tigran123:
Ok, but shouldn't we also update fileinfo.lua to show the correct last reading position when the user presses Right key? The code looks for `last_percent` but now the history file contains `last_xpointer`.

houqp:
Oh, yes, you are right, I totally missed that... Should do a grep before committing. Will fix that after I cooked my dinner :)

PR #706, opened 2013-01-12
cre: load document afer set font

houqp:
This PR depends on #704 (I did it by mistake...), so you might want to test that one first.


tigran123:
I will test this and #704 today.

PR #709, opened 2013-01-13
couple of enhancements to crengine

houqp:
This RP also brings back the cache fix from master


chrox:
Tested on emulator since my paperwhite was bricked again and I don't think it can be recovered without touching the serial console. So sorry I cannot test this patch on paperwhite. You can merge this PR if you have already tested on real Kindle.

houqp:
Sorry to hear that :( Good luck to your PW...

PR #713, opened 2013-01-13
two bug fixes

houqp:
- restore line space
- handle page mode in redrawCurrentPage

bugs fixed by this commit can be reproduced by simply invoke help page and then return to document.


tigran123:
Aha, now changing gamma won't jump to the end of the book, thank you!

PR #714, opened 2013-01-14
some config UI changes

chrox:
Summary:
1. make the dimension of config dialog change according to the size of current config options
2. invert icon color temporarily when tapping on menubar


houqp:
After selecting one of the option in config UI, both the render result and selected item is not updated? I have to close the UI to force an update.

chrox:
How about displaying some options as icons so that tapping on these options could be easier on Kindle?

houqp:
Yes, icons are definitely better for touch screen :)

PR #716, opened 2013-01-14
bug fix for crereader's toc

### LINKED ISSUE #456: epub table of contents not recognized
Epub don't have a page number so it will be difficult to locate certain location without table of content support. 

Some epub's table of content is not recognized by kpv but can be used in pc software or calibre viewer, is it some compatibility problem?

https://www.box.com/s/7ibfr8uyo7l7ts7vf13r

houqp:
This fixes #456

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -193,7 +193,7 @@ static int walkTableOfContent(lua_State *L, LVTocItem *toc, int *count) {
 	int i = 0,
 		nr_child = toc->getChildCount();
 
-	for(i = 0; i < nr_child; i++)  {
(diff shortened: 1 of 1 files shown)


tigran123:
Oh, I'll need to test it on fb2 files first as the issue #456 is related to epub files.

tigran123:
Perfect, now Herodotus's History is showing Book I in TOC as well (not just Book II and later). I wonder why this bug only showed in some books but not all (maybe because most books have non-interesting first TOC entry like "title page" or some such nonsense which is why it was never missed).

PR #719, opened 2013-01-17
use page info to calculate percent finished in crereader

chrox:
It seems that the pos info is much less accurate than the page info for percent calculation.


houqp:
cre module has a method called `getCurrentPercent`, I will try that later.

Issue #711, opened 2013-01-13
add css selection menu for crereader


houqp:

kai771:
If anyone is interested, this feature is present in Librerator.

tigran123:
If this feature is really worth having then someone would bother porting it to KPV. I, for one, don't think it is worth spending time on it as CoolReader3 for reading fb2 files is still superior to KPV, even with the latest addition of hyphenation and page mode to the latter... The real strength and value of KPV is in the PDF and DjVu area, not in crereader.

But there is no harm in porting it, if someone has the time of course :)

houqp:
closed since feature implemented in master

Issue #710, opened 2013-01-13
add hyphenation toggle for crereader


houqp:

dracodoc:
I think maybe just modify the css to disable hyphenation will be a more general method. I modified my epub.css and saw the difference. So this request is not needed anymore if the css selection is implemented.

houqp:
Yes, BTW, how did you managed to disable hyphenation in CSS?

dracodoc:
The original epub.css

---

```
body { 
    text-indent: 0px; 
    margin: 0; 
    text-align: justify; 
}
```

---

My modified epub.css

---

```
body { 
    text-indent: 40px; 
    margin: 0; 
    text-align: justify; 
    hyphenate: none; 
}
```

---

tigran123:
Implementing a simple toggle for hyphenation is easier than requiring the users to edit their css files or having to supply a bunch of different css files to suit every possible user's wish.

dracodoc:
I don't understand why you would call adding a toggle in KPV easier than adding one line in css. 

The css already defined hyphenation, use an application toggle break css definitions, this is definitely not the right way.

Besides, what if user want to toggle justify, indent and every possible user preference for every different made ebook? Do you have enough time to do all those in toggles? Even you do, user have to remember shortcut key for every feature, or customize for every single ebook, is that good?

Just supporting load custom css is an obvious useful feature, and it's very easy to implement in KPV. KPV already load different css for different extensions, I knew I can do what I wanted by reusing the extensions with current KPV, but supporting custom css is the better way.

And it's not kpv developer's burden to provide all kinds of css files. If user have the need, he can modify css very easily or ask somebody else who can edit css.

tigran123:
I completely agree that loading a custom css file (and remembering the choice in the settings) is a very useful feature.

What I meant was that users shouldn't expect the KPV distribution to ship all different combinations of possible css styles.

dracodoc:
Of course, I didn't suggest KPV to provide all kinds of css, that will be an impossible task. Every user have different preferences, and there are many ebooks made by all kinds of people. The current css can be a very good starting point, user can modify them or learn from other css files.  And current KPV always use its own css to replace the css in ebook, I personally like this approach, but maybe some users may want to use their css or the css in ebook, the load custom css feature will solve everybody's problem.

kai771:
[...]

PR #721, opened 2013-01-18
add css file selection menu for crengine

houqp:


ghost:
Do you see some reason for user to switch stylesheet on the fly?

tigran123:
@NuPogodi Welcome back. The only reason mentioned so far was the ability to toggle hyphenation on and off. Do you think it would be better to implement this (hyphenation control) as a separate toggle instead? Or maybe you have an altogether different better idea?

chrox:
It seems that font-family property in stylesheet is overwritten by selected font. And the `font-family` definition is never used in the following CSS snippet.

```
p {
    border: 0;
    display: block;
    font-family: FZYaSong-M-GBK, 宋体, serif;
    font-size: 1.0em;
    line-height: 1.5;
    margin: 0;
    padding: 0;
    }
```

And I think it's better to use CSS shipped with the EPUB file by default and leave the freedom of choosing pre-defined CSS to users.

ghost:
I'm not exactly back (not enough time and inspiration to learn new KPV features), just a curiosity about the reason to switch css's. Hyphenation? Some time ago I've already introduced some corrections allowing me to select between "no hyphenation", "algorithmic hyphenation" (universal method -- it works properly with Russian texts, but gives terrible results when applied to, say, German books) and tex-dictionaries available in crengine distribution -- iirc, in crengine/data/hyph. I'm not sure whether this way is right or wrong, but it works somehow.

houqp:
@chrox, font-family overwritten by selected font is somehow expected. I guess the setting in css should has lowest priority. 

Using the embed CSS is indeed the right thing to do, I totally missed that ;p Will check this on weekend.

@NuPogodi , long time no see ;p 

Currently for me the most useful thing of switching css is to adjust indent. One of my book is rendered with very large indent, which looks very uncomfortable.

ghost:
@houqp , iirc, crengine generally allows to switch on/off the embedded styles -- did you try to toggle them via self.doc:setEmbeddedStyles(...) ?

static int setEmbeddedStyles(lua_State _L) {
    CreDocument *doc = (CreDocument_) luaL_checkudata(L, 1, "credocument");
    doc->text_view->doCommand(DCMD_SET_INTERNAL_STYLES, luaL_checkint(L, 2));
    return 0;
}

houqp:
@NuPogodi , aha! That's what I am looking for :) Thanks a lot, will test it tonight.

dracodoc:
@NuPogodi , this feature was suggested by me because KPV was using its own css for all epubs, so that single css will not have best result for these different configurations:
1. Chinese Epub and English epub will have different fonts. KPV can have a default font for crengine but that means user need to manually select font again and again for the other language epub.
2. Some epub have text indented, some have no indent and expect the css to define indent, so you will need 2 different configuration for these.
3. And of course there are many other css combinations to best suit different epubs.

If KPV was using the epub embeded css, this feature may not seemed useful at first. But I think not using embeded css actually is good, because many epubs are not build with best format, or there could always be some settings you don't like, so have some preset css is a easier method.

@tigran123 , I think we discussed this before, 
https://github.com/hwhw/kindlepdfviewer/issues/710
I talked about all kinds of possible usage for custom css, and you acknowledged them. I don't know why would you say the only mentioned use for custom css is for hyphenation toggle?

I also cannot agree about your comment "The real strength and value of KPV is in the PDF and DjVu area, not in crereader." https://github.com/hwhw/kindlepdfviewer/issues/711
At least for me, the value about DjVu is ZERO, and epub feature can take 30-40% in my usage.

houqp:
Ha, yes, forgot to mention that the idea of customized css is from @dracodoc :)

@NuPogodi , did you succeed with `doc->text_view->doCommand(DCMD_SET_INTERNAL_STYLES, 1)` before? Seems that it's not working for me :(

ghost:
@houqp, i did not test it, but i suggest that "1" (default value) causes crengine to use the styles embedded into the book, whilst "0" means the external styles (css)
@dracodoc, your idea might work, i'm just a bit afraid of the css-file number, that drastically increases with the number of parameters to be adjusted. i'd suggest to add only two toggles -- for embedded fonts and for embedded styles -- and activate them whenever i see bad formatted book. Saving these toggles to the book history also looks quite reasonable. It's nothing more then just my personal feeling... i just did it this way in my outdated kpv-version, but unfortunately never have an opportunity to test how it works -- i mostly read fb2's without embedded styles & fonts

houqp:
OK, the `DCMD_SET_INTERNAL_STYLES` is actually working :) I was misled by it because when I set it to `1`, crengine won't clear all the existing styles, but just apply the embedded styles to current document.

Thanks @NuPogodi 

Now we only use embedded css by default, and use can change to other css by using the menu and toggle embedded css on/off in the reader config dialog.

Issue #569, opened 2012-11-13
configuable epub styles


dracodoc:
KPV use its own css for epub, which have lots of advantages and I actually like it. However that also means I need to adjust font size and line spacing manually. This is not a problem in many times because I will simply use same settings for all my English books, but I need to select font, line spacing for every new Chinese epub, because I cannot make the font selection default, that will break font selection for English epub.

So I'm suggesting to make configurable styles like following:
1. press a key to save current display styles, including font, font size, line spacing etc. The style name could be inputted or just use the current book name.
2. press a key to load previous saved display styles and apply one of them.

This should be easy to implement and can work as a alternative to customizable css support without big changes on crengine.

Besides, since the format information talked here are saved in file history already, these "styles" can be seen as a special file history. So "save current style" in a book could mean just pick part of the history information and write to a special "history file". Then any other book with same type could "apply styles" and select one of the special history file, actually just merge the special history with current file history and reopen the file with the new history, this will serve the purpose of applying styles. 

Of course this is just one way to do this, and I'm sure there could be other better methods.

mayabtlaloc:
I also join this function is implemented

kai771:
If anyone is interested, you can choose a style sheet to use per book in Librerator.

houqp:
closed since feature implemented in master

PR #727, opened 2013-01-22
add css menu and embedded css toggle

houqp:
Now, only embedded css is loaded by default. I am not sure whether crengine is designed to load `format.css` by default, so we need further testing on this change.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -45,11 +45,9 @@ static int initCache(lua_State *L) {
 }
 
 static int newDocView(lua_State *L) {
-	const char *style_sheet = luaL_checkstring(L, 1);
-	int width = luaL_checkint(L, 2);
-	int height = luaL_checkint(L, 3);
-	LVDocViewMode view_mode = (LVDocViewMode)luaL_checkint(L, 4);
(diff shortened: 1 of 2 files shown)


houqp:
this should close #569 and #711

tigran123:
Thank you, I will test it today.

tigran123:
Trying to open an fb2 file results in:

```
FATAL ERROR #-1: Unknown fatal error
```

But maybe this is because I have history files from the pre-PR state. I'll remove them and will try again.

tigran123:
One bug: invoking css selection menu via `S` and then pressing `Back` to get out of it needs to refresh the page. Currently it doesn't and the css selection menu is still shown on the screen until the user turns the page or does something else that refreshes it (e.g. `Space` for manual refresh).

houqp:
Oh, I put the redraw current page in the wrong place ;p

houqp:
BTW, did you still have the history file that crashes the reader? I am curious about the cause of the crash.

tigran123:
No, I didn't save those history files unfortunately.

PR #734, opened 2013-02-02
add manual page crop for pdf/djvu documents

chrox:


frontend/ui/reader/readercropping.lua
```diff
@@ -0,0 +1,205 @@
...
+	UIManager.repaint_all = true
+end
+
+function BBoxWidget:onCancelCrop()
+	UIManager:close(self)
+	self.ui:handleEvent(Event:new("SetZoomMode", self.orig_zoom_mode))
+	self.document.configurable.text_wrap = self.orig_reflow_mode
+	UIManager.repaint_all = true
+end
+
```
houqp, line 2: shall we put all the BBoxWdiget related definitions into a bboxwidget.lua file? That will make the structure more clear and if other widget want to use BBoxWdiget, they don't need to include readercropping.lua.

chrox, line 2: Sounds great. Patch appended.

frontend/ui/bbox.lua
```diff
@@ -0,0 +1,150 @@
...
+					range = Geom:new{
+						x = 0, y = 0,
+						w = Screen:getWidth(),
+						h = Screen:getHeight()
+					}
+				}
+			},
+			CancelCrop = {
+				GestureRange:new{
+					ges = "hold_release",
```
houqp, line 88: will be better if we use "hold" here, so use can see the effect once hold event is fired.

chrox, line 88: Yes. "hold" should be more natural and reasonable. I will send another PR on this issue.

PR #735, opened 2013-02-02
set menu font size according to screen dpi

chrox:


frontend/ui/menu.lua
```diff
@@ -195,13 +195,13 @@ Widget that displays menu
...
+	cface = Font:getFace("cfont", math.floor(22*Screen:getDPI()/167)),
 	-- face for menu title
-	tface = Font:getFace("tfont", 25),
+	tface = Font:getFace("tfont", math.floor(25*Screen:getDPI()/167)),
 	-- face for paging info display
-	fface = Font:getFace("ffont", 16),
+	fface = Font:getFace("ffont", math.floor(16*Screen:getDPI()/167)),
 	-- font for item shortcut
-	sface = Font:getFace("scfont", 20),
+	sface = Font:getFace("scfont", math.floor(20*Screen:getDPI()/167)),
```
houqp, line 205: I think this makes the code difficult to read and maintain, how about moving the calculation to `font.lua` in side `Font:getFace(font, size)` so we can hide this from UI programming?

PR #737, opened 2013-02-06
cache bug fix

houqp:
As reported in http://www.mobileread.com/forums/showpost.php?p=2410243&postcount=95

Any one who want to get the test file, just send me a mail :)

### DIFF
--- a/unireader.lua
+++ b/unireader.lua
@@ -1055,6 +1055,12 @@ function UniReader:saveLastPageOrPos()
 	self.settings:saveSetting("last_page", self.pageno)
 end
 
+function UniReader:cacheFreeItem(key)
+	self.cache[key].bb:free()
+	self.cache_current_memsize = self.cache_current_memsize - self.cache[key].size
+	self.cache[key] = nil
+end
+
 -- guarantee that we have enough memory in cache
 function UniReader:cacheClaim(size)
 	if(size > self.cache_max_memsize) then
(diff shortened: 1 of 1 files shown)


unireader.lua
```diff
@@ -1147,6 +1151,8 @@ function UniReader:drawOrCache(no, preCache)
 			return pagehash,
 				offset_x_in_page - self.cache[pagehash].x,
 				offset_y_in_page - self.cache[pagehash].y
+		else
+			self:cacheFreeItem(pagehash)
```
hwhw, line 1155: does this really fix the problem? Because if the cache isn't big enough, not used parts should get evicted anyway. Maybe the reserved cache size was too big in general? In that case, this is just masking the problem.

houqp, line 1155: It was too late last night, so I did not dig into the problem. Not sure whether I am right. Following is what I observed:

In two column mode, firstly, part of page 2 is cached, which roughly took more then half of the whole cache. Then when the user request the other part, and it turns out that the cached part does not contain the requested area. So unireader render the new part and put it into cache and increases the current cache size.

Here is how things went wrong, since these two parts are in the same page, their related cache item have the same hash value. So when the second cache item is put into the cache pool, it overwrite the first one using the same hash key. Then we end up with a cache pool with only one cache item but the current cache size is the sum of first and second cache item. This caused a endless loop in the cacheClaim function's while loop.

So in sum, the not used part is actually overwritten instead of being evicted.

hwhw, line 1155: Hi,

ah, I see, a race. Well, if it fixes it for now, we can let it go...

-hwh

2013/2/6 {Qingping,Dave} Hou notifications@github.com

PR #739, opened 2013-02-08
bugfix: restore original zoom mode after cropping

chrox:


houqp:
Thanks, cropping works great now

PR #740, opened 2013-02-09
KUAL extensions

twobob:
KUAL extensions added (since kindlepdfviewer will soon support all devices)
Added README.txt with helpful links to this extensions package.

Support notes thread here http://www.mobileread.com/forums/showpost.php?p=2414040&postcount=5

### DIFF
--- a/extensions/kpdf/README.txt
+++ b/extensions/kpdf/README.txt
@@ -1,0 +1,23 @@
+TITLE: kindlepdfviewer
+
+CONTAINS: KUAL files for installation in /mnt/us/extensions/ 
+
+REF: Main KUAL thread http://www.mobileread.com/forums/showthread.php?t=203326
+
+UPSTREAM SOURCE: https://github.com/hwhw/kindlepdfviewer
+
+SUPPORTS: 
+DX Yes ( All Revisions )
+K3 Yes ( All Revisions )
+K4 Yes ( All Revisions )
+Touch (check github and threads for support information)
+PW (check github and threads for support information)
+
+REQUIRES: KindlePDFviewer http://www.mobileread.com/forums/showthread.php?t=157047
+
+ARCHIVAL LINK: http://www.mobileread.com/forums/showpost.php?p=2406316&postcount=5
+
+ORIGINAL AUTHOR: Hawhill, See GIT for subsequent contributors listing
+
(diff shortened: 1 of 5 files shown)


twobob:
On a touch device on the current kindlepdfviewer-nightly-20130209 nightly build this would result in a harmless, benign, silent fail.

``` shell
[root@kindle kpdf]# bin/kpdf.sh 
bin/kpdf.sh: line 4: can't create /proc/keypad: nonexistent directory
bin/kpdf.sh: line 5: can't create /proc/fiveway: nonexistent directory
lua config error: ./reader.lua:119: only 4bpp is supported for now, got 8 bpp
Segmentation fault
/dev/loop/4 /mnt/us/kindlepdfviewer/fonts/host cramfs ro,noatime,nodiratime 0 0
lipc-wait-event exited with status 15.
lipc-wait-event terminated by signal: 15.
```

Other devices should work OOTB. Once underlying core support is in place the extension should "just work".

tigran123:
Thank you. Is there any reason why you use the old version of kpdf.sh with all that killall lipc\* and cat crash.log etc? Doesn't the new way (via popen_noshell) work with KUAL? Just curious...

twobob:
" tigran123 commented 4 hours ago
Thank you. Is there any reason why you use the old version of kpdf.sh with all that killall lipc\* and cat crash.log etc? Doesn't the new way (via popen_noshell) work with KUAL? Just curious..."

Not at all. It's purely an oversight, and in retrospect would be better reworked as simply a link to the /mnt/us/kindlepdfviewer/kpdf.sh I think?

twobob:
Scratch this then. I'll rebase and do an internally dependent version. Raise another pull when I'm done yeah?

tigran123:
In the source tree sym links are OK, because the development system is Linux with ext4 filesystem, but on the target Kindle the fs is VFAT which has no concept of symbolic links.

twobob:
yep. understood. this will be a shell call direct the the relevant script.

Won't be affected :)

houqp:
@twobob , will you send another PR to add KUAL support?

twobob:
yep. was beavering away on testing the touch stuff as I really want this to do all devices.

Will raise PR ASAP

PR #741, opened 2013-02-10
bugfix: calculate current config value or arg at runtime

chrox:
This patch fixed the screen rotation bug that once going into landscape mode from document menu the config menu cannot rotate screen any more.

### DIFF
--- a/frontend/ui/reader/readerview.lua
+++ b/frontend/ui/reader/readerview.lua
@@ -132,7 +132,16 @@ end
 
 function ReaderView:onSetScreenMode(new_mode)
 	if new_mode == "landscape" or new_mode == "portrait" then
+		self.screen_mode = new_mode
 		Screen:setScreenMode(new_mode)
+		self.ui:handleEvent(Event:new("SetDimensions", Screen:getSize()))
+	end
+	return true
+end
+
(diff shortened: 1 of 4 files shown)


houqp:
Thanks for the fixes and Happy Chinese new year!

chrox:
The postInitCallback facility is great. I will fix it that way tonight. And I think it's better to use screen dimension explicitly in page hash since it will save several CPU cycles.

chrox:
Thanks. And happy Spring Festival!

frontend/ui/reader/readerview.lua
```diff
@@ -132,12 +132,21 @@ end
...
 function ReaderView:onSetScreenMode(new_mode)
 	if new_mode == "landscape" or new_mode == "portrait" then
+		self.screen_mode = new_mode
 		Screen:setScreenMode(new_mode)
 		self.ui:handleEvent(Event:new("SetDimensions", Screen:getSize()))
 	end
 	return true
 end
 
+function ReaderView:onRestoreScreenMode(old_mode)
```
houqp, line 142: This event handler is created just for restoring screen mode when returning to filechooser right?

PR #743, opened 2013-02-17
fix visible_area recalculate for manual crop in reflow mode

houqp:
Calling `self.ui:handleEvent(Event:new("SetZoomMode", "page"))` will have no effect because the original page mode is always page mode in reflow mode. In that case, we need to call ReaderView:recalculate to  force update visible_area.

### DIFF
--- a/frontend/ui/reader/readercropping.lua
+++ b/frontend/ui/reader/readercropping.lua
@@ -5,14 +5,17 @@ ReaderCropping = InputContainer:new{}
 function ReaderCropping:onPageCrop(mode)
 	if mode == "auto" then return end
 	local orig_reflow_mode = self.document.configurable.text_wrap
-	self.document.configurable.text_wrap = 0
 	self.ui:handleEvent(Event:new("CloseConfig"))
 	self.cropping_zoommode = true
 	self.cropping_offset = true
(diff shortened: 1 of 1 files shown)


chrox:
We have to set zoom mode to 'page' before manual cropping in case that we are not in reflow mode and zoom mode is not 'page'.

houqp:
Oh, yes, you are right. Fix attached :)

frontend/ui/reader/readercropping.lua
```diff
@@ -5,12 +5,17 @@ ReaderCropping = InputContainer:new{}
...
-	self.ui:handleEvent(Event:new("SetZoomMode", "page"))
-	self.cropping_zoommode = false
+	if self.document.configurable.text_wrap == 1 then
+		self.document.configurable.text_wrap = 0
+		-- if we are in reflow mode, then we are already in page
+		-- mode, just force readerview to recalculate visible_area
+		self.view:recalculate()
+	else
+		self.ui:handleEvent(Event:new("SetZoomMode", "page"))
+	end
```
chrox, line 18: Should add `self.cropping_zoommode = false` here to filter out zoom mode in cropping so that `ReaderCropping:onSetZoomMode` could capture real original zoom mode.

PR #745, opened 2013-02-17
fix bug in getPitch, returning half of ptich for 4bpp fb

houqp:


chrox:
Is this patch to fix the screen size bug? But it still returns 

```
# Screen size {
    ["h"] = 1024,
    ["w"] = 768
}
```

after applying this patch.

houqp:
No, it's not. Still debuging OS project with my partner at school right now :(  I hate multi threading....

I need to have my KPW to do the test on real FB.

PR #747, opened 2013-02-18
fix visible_area recalculate for manual crop in reflow mode

chrox:


chrox:
This PR is sent to replace #743.

PR #748, opened 2013-02-18
add extensions directory in customupdate

chrox:


houqp:
After we have KUAL, I think it's safe to remove kite since it's no longer under maintenance :P

hwhw:
absolutely agreed. I also have some reservations towards kite - as it broke
for me, too. Cannot check its sources - another problem for me. So yes, I
think we should go KUAL and launchpad.

houqp:
OK, done :)

PR #750, opened 2013-02-18
bug fix for einkfb.c and screen.lua

### LINKED ISSUE #744: Screen getSize returns wrong width on Paperwhite
Now Screen:getSize() returns 

```
# Screen size {
    ["h"] = 1024,
    ["w"] = 768
}
```

It ought to be 1024*758. This bug was introduced in #732.

houqp:
this should fix #744


chrox:
Cool. It works now.

PR #758, opened 2013-02-19
set zoom mode to content after manual cropping

chrox:


chrox:
This PR is closed since it is included in #760.

Issue #723, opened 2013-01-18
zoom mode and reflow


th0mass:
I think zoom mode disturbes the reflow process. For example when I try to use reflow when it is on [contentwidth] zoom  mode output will be a mess. How can I set the KPV so that zoom mode of the original view format will not have any effect on the reflow format?
[Touch with the nightly builds ]

houqp:
um... I am not the reflow expert here, @chrox should have the answer. But I am quite sure he is on vacation right now ;p

chrox:
For now when switched to reflow mode the zoom mode is set to 'page' but the original zoom mode cannot be restored yet. I will send a PR to restore the original zoom mode after switching from reflow mode to ordinary mode.

chrox:
Close this issue since the fix is already merged.

PR #760, opened 2013-02-19
restore zoom mode when toggling reflow on/off

chrox:


chrox:
This patch fixes zoom mode issue in #723.

PR #766, opened 2013-02-23
add reading progress bar at the bottom of ReaderView for pdf/djvu reader

chrox:
And there is also a toggle full screen option to turn on/off the progress bar.

### DIFF
--- a/frontend/document/koptinterface.lua
+++ b/frontend/document/koptinterface.lua
@@ -39,6 +39,15 @@ KoptOptions = {
 		icon = "resources/icons/appbar.column.two.large.png",
 		options = {
 			{
+				name = "full_screen",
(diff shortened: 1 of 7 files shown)


houqp:
Looks like footer.lua is not a generic ui widget, i.e. it depends on ReaderView and document modules.

How about moving footer.lua to `frontend/reader` directory and rename to ReaderFooter?

chrox:
Yes. That's it. I'm afraid I forgot to git add the widget.lua in this PR which has a better looking progress bar widget. I will add it right now.

houqp:
Looks good, thanks :)

I thinks it's time to use white as default background color now. Grey doesn't come very well with progress bar.

PR #774, opened 2013-02-27
record multiple slots for multi-touch input

houqp:
for now, only slot 0 is parsed, the rest are ignored.


chrox:
It works now. Thanks.

PR #776, opened 2013-03-02
add multiple slots in gesture detector

chrox:


houqp:
Thanks for this nice feature :)

PR #779, opened 2013-03-06
add screenshot with two-finger tap

chrox:


houqp:
Thanks, this is awesome :)

PR #780, opened 2013-03-09
switch view_mode to scroll for search results

dpavlin:
This will position page on first search result.


houqp:
long time no see @dpavlin  ;p

dpavlin:
I had been busy last few months, but I found time again to read something, so I found this bug :-)

PR #781, opened 2013-03-10
add scroll mode for pdf/djvu reader

chrox:


houqp:
wow! finally scroll mode for pdfs. I will test it tonight :)

PR #798, opened 2013-03-15
add touchmenu

houqp:
changes included:
- touchmenu
  - in case you wonder why I waste a whole bottom line to just display a go back button: currently, touchmenu does not support multiple page, so that bottom will be used to display current page/ total page in the future.
- LvDEBUG, so we can specify recursive level to avoid stack overflow when dumping a widget
- invert support in framecontainer
- some other fixes I cannot remember :(


chrox:
Great work. Thanks.

PR #804, opened 2013-03-16
bugfix: calculate width of TouchMenu each time when TouchMenu is created

chrox:


frontend/ui/reader/readermenu.lua
```diff
@@ -75,6 +75,7 @@ function ReaderMenu:onShowMenu()
 	if Device:isTouchDevice() then
 		main_menu = TouchMenu:new{
 			name = "wocao",
```
houqp: you really should delete this debug code for me....

chrox: I thought it was a placeholder or something. :P

PR #809, opened 2013-03-17
Add a home button to readermenu

clenton:


houqp:
wow, that's very fast :)

Just one more thing, a better practice for collaborating on git is always create your own branch and make changes inside that. Once you finished it, sent a pull request to the upstream main branch (in our case new_ui_code). That way, you can only use your local main branch to sync with the upstream one, thus avoiding conflicts.

Other than that, perfect :+1:

chrox:
And I found it useful to reset your local new_ui_code with the upstream one before creating new branch like this.

```
git checkout new_ui_code
git fetch upstream
git reset --hard upstream/new_ui_code
git checkout -b fix_branch
# some fix #
git push origin fix_branch
```

clenton:
Thanks for your help, I am doing it

Issue #815, opened 2013-03-23
readers crash when paging from TOC


chrox:
I didn't reproduce this bug in the emulator. But it happened in both pdf reader and epub reader on Kindle. Debug info says 

```
*** glibc detected *** ./kpdfview: double free or corruption (!prev): 0x00655630 ***
```

or simply: `Segmentation fault`.

chrox:
And more tests show that it crashes much more often in landscape mode than portrait mode. You can reproduce it by opening any pdf documents with TOC information and just selecting several TOC entries. It will crash randomly with these info:

```
Segmentation fault
```

or

```
*** glibc detected *** ./kpdfview: double free or corruption (!prev): 0x00655630 ***
```

or

```
*** glibc detected *** ./kpdfview: malloc(): smallbin double linked list corrupted: 0x007084a8 ***
```

or

```
kpdfview: malloc.c:4636: _int_malloc: Assertion `victim->bk_nextsize->fd_nextsize == victim' failed.
```

chrox:
This bug also exists in kindlepdfviewer-v2012.04.2-676-g72eecac.zip, the very first version we introduced landscape screen mode.

houqp:
um, heap corruption bug....

will take a look tomorrow :P

houqp:
close for now

PR #820, opened 2013-03-24
bugfix: set toc/bookmark window dimension smaller than screen size to fix Toc crash in #815

chrox:


houqp:
good catch. Though I don't really understand why ;p

But I guess you need to send it to new repo now :)

chrox:
Since the nightly build script hasn't supported koreader yet we probably
need to fix it in the new_ui_code branch. And then cherry-pick this patch
into koreader repo.

hwhw:
Agreed. But we probably need to get rid of the new_ui_code branch as fast as possible. However - it's clear that the build workflow must be working and the nightly builds set up, yes.

houqp:
Oh, yeah, you are right :)

PR #825, opened 2013-03-25
Adaptions for new koreader-base framework

hwhw:
- text rendering: uses 0/1.0 for foreground/background
- emu input: uses fifo file now

This is not yet perfect, rendering text doesn't seem to "add" the values
correctly.

### DIFF
--- a/reader.lua
+++ b/reader.lua
@@ -98,7 +98,9 @@ else
 end
 
 if util.isEmulated()==1 then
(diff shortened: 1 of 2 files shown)


hwhw:
At least, it will now run again in emulation mode. There might be other oversights, though.

houqp:
thanks :)

PR #826, opened 2013-03-26
add background color

dpavlin:
Fix crash on file browser.

This additional parameter is probably reason why we see white letters on backgrounds which aren't white, but I didn't have time to investigate that yet.

Great work so far, thanks everyone!


houqp:
:+1: 

Hopefully I can also get sometime to work on kindlepdfviewer this weekend :)

PR #830, opened 2013-04-09
add gettext support

houqp:


houqp:
@chrox , please review it again as I have backported gettext support from koreader

PR #850, opened 2013-08-07
Adaptions for dynamic library aware koreader-base and some streamlining

hwhw:
This should be extended by a submodule update after koreader-base got really updated by the relevant PR there.


hwhw:
this should also be ready now.

PR #852, opened 2013-08-20
update koreader-base

houqp:


houqp:
I was amazed at how hard it is to update the api call in kpv code base ;p

